Issue #13, opened 2013-11-25
Interop with Non-NServiceBus Endpoints


gsogol:
When a non-NSB endpoint ((e.g. Spring Integration/Java, etc) needs to publish a message to a NServiceBus endpoint, .Net engineers force non-NSB producers to add NSB-specific semantics:

- producers have to make sure to properly add a .Net namespace Company.Package.SubPackage.

- due to NSB's serialization, objects in the XML have to be in Camel Case instead of language's (java, python, whatever) own semantics.

- message type had to be added to the message header with the exact same
name as the Exchange. For example,  Company.Package.SubPackage.Events.QuoteRequested is the XML
object we send and the Exchange and had to be defined as the Message Type
in the message header (see attached image)

- content_type had to be set to text/xml. A language like Java is defaulting to
octet-stream. Potentially, telling NServiceBus to treat X as though it's Y can solve an issue.

- CorrelationID needs to be set to the same value as the MessageID in
the message header. No reply is being expected. Java producers are confused why they need such a header. Again, potentially, telling NSB to treat some property as the CorrelationID would suffice. Producers can decide what they want to name things and we just need to tell NSB to treat some header as CorrelationID.

In summary, NServiceBus needs to be flexible when a consumer of messages. When used in a company that uses non-.Net technologies, NSB should be allowed to be easily bent to conform what producers dictate via conventions, not the other way around.

![quoterequested](https://f.cloud.github.com/assets/2625918/1614859/d9a01b3a-55f0-11e3-89e3-6939b2abe3b1.png)

johnsimons:
@andreasohlund should this be moved to the requirements repo ?

andreasohlund:
@gsogol According to this:

 https://github.com/Particular/NServiceBus.RabbitMQ/blob/message-id-strategy/src/NServiceBus.RabbitMQ.AcceptanceTests/When_using_a_custom_message_id_strategy.cs#L39

content type isn't needed. Can  you confirm?

andreasohlund:
@gsogol it also seems like correlation id isn't needed? (can you confirm this as well)

andreasohlund:
> should this be moved to the requirements repo ?

I think most of this issue is fixed. Except for the message type beeing needed. I talked to @SimonCropp and we belive we can adjust at least the xml serializer to use the root node in the xml to determine message type and if that is true we should be able to close this one

gsogol:
[...]

PR #69, opened 2015-01-07
Pluggable message id strategy

andreasohlund:
The strategy for detecting message id's are now pluggable to provide better interop with non NServiceBus endpoints

Part of fixing #13 

### Documentation

By default NServiceBus uses the `message-id` property of the [AMQP standard](https://www.rabbitmq.com/amqp-0-9-1-reference.html) to relay the stable message id. If this header isn't set the transport will throw an exception since NServiceBus needs a message id in order to perform retries, deduplication etc in a safe way. In integration scenarios where you might not control the sender you might wan't to use your own custom scheme eg. a customer header or some data contained in the actual message body. In these cases you ca plugin in your onw strategy by calling: `insert final config call here`.


src/NServiceBus.RabbitMQ.AcceptanceTests/When_using_a_custom_message_id_strategy.cs
```diff
@@ -0,0 +1,81 @@
...
+                       using (var stream = new MemoryStream())
+                       {
+                           serializer.Serialize(new MyRequest(), stream);
+
+                           using (var channel = connectionManager.GetPublishConnection().CreateModel())
+                           {
+                               var properties = channel.CreateBasicProperties();
+
+                               //for now until we can patch the serializer to infer the type based on the root node
+                               properties.Headers = new Dictionary<string, object> { { Headers.EnclosedMessageTypes, typeof(MyRequest).FullName } };
```
andreasohlund: @SimonCropp @johnsimons @yvesgoeleven should we consider making the serializers smarter when it comes to inferring messages type to make external integrations easier? (json serializers also needs to the header to be set)

For XML we can match on the root node and for json we can use the $type thing?

andreasohlund:
@johnsimons can you review this one?

PR #71, opened 2015-01-10
Only open the admin connection when we need to use it

andreasohlund:
@johnsimons I noticed that all endpoints keep the admin connection (to perform sub/unsubscribe) open. This seems wasteful since all subscribes usually happens on start up.

What do you think?

### DIFF
--- a/src/NServiceBus.RabbitMQ/RabbitMqConnectionManager.cs
+++ b/src/NServiceBus.RabbitMQ/RabbitMqConnectionManager.cs
@@ -36,7 +36,7 @@
             //note: The purpose is there so that we/users can add more advanced connection managers in the future
             lock (connectionFactory)
             {
-                return connectionAdministration ?? (connectionAdministration = new PersistentConnection(connectionFactory, connectionConfiguration.RetryDelay,"Administration"));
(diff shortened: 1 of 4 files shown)


johnsimons:
Looks good

andreasohlund:
The issue is that we're now mixing the responsibility for disposing the
connection. Lets talk it through in the bar in Cyprus

andreasohlund:
@johnsimons @johnsimons while I think the fix is good (no need to keep the admin channel open) I have a bad feeling about the RabbitMqConnectionManager class. It seems wrong that some calls returns a connection that is managed by the connection manager (publish + consume) and one (admin) that is supposed to be managed by the caller. Do U have any idea on how to clean this up?

johnsimons:
@andreasohlund how about making the API more explicit?

- GetPublishConnection rename to GetOrCreatePublishConnection
- GetConsumeConnection rename to GetOrCreateConsumeConnection
- GetAdministrationConnection rename to CreateAdministrationConnection

andreasohlund:
Yes that is better. Since it would be a breaking change I'll raise a new
issue to track it and slate if for v 3.0.0

PR #91, opened 2015-06-04
Remove the multi broker anti-feature

adamralph:


andreasohlund:
I like your commit msg prefixes. Are you following some kind of standard?

Are you using the metadata in any way?

src/NServiceBus.RabbitMQ/Connection/IClusterHostSelectionStrategy.cs
```diff
@@ -1,42 +0,0 @@
-namespace EasyNetQ
```
andreasohlund: We have a praxis of "doing more than semver requires" by leaving any public type in for one extra major to give users a helpful obsolete message. Here is how it's done in the core:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/v6-obsoletes.cs

Does this make sense?

src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationTests.cs
```diff
@@ -11,7 +11,8 @@ public class ConnectionConfigurationTests
         #region Setup/Teardown
```
andreasohlund: I know you didn't add them, but boo for regions! ;)

andreasohlund:
Nitpick: we tend to name branches with a descriptive name to avoid having to look up what `19` is

Perhaps `remove-multi-broker-feature` ?

or `19-remove-multi-broker-feature`

src/NServiceBus.RabbitMQ/Connection/RabbitMqConnectionFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    class RabbitMqConnectionFactory
+    {
+        public ConnectionConfiguration Configuration { get; private set; }
+
+        public RabbitMqConnectionFactory(ConnectionConfiguration connectionConfiguration)
+        {
+            if (connectionConfiguration == null)
+            {
+                throw new ArgumentNullException("connectionConfiguration");
```
SimonCropp: why r we arg checking in an internal class? shouldnt this be verified at the point of the public API?

adamralph:
@andreasohlund I'm not following any standard with my commit prefixes, they're just something I made up. I like to isolate my commits to red, green, refactor whenever possible. I don't use the meta data in code, at the moment it's just eyeball consumption.

adamralph:
@andreasohlund fair point about the branche name. `91` is a little lazy. I'll use the `{number}-{description slug}` format in future.

src/NServiceBus.RabbitMQ/Connection/IClusterHostSelectionStrategy.cs
adamralph: Yep, makes sense. I'll re-instate with obsoletes.

src/NServiceBus.RabbitMQ/Connection/RabbitMqConnectionFactory.cs
adamralph: Fair point. BTW, this class is a rename from `ClusterAwareConnectionFactory` so most of the code, including this check, comes from there. I'll re-visit and see if it makes sense to remove/move it.

src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationTests.cs
adamralph: double boo! I'll kill them :wink:

src/NServiceBus.RabbitMQ/Connection/RabbitMqConnectionFactory.cs
SimonCropp: Ok no worries. If this arg checking was preexisting then it doesn't need to be fixed in this pr

adamralph: OK, I'll leave it for now. I think I've come to a point with this PR that I'm polishing turds which may be removed entirely by a deeper refactoring.

It's sooo tempting to jump down the refactoring rabbit hole in this repo... :stuck_out_tongue:

src/NServiceBus.RabbitMQ/Connection/IClusterHostSelectionStrategy.cs
adamralph: @andreasohlund I've re-introduced the obsoleted interface in v3-obsoletes.cs. Do you think the PR is good to go?

andreasohlund:
Looks good to go

PR #107, opened 2015-12-11
Upgrade to v6 core

johnsimons:
Upgrade transport to use NServiceBus v6.
Connects to Particular/NServiceBus#3125

Still left to do:
- [ ] Always use ack, autoack can cause message loss, see #122
- [ ] Implement callbacks
- [ ] Decide whether SecondaryReceiveSettings still needed with current MessagePump
- [ ] Decide whether PrefetchCount config setting should be removed (and use maxconcurrency instead) or keep to override
- [ ] Read Circuit breaker settings from config
- [ ] DequeueTimeout setting not currently used, do we still need it?
- [ ] Decide whether separate HostConfiguration is still needed (we don't support multiple host connection strings any more)
- [ ] Test that inflight messages are ok to use taskid for dictionary keys
- [ ] DirectRoutingTopology "broken": should use multiple keys, not one, can't handle multiple interfaces, see #123
- [ ] Update docs
- [ ] Update samples Particular/docs.particular.net#1011


andreasohlund:
@bording @johnsimons let's rebase this one so that we can figure out if the acceptance tests hangs on this one as well? (to narrow down the issue with #114 )

Can you look at this @bording ?

bording:
I can look into rebasing this, I'm assuming you mean to rebase it onto develop since it's currently on master instead?

Regarding the tests hanging, while I've been working on the getting the main tests project green, I've noticed that it takes a very long time for the process to actually stop after a test run. Watching the RabbitMQ admin site, there are connections that are being maintained long after the test appears to have finished running, so I suspect that is the same behavior you're seeing. If it is, then this branch has the same problem and isn't related to the 1328 upgrade.

If I wait and watch the admin page, the connections do eventually disappear. I haven't timed it, but it's several minutes.

I suspect that something isn't being shutdown correctly, I'll see if I can figure out the exact cause.

bording:
I have rebased this onto develop. 

@andreasohlund has suggested we go ahead and merge this into develop and have smaller branches as we continue working vs. this single long-running branch. Anyone opposed to this?

Also, thoughts on keeping all the commits vs. squashing before merging?

johnsimons:
@bording I will deal with this when I am back after new years

mikeminutillo:
The concurrency limit is not working as expected. I'm pretty sure that it's related to the `EventingBasicConsumer` `OnMessageReceivedHandler` being` async void`.

To explain why, here is a LINQPad script:

```
void Main()
{
	var eventer = new Eventer();
	
	eventer.MessageReceived += OnMessageReceived;
	
	eventer.FireEvent();
}

public static async void OnMessageReceived(object sender, EventArgs args)
{
	"Handler - BEFORE".Dump();
	await Task.Delay(100);
	"Handler - AFTER".Dump();
}

class Eventer
{
	public event EventHandler<EventArgs> MessageReceived;
	
	public void FireEvent()
	{
		"Eventer - BEFORE".Dump();
		var evt = MessageReceived;
		if(evt != null)
			evt.Invoke(this, EventArgs.Empty);
		"Eventer - AFTER".Dump();
	}
}
```

When you run this you get the following output:
```
Eventer - BEFORE
Handler - BEFORE
Eventer - AFTER
Handler - AFTER
```

This means that the code that comes after the event is fired as soon as the first await is hit. Essentially everything before the `await` is code that constructs and returns a `Task` instance with the rest of the method as a continuation. The thing is, for `async void` this task is just thrown away and never examined.

All of this is run in the context of a Task that belongs to our concurrency limited task scheduler. In order to ensure that this Task is not returned to the scheduler for re-use, the event handler MUST block.

Now the issue is, if I make that event handler synchronous I get a deadlock and I'm not sure why. Thoughts?

mikeminutillo:
OK. Looking at it, it is obvious why this happens. The ack/reject calls try to run exclusively on the task scheduler which means we need to wait for the event handler to finish before we execute them BUT the event handler is waiting for those exclusive calls to finish. Deadlock. 

This was OK before because we were leaking our main tasks back to the scheduler as soon as we hit the first await.

andreasohlund:
Have you talked to our async master :) ? ( @danielmarbach )

danielmarbach:
@mikeminutillo I would like to pair up after Thailand with you guys. I'm almost certain we can avoid the custom scheduler.

mikeminutillo:
@danielmarbach or in Thailand if we have time. The custom scheduler is the recommended approach to limit concurrency from Rabbit .NET docs.

bording:
@mikeminutillo Where are you seeing evidence of the scheduling actually not working? Even with the async void handler and not waiting there, the actual work being done in `ProcessMessage()` will be done on a thread handled by the `ConcurrentScheduler`, so that thread will still be involved in the in the concurrency limit.

danielmarbach:
Guys, I did some crazy things and now rabbitmq client will go full async :)

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/149#issuecomment-170088100

bording:
In an effort to get more individual issues on the remaining work, should we go ahead and merge this into develop? If so, do we want to squash anything before doing so, or just bring everything in as is?

danielmarbach:
@bording I never understood why this is happening on a release branch. And the history on this branch is way to large. If you can make it compilable I'd say go to develop and from their on focus on making small PRs against develop

bording:
I think that's just the name that @johnsimons chose for this branch when he created it.

This branch is fully compilable already, no problems there.

So you're thinking squash this down and merge it in?

bording:
Here is a list of all of the various remaining tasks that I'm aware of and things that might need to be investigated:

- noAck should never be enabled, can cause message loss (this is a good candidate for a backport fix)
- callbacks
- SecondaryReceiveSettings still needed with current MessagePump?
- scaleout / instance discriminator
- prefetchCount config setting: remove (and use maxconcurrency instead) or keep to override?
- circuit breaker settings from config?
- DequeueTimeout setting not currently used, still needed?
- separate HostConfiguration stuff still needed if we don't support multiple host connection strings any more?
- inflight messages ok to use taskid for dictionary key?
- DirectRoutingTopology "broken": should use multiple keys, not one, can't handle multiple interfaces
- BasicPublish calls: mandatory vs. publisher confirms & handling rejection/failure here
- Connection / channel management: removing GetConsumeConnection() and some way to reuse channel across multiple sends for perf
- Acceptance test cleanup: instead of just purging queues, why not delete queues and exchanges?

I'm planning on opening up separate issues for these where I'll go into more detail, but I wanted to get this list up somewhere, and this PR seems like the best place for now.

johnsimons:
Let's talk about it in about 2 hrs

On Thu, 28 Jan 2016 at 06:25 Brandon Ording <notifications@github.com>
wrote:

> Here is a list of all of the various remaining tasks that I'm aware of and
> things that might need to be investigated:
>
>    - noAck should never be enabled, can cause message loss (this is a
>    good candidate for a backport fix)
>    - callbacks
>    - SecondaryReceiveSettings still needed with current MessagePump?
>    - scaleout / instance discriminator
>    - prefetchCount config setting: remove (and use maxconcurrency
>    instead) or keep to override?
>    - circuit breaker settings from config?
>    - DequeueTimeout setting not currently used, still needed?
>    - separate HostConfiguration stuff still needed if we don't support
>    multiple host connection strings any more?
>    - inflight messages ok to use taskid for dictionary key?
>    - DirectRoutingTopology "broken": should use multiple keys, not one,
>    can't handle multiple interfaces
>    - BasicPublish calls: mandatory vs. publisher confirms & handling
>    rejection/failure here
>    - Connection / channel management: removing GetConsumeConnection() and
>    some way to reuse channel across multiple sends for perf
>    - Acceptance test cleanup: instead of just purging queues, why not
>    delete queues and exchanges?
>
> I'm planning on opening up separate issues for these where I'll go into
> more detail, but I wanted to get this list up somewhere, and this PR seems
> like the best place for now.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/NServiceBus.RabbitMQ/pull/107#issuecomment-175838002>
> .
>
-- 

*John Simons*
*Engineer, Particular Software*

Email*:*  john.simons@particular.net
Web:    www.particular.net

johnsimons:
> I never understood why this is happening on a release branch

@danielmarbach that is bad naming from my side, we should rename it to core-v6-upgrade, @bording want to do the renaming ?

@danielmarbach regarding using develop branch for the upgrade to v6, the reason I have deliberately are working on this feature branch is so that this PR is scoped to only the v6 upgrade and no scope creep happens, does this make sense ?

johnsimons:
> instead of just purging queues, why not delete queues and exchanges?

@bording I am pretty sure @mikeminutillo and I tried that and it was unreliable

johnsimons:
@bording updates description with task list as discussed.

Not sure about the following:
- BasicPublish calls: mandatory vs. publisher confirms & handling
    rejection/failure here
- Connection / channel management: removing GetConsumeConnection() and
    some way to reuse channel across multiple sends for perf

johnsimons:
Replaced by #124

PR #109, opened 2015-12-16
Moved the callback handling into the pump and dispatcher

andreasohlund:
@johnsimons what do you think?


andreasohlund:
There is a bit more coupling here but I'd say its ok since callbacks very much is tied to receiving a message and also dispatching messages

andreasohlund:
Synked with @johnsimons and we decided to merge

PR #110, opened 2015-12-16
Use task.run

andreasohlund:


andreasohlund:
@johnsimons lets see if this works

andreasohlund:
@johnsimons this seems to fix the inspection issue

Issue #100, opened 2015-08-24
Default PrefetchCount


JoelGoben:
The default PrefetchCount is now the MaximumConcurrencyLevel, aka the number of processing threads.  This is a per thread setting though, so if for example 10 threads are configured, the result is that each of the 10 threads prefetches another 10 messages, for a total of 100 prefetched messages.  This seems less than ideal.  The PrefetchCount setting should default to prefetch 1 record per thread.

andreasohlund:
Makes sense now, I got confused by this https://www.rabbitmq.com/consumer-prefetch.html mixing "connection"/ "channel" in a vierd way :(

Since you can configure the prefetch  explicitly, http://docs.particular.net/nservicebus/rabbitmq/configuration-api#configuring-rabbitmq-transport-to-be-used, I assume you're not blocked by this?

JoelGoben:
Correct, I have set PrefetchCount to 1 explicitly, and am not blocked.

bording:
We've made changes to the still-in-progress next version of the transport that will work with the NSB v6 release, and the behavior around the PrefetchCount setting is changing.

With the way the new message pump is designed, we won't need to expose a separate PrefetchCount setting, and will instead rely on the MaximumConcurrency setting to determine what the PrefetchCount should be. We can do this because we're handling concurrency by using a single consumer and the PrefetchCount determines the number of simultaneous in-flight messages.

PR #116, opened 2015-12-24
[WIP] New MessagePump

bording:
I wasn't really happy with how complex the RabbitMqMessagePump was, and there seemed to be some concurrency problems, and some problems related to how it was attempting to pool channels.

I decided to try for a redesigned, simpler version, and here is what I have so far. All of the unit tests are passing (except for the callbacks header one which we still need to decide on how that's going to work)
and most of the ATs are too.

Things that I know still need to be added:
- ~~Poison messages being moved to the error queue~~
- ~~MaxConcurrency~~
- Hooking the other configurable bits back up (~~noAck~~, circuit breaker, ~~PurgeOnStartup~~, ~~BasicQos / PrefetchCount~~)
- Other misc error handling
- ~~Using the callbacks header address~~

Most of that should be pretty straightforward to add back in. To handle the concurrency, the RabbitMQ docs recommend using a custom Task Scheduler for that. Microsoft has a ParallelExtensionsExtras library that has a LimitedConcurrencyLevelTaskScheduler class that I was planning on using for that.

Let me know what you think of this. I think this is a better approach and it already appears to be working better so far!


mikeminutillo:
:+1: It looks good so far

andreasohlund:
>Microsoft has a ParallelExtensionsExtras library that has a LimitedConcurrencyLevelTaskScheduler class that I was planning on using for that.

Just copy paste the code and attribute it properly, dependencies FTL :)

src/NServiceBus.RabbitMQ/MessagePump.cs
```diff
@@ -0,0 +1,163 @@
...
+            return TaskEx.Completed;
+        }
+
+        public void Start(PushRuntimeSettings limitations)
+        {
+            var taskScheduler = new LimitedConcurrencyLevelTaskScheduler(limitations.MaxConcurrency);
+            var factory = new RabbitMqConnectionFactory(connectionConfiguration, taskScheduler);
+            var connection = new PersistentConnection(factory, connectionConfiguration.RetryDelay, "Consume");
+            var model = connection.CreateModel();
+            model.BasicQos(0, GetPrefetchGount(limitations.MaxConcurrency), false);
```
andreasohlund: I think this is wrong, see the discussion here: https://github.com/Particular/NServiceBus.RabbitMQ/issues/100

src/NServiceBus.RabbitMQ/MessagePump.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            consumer = new EventingBasicConsumer(model);
+            consumerShutdownCompleted = new TaskCompletionSource<bool>();
+
+            consumer.Received += (sender, eventArgs) =>
```
andreasohlund: This means that we have a "single threaded peek" right? (I remember reading that you need multiple consumers in order to max out receive performance, needs to be tested though)

src/NServiceBus.RabbitMQ/MessagePump.cs
bording: Right, I was thinking the same thing, which is one of the reasons I hadn't added it back in yet. I was planning on doing some investigation on the ideal setting for that.

src/NServiceBus.RabbitMQ/MessagePump.cs
bording: Looking at the implemenation of `EventingBasicConsumer`, there is nothing about how it works that would be different from the `DefaultBasicConsumer` from a concurrency standpoint.

The [docs](https://www.rabbitmq.com/dotnet-api-guide.html#concurrency) say that "Consumer callbacks are invoked in tasks dispatched to the default TaskScheduler provided by the .NET runtime."

By using a custom TaskScheduler instead of the default, we can set a maximum concurrency value.

Do you have a link to something that talks about needing multiple consumers?

andreasohlund: Can't find anything, will continue to dig...

The eventing consumer is flagged as experimental

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/events/EventingBasicConsumer.cs#L45

are we concerned with that? (perhaps ask the maintainers?)

bording: As of the 3.6.0 release (which we still need to upgrade to, currently on 3.5.6) it's no longer experimental:

https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/4384c09e76bf09865af4a52d993e2d5e0f1f27f6

Though it looks like they missed line 45 in that change.

andreasohlund: :+1:

johnsimons:
@bording it seems you are sharing the channel in multithreading ?

johnsimons:
For optimal performance we need multiple channels, and they cannot be reused by multiple threads

johnsimons:
I am back on the 4th and can review this better.

mikeminutillo:
In a review with @johnsimons 
- We cannot use a Concurrency Limited Scheduler as this means any user code that starts a new task (say in a handler) will inherit this task scheduler (I did not know that would happen). This could lead to deadlock situations.
- We shouldn't share a single`IModel` instance between multiple threads but we should Ack/Reject on the same channel that did the receive. See [Sharing Channels Between Threads](https://www.rabbitmq.com/dotnet-api-guide.html). This means that we'd need to put locks around our Ack/Reject calls to ensure that multiple threads aren't calling them at the same time. The legacy implementation had one receiver model per concurrency thread which solves this. The other async implementation passes the model to the new thread but an Object Pool implementation ensures that only one thread at a time can access them.

PR #119, opened 2016-01-12
Using cancellation tokens to support canceling of the receive

Scooletz:
Related to Particular/NServiceBus#3254
Allowing to cancel the in flight message processing by the CancellationTokenSource passed via PushContext.
I'm following this approach on propagating this change:
- PR for the core
- PRs for all the transports aligned to the artifacts from the PR on the core.

So there's no artifacts on myget yet. The build was done with the following core artifacts: https://builds.particular.net/viewLog.html?buildId=140493&buildTypeId=NServiceBusCore_Build&tab=artifacts and will be feasible to build against core as soon as I've got the initial ack on all the transports.


Scooletz:
@andreasohlund @johnsimons If you could do the initial review, that would be great.

src/NServiceBus.RabbitMQ/MessagePump.cs
```diff
@@ -157,14 +158,18 @@ public void Start(PushRuntimeSettings limitations)
...
                 {
-                    await PushMessageToPipe(messageId, headers, new MemoryStream(message.Body ?? new byte[0])).ConfigureAwait(false);
+                    tokenSource = new CancellationTokenSource();
+                    await PushMessageToPipe(messageId, headers, tokenSource, new MemoryStream(message.Body ?? new byte[0])).ConfigureAwait(false);
                 }
 
-                if (!noAck)
+                var token = (tokenSource?.Token).GetValueOrDefault();
+
+                if (!noAck && !token.IsCancellationRequested)
```
johnsimons: @Scooletz wouldn't it make more sense to just check if `token.IsCancellationRequested == true` and then call `channel.BasicReject` ?

src/NServiceBus.RabbitMQ/MessagePump.cs
```diff
@@ -157,14 +158,18 @@ public void Start(PushRuntimeSettings limitations)
...
                 {
-                    var task = new Task(() => { channel.BasicAck(message.DeliveryTag, false); });
+                    var task = new Task(() => { channel.BasicAck(message.DeliveryTag, false); }, token);
```
johnsimons: Why do we need the token to be passed in ?

src/NServiceBus.RabbitMQ/MessagePump.cs
Scooletz: That's true. Will change that. Leaving message unacked isn't equal to rejecting that.

src/NServiceBus.RabbitMQ/MessagePump.cs
Scooletz: As the PushMessageToPipe is awaited before, that's not needed. Will update.

Scooletz:
Ping @johnsimons :)

Scooletz:
I've merged the PR to the core. This branch has been rebased and is ready to be merged. You need to update the NServiceBus dependency. Can you finish the merge process? /cc @johnsimons

johnsimons:
All packages should be at the same pre-release version of the core.
It seems you have only updated NServiceBus to 1386, we need to also do NServiceBus.AcceptanceTesting and NServiceBus.AcceptanceTests.Sources

Scooletz:
Oh, I was talking about publishing to MyGet with @timbussmann . It's been said that we publish them only when sth is changed in the two mentioned above. I've published only the core. Shall I publish them all together then?

andreasohlund:
yes

timbussmann:
> All packages should be at the same pre-release version of the core.

I wasn't aware of this guideline either. I told @Scooletz it would be fine to only release core since there were no changes in the testing packages.

Scooletz:
@johnsimon @bording I've pushed a version with updated packages. I've seen that there's ongoing work connected with moving RabbitMQ to version. The build is broken and I'd great if I could get some help as I got stuck with closing this PR.

Scooletz:
@johnsimons @bording Updated the reference to the core to unstable1414 and made the build running. There are three acceptance tests failing. I'd be grateful if you could take a look at these. We can either have a discussion and I'll remove them or you can do it, it's up to you :)

bording:
Two of the failing tests are related to the still in-progress work around handling the scaleout stuff correctly.

The third one appears to be new, and it's not correctly handling running against a transport that doesn't support DTC, so that's going to need an upstream fix.

The commits in this PR have a decent amount of noise in them now with all of the intermediate core updates, so I'm going to clean them up and merge.

Scooletz:
Thank you @bording !

Issue #123, opened 2016-01-28
DirectRoutingTopology "broken" for multiple interfaces


johnsimons:
DirectRoutingTopology should use multiple keys, not one.
It currently can't handle multiple interfaces.

mauroservienti:
should we consider this in scope for V6? if yes, could you please ad the expected label? cheers.

andreasohlund:
:-1: to include in "v6", this has been broken since the dawn of time and the workaround is to pass your own routing topology? ( @johnsimons can you confirm?)

bording:
When using the DirectRoutingTopology, you can pass in your own routing key convention, so in theory an end user could construct their own keys to make their scenario work.

Another workaround is to use the ConventionalRoutingTopology, which is the default. It handles multiple interfaces properly.

Since this has been a problem all along, and not something directly v6-related, it makes sense to hold off on it for now.

We might want to review the docs around this and make sure it explicitly points out this limitation.

bording:
Looking through the issues, I see that this has been brought up before, #11, #56.

Since #56 was around first, I'm going to close this as a dupe.

Issue #122, opened 2016-01-28
Autoack can cause message loss during shutdown


johnsimons:
Task force: @bording @WojcikMike

When auto-acknowledgements are turned on, the server acknowledges the message as soon as it is sent to the client. 

RabbitMQ uses the `BasicQos()` method as its server->client flow control mechanism. You set the `prefechCount` parameter to tell the server how many unacknowledged messages to send the client before it waits to send more. If you've turned on auto-acknowledgements, that means you've effectively turned off flow control, because you never have unacknowledged messages.

This means that those messages build up in-memory in the RabbitMQ client, waiting for us to process them. If we then shut down the endpoint with those messages still unprocessed, they are lost because they've already been acknowledged on the server.

Currently, we turn auto-acknowledgements on whenever an endpoint has been configured to not use transactions (for example, `TransportTransactionMode.None` in v6).

I'm not really sure why auto-acknowledgements has been tied to transaction settings. They don't really feel like related concepts. Based on how acknowledgements are used for flow control, I don't think we should ever turn on auto-acknowledgements.

One additional problem related to using auto-acknowledgements mode is that you can potentially hit an out of memory scenario.

Since the server is sending messages as soon as it gets them without any sort of rate control, if an endpoint is too slow at processing them, the buffer that the RabbitMQ client is storing the messages in could potentially fill up and throw an OutOfMemoryException:

rabbitmq/rabbitmq-dotnet-client#123

----

Original comment

@bording can u expand on this

mauroservienti:
If I recall correctly when autoAck is on messages are acknowledged in the Q as they are picked up and then stored in a in memory Q by the RabbitMQ client waiting to be processed. Given that they are in memory and acknowledged to the broker if the endpoint shuts down or crashes those messages are lost.

mauroservienti:
should we consider this in scope for V6? if yes, could you please ad the expected label? cheers.

andreasohlund:
If it's message loss shouldn't we fix it in the current version + backport?

mauroservienti:
@andreasohlund agree, we should fix it. IIRC the autoAck is turned on, in the actual production version, when the user disables transactions. @bording proposal is to keep if off regardless of the transaction settings.

andreasohlund:
>when the user disables transactions

In that case I'd say we don't need to backport since tx off means that you will loose messages anyway?

But of course we should not have code that contributes further to message loss.

mauroservienti:
[...]

PR #124, opened 2016-02-01
Core v6 update

johnsimons:
Upgrade transport to use NServiceBus v6.
Connects to Particular/NServiceBus#3125

Still left to do:
- [x] Always use ack, autoack can cause message loss, see #122
- [x] Implement callbacks
- [x] Decide whether SecondaryReceiveSettings still needed with current MessagePump: not relevant anymore.
- [x] Remove PrefetchCount and use PushRuntimeSettings.MaxConcurrency for BasicQos()
- [x] Read Circuit breaker settings from config
- [x] `RepeatedFailuresOverTimeCircuitBreaker.Failure()` not actually being called anywhere
- [x] DequeueTimeout setting not used, remove the config option
- [x] Remove HostConfiguration and just use Host
- [ ] Test that inflight messages are ok to use taskid for dictionary keys
- [ ] DirectRoutingTopology "broken": should use multiple keys, not one, can't handle multiple interfaces, see #123
- [ ] Update docs
- [ ] Update samples Particular/docs.particular.net#1011
- [ ] Clean up xml doc comments


andreasohlund:
>Decide whether PrefetchCount config setting should be removed (and use maxconcurrency instead) or keep to override

Isn't PrefectCount = 1 the correct default? (see #100 for full discussion)

bording:
> Isn't PrefectCount = 1 the correct default? (see #100 for full discussion)

With the way the previous version is implemented, setting the PrefetchCount to 1 is likely the right thing ti do (though some perf testing I've been doing shows that a higher value is still beneficial).

However, with the new message pump design, that is no longer true. It uses a single channel/consumer and the PrefetchCount is set to the MaxConcurrency level, so you can control the value through that single setting. Allowing end users to independently set the PrefetchCount doesn't really make sense in the current design.

andreasohlund:
@bording thanks for clarifying!  Can you update and close #100 ?

bording:
While testing some things, I discovered that SendOnly endpoints aren't calling `RabbitMQTransport.Dispose()` when the endpoint is stopped, so the process won't quit properly.

I've filed Particular/NServiceBus#3392 with the info I discovered on the reasons behind it.

bording:
@johnsimons Looking at how DequeueTimeout was used in the previous version, we definitely don't need it anymore, so I removed it. I also made the parser throw an exception if it finds a configuration string that still has it. 

The error message I have there is pretty basic, and could probably use some more tweaking.

Thoughts?

bording:
While this isn't 100% finished yet, the vast majority of the work has been done, and all the acceptance tests are passing.

I'm going to go ahead and merge this to develop, and will continue working on it from there, opening up separate issues for the remaining work items.

PR #126, opened 2016-02-02
Callback using routing spike

SzymonPobiega:
This is a spike based on this core spike https://github.com/Particular/NServiceBus/pull/3374

It removes all the callback support from the RabbitMQ transport. Acceptance tests are green but they don't exercise the callbacks. I will work on a proper demo/test later.


johnsimons:
@SzymonPobiega I don't see anything about creating temporary queues in rabbit ? We need to pass different flags to https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/RabbitMqQueueCreator.cs#L18

SzymonPobiega:
No temp queues in this WIP. For now assuming the callback package is going to generate stable unique IDs for the queues might be permanent.

SzymonPobiega:
Rebased on most recent `update-v6-core` branch. All ATTs are passing. I think this one is good for merging. I've added a test that verifies if the old callback address header is handled properly (via a dedicated feature).

SzymonPobiega:
@Particular/transport-maintainers please take a look.

bording:
I've made a few formatting/naming changes and pushed them up.

I've also uncommented a callbacks test and got it working, though at this point I'm not sure if it makes sense to be testing callbacks at the transport level anymore or not.

This looks good to me, so I'll be merging it into `core-v6-update`.

SzymonPobiega:
Thanks @bording . I forgot about that test. I think this test https://github.com/Particular/NServiceBus.Callbacks/blob/develop/src%2FNServiceBus.Callbacks.AcceptanceTests%2FWhen_using_callbacks_in_a_scaleout.cs covers the same scenario so we will be able to remove that one from Rabbit in future.

PR #127, opened 2016-02-17
Simpify legacy callback support

SzymonPobiega:
* Removed the incoming pipeline behavior
 * Also refactored the callback test to not actually use the callback package (using `RouteReplyToThisInstance` instead) because Callbacks package enforces instance ID being present.


SzymonPobiega:
Wrong target branch.

PR #128, opened 2016-02-17
Simplified legacy callback support

SzymonPobiega:
* Removed the incoming pipeline behavior
 * Refactored `When_scaling_out_senders_that_uses_callbacks` test to use callback-like behavior without the need for callbacks package (which forces providing instance ID everywhere when it is loaded)


SzymonPobiega:
@bording @johnsimons please review.

bording:
:+1:

PR #131, opened 2016-02-26
Release 3.1.0

### LINKED ISSUE #117: Update RabbitMQ.Client to 3.6.1
Heads up , quite a few breaking changes - http://apicomparer.particular.net/compare/rabbitmq.client/3.5.0...3.6.0

andreasohlund:
Closes #117

### DIFF
--- a/src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
+++ b/src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
@@ -12,7 +12,7 @@
 
 
     class RabbitMqDequeueStrategy : IDequeueMessages, IDisposable
-    {   
+    {
         public RabbitMqDequeueStrategy(IManageRabbitMqConnections connectionManager, RepeatedFailuresOverTimeCircuitBreaker circuitBreaker, ReceiveOptions receiveOptions)
         {
             this.connectionManager = connectionManager;
@@ -26,7 +26,7 @@
             this.endProcessMessage = endProcessMessage;
             workQueue = address.Queue;
 
-            noAck = !transactionSettings.IsTransactional;
+            shouldAckBeforeProcessingMessage = !transactionSettings.IsTransactional;
 
             if (receiveOptions.PurgeOnStartup)
             {
@@ -46,7 +46,7 @@
             {
                 actualPrefetchCount = Convert.ToUInt16(maximumConcurrencyLevel);
 
-                Logger.InfoFormat("No prefetch count configured, defaulting to {0} (the configured concurrency level)", actualPrefetchCount);                
+                Logger.InfoFormat("No prefetch count configured, defaulting to {0} (the configured concurrency level)", actualPrefetchCount);
             }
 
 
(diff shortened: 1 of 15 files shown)


andreasohlund:
@Particular/transport-maintainers please review, ok with releasing this as a minor?

SimonCropp:
"ok with releasing this as a minor"

yes

bording:
One thing I'm curious about is how this will interact with the stuff that is on develop now. All of these changes have already been made there.

Once this is merged into master, what is the expectation for these commits in regard to develop?

bording:
Also, do we want to include a fix for #122 in this release?

mauroservienti:
In light of `safe by default`, IMO, yes we should fix #122

andreasohlund:
Yea, lets include #122

andreasohlund:
>One thing I'm curious about is how this will interact with the stuff that is on develop now. All of these changes have already been made there.

>Once this is merged into master, what is the expectation for these commits in regard to develop?

I guess we'll find out once that happens :)

bording:
A fix for #122 has been pushed to the branch.

andreasohlund:
Seems like it broke 2 tests?

src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
```diff
@@ -156,7 +154,7 @@ void ConsumeMessages(object state)
 
                     var consumer = new QueueingBasicConsumer(channel);
 
-                    channel.BasicConsume(parameters.Queue, noAck,receiveOptions.ConsumerTag, consumer);
+                    channel.BasicConsume(parameters.Queue, false, receiveOptions.ConsumerTag, consumer);
```
andreasohlund: Since tx == off means "fire and forget" we need to ack right away "if txmode == none" ?

bording:
I pushed up a revised commit that should address the tests that were failing. I didn't realize that the transport was responsible for requeuing messages in v5 for FLR.

I revised it to look at the `IsTransactional` flag to determine if the `BasicReject` calls should requeue the message or not.

This should provide the desired behavior in the majority of cases. The only time I can think where it would be different would be if the Rabbit connection closed before `BasicReject` could be called. In that case, the message would be requeued since it was not acked or rejected, even if transactions are disabled.

Is that ok? If not, I can revise it further.

SimonCropp:
A new rabbitmq clent was released last night. Does that effect anyhting here?

bording:
I was just about to push that up too. If we're updating the client, makes sense to me to go for the latest.

bording:
Just pushed up another revision to the autoack fix. Now, if `IsTransactional` is false, `BasicAck()` gets called before we try to process the message.

This makes it much less likely that a connection interruption could cause the message to be requeued when we don't want it to be. I'm pretty sure there is still a potential race between calling `BasicAck()` and the client knowing that a channel has been closed, but I'm not sure there is anything we can do about that.

andreasohlund:
Looks good, can you go ahead and release this @bording ?

bording:
3.1.0 has been released and announced on GG.

PR #132, opened 2016-03-03
fix nuspec to allow core unstable

SimonCropp:


SimonCropp:
@Particular/transport-maintainers pretty basic change to allow the unstable core range so we can smoke test samples

Issue #135, opened 2016-03-25
Sending to a non-existent endpoint can have unexpected behavior and message loss


bording:
### Current status

<table>
<tr><th></th><th>Publisher Confirms Enabled</th><th>Publisher  Confirms Disabled</th></tr>
<tr><th>ConventionalTopology</th><td><sup>1</sup> :smile: "Channel has been closed" exception for first bad message</td><td><sup>2</sup> :cry: message loss if there are only a few messages, and the occasional `AlreadyClosedException` when the channel closed out from underneath us at various points</td></tr>
<tr><th>DirectTopology</th><td><sup>3</sup> :smile: "Message could not be routed" for first bad message</td><td><sup>4</sup> :smile: : Warning log entry "Message could not be routed"</td></tr>
</table>

---

When an endpoint sends a message to a non-existent endpoint (or one that has never been started at least once so no exchanges/queues have been created for it yet) the behavior varies based on a couple of factors.
### ConventionalRoutingTopology

When using the default ConventionalRoutingTopology, the behavior you see depends on if publisher confirms are enabled or not.

If publisher confirms are enabled, then you will see a `QueueNotFoundException` because we catch an `AlreadyClosedException` and check to see if the channel was closed because the destination exchange didn't exist.

If publisher confirms are disabled, things get weird. In my testing, I've seen the following things:
- no indication that anything was wrong at all
- an `AlreadyClosedException` gets thrown
- no errors, but when shutting the endpoint down the RabbitMQ client throws an `OperationInterruptedException`

Which one of these you get depends on how many messages you're trying to send in a batch (from inside a handler). If you only send one message, then we don't keep the channel alive long enough for the error/closed message to make it back from the server. If we are reusing the channel to send a batch of messages, then eventually we get the `AlreadyClosedException` once the client catches up to the server sending the error/closed message back.

Sometimes when sending a batch, instead of getting the `AlreadyClosedException`, everything would work fine and instead I'd see the `OperationInterruptedException` when the endpoint shut down. I'm still not exactly sure what's going on with that.
### DirectRoutingTopology

[...]

PR #136, opened 2016-03-28
Implement a ChannelProvider that pools ConfirmsAwareChannels

bording:
Instead of creating and disposing channels for each send operation, the `ChannelProvider` will pool the created channels and reuse them.

Keeping the channels around means that event handlers for `BasicAcks`, `BasicNacks`, `BasicReturn`, and `ModelShutdown` can be registered. Using these events along with `TaskCompletionSource` allows us to preserve the ability to associate a sent message with a confirmation when using publisher confirms, but without having to block a thread and wait for them to come back.

This provides a substantial performance boost for several sending scenarios.

Being able to use the `BasicReturn` event should also go a long way to resolving the problems discussed in #135 also.

Connects to https://github.com/Particular/PlatformDevelopment/issues/719

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationTests.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationTests.cs
@@ -62,12 +62,6 @@
         }
 
         [Test]
-        public void Should_set_default_max_wait_time_for_confirms()
-        {
-            Assert.AreEqual(defaults.MaxWaitTimeForConfirms, TimeSpan.FromSeconds(30));
-        }
-
-        [Test]
         public void Should_set_default_retry_delay()
         {
             Assert.AreEqual(defaults.RetryDelay, TimeSpan.FromSeconds(10));
--- a/src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionStringParserTests.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionStringParserTests.cs
@@ -10,7 +10,7 @@
     {
         const string connectionString =
             "virtualHost=Copa;username=Copa;host=192.168.1.1:1234;password=abc_xyz;port=12345;requestedHeartbeat=3;" +
-            "usePublisherConfirms=true;maxWaitTimeForConfirms=02:03:39;retryDelay=01:02:03;useTls=true;certPath=/path/to/client/keycert.p12;certPassPhrase=abc123";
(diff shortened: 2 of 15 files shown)


bording:
Rebased this to latest develop to make it no longer need unstables that have been removed.

bording:
This is now almost done, though I've got a couple things that would be worth getting some feedback one.

First, I had to add a header to outgoing messages to link them back their `TaskCompletionSource` when a `basic.return` message is received. I'm currently using `ConfirmationId` for this, but is there a better name? Are there guidelines I should be following for this?

Second, the MaxWaitTimeForConfirms config value used to be used as a timeout for how long the thread would block waiting for publisher confirms, but since we don't need to block at all now, I'm not sure this has a purpose anymore. I do have some commented out code in ConfirmsAwareChannel that was trying to use this value to set a per-message timeout, but this was pretty heavy, since it had to delay / create a timer for every single sent message. I saw some pretty decent allocation improvements after commenting it out.

I do also have a bit of debug code in `ConfirmsAwareChannel.GetConfirmationTask()` that still needs to be cleaned up.

Thoughts?

@johnsimons
@mat-mcloughlin

bording:
Oh and the failing tests aren't related. They are flaky tests that need to be fixed in core.

johnsimons:
The header probably should be more qualified, eg. `NServiceBus.RabbitMq.ConfirmationId` ?
Also what happens for backwards compatibility ?

WojcikMike:
I would second NServiceBus.RabbitMq.ConfirmationId or even:
NServiceBus.Transports.RabbitMq.ConfirmationId
Which will make it even more alligned

bording:
> Also what happens for backwards compatibility ?

There shouldn't be any backwards compatibility issues because this header only matters for V4 senders to enable them to get information about a returned message via `basic.return`, which doesn't include the publisher confirm id like `basic.ack` and `basic.nack`.

The receiver of the message doesn't care about the header at all.

I'll change the header value to be more qualified.

bording:
For the header, I decided to go with `NServiceBus.Transport.RabbitMQ.ConfirmationId`. It's a bit long, but it matches the new namespace stuff, so it seemed like the best choice.

Since no one commented on the `MaxWaitTimeForConfirms` stuff, I went ahead and removed it for now. If we decide we need it after all, we can always add it back.

bording:
I believe this is ready to go now.

I've left in a debug check that will throw an exception for now, because I'd like to see if anyone who tries the beta runs into that scenario. It shouldn't happen, but at one point I was seeing it.

johnsimons:
👍 to merge

PR #144, opened 2016-04-28
[WIP] Getting started

colin-higgins:
I wanted to make sure this was up in a pull request so it wasn't lost.

These are the steps I went through with @bording to get this code base running locally.

It still needs the Rabbit as a service option as well but this is a good starting point.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,0 +1,35 @@
+## Setup NServiceBus.RabbitMQ
+
+- Install virtualization software like VirtualBox
+- Create Win10 VM
+ - Name VM something sane
+- Install Erlang
+- Install RabbitMQ
+
+Determine whether the RabbitMQ install worked
+ - Verify that it's available to anything other than localhost
+ - If it didn't work, we need the workaround.. TODO
+	
+Open the rabbit cmd prompt as elevated
+ - Manually navigate to "C:\programfiles\rabbitMQ Server\rabbitmq_server-3.6.1\sbin"
+	
+Turn on RabbitMQ management site at http://localhost:15672
+ - rabbitmq-plugins enable rabbitmq_management
+
+Create the rabbitmq.config file in user appdata ROAMING/RabbitMQ/rabbitmq.config (This will allow you to use the loopback credentials in a connection string)
+ - with this text: "[{rabbit, [{loopback_users, []}]}]."
+
+To force rabbit to use the config file, run the following commands
+ - rabbitmq-service stop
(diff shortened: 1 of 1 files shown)


johnsimons:
I think @distantcam actually uses a docker container for this

distantcam:
@danielmarbach Also added a file on how to get started. https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/Docker.md

adamralph:
Shouldn't both this and https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/Docker.md be in our docs site?

colin-higgins:
I think that makes sense, and we can link to the getting started documents from the README

johnsimons:
> Shouldn't both this and https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/Docker.md be in our docs site?

@adamralph no it should not, this is about setting up for development time.

johnsimons:
@colin-higgins given the docker setup is much simpler and also easier to keep up to date, do you feel we also need this ?

bording:
@johnsimons Docker on windows is still a bit more painful than I'd want to mess around with right now. 

For example, I'm pretty sure it's not compatible with having Hyper-V enabled at the same time.

johnsimons:
@bording Should we test it ?
@distantcam @danielmarbach any issue with docker and hyper-v ?

bording:
@johnsimons The current Docker for Windows installation uses VirtualBox, and you can't have that and Hyper-V enabled at the same time. The newer version that's in beta does move over to using Hyper-V.

adamralph:
@johnsimons 

> no it should not, this is about setting up for development time.

OK, makes sense.

GETTING_STARTED.md
```diff
@@ -0,0 +1,35 @@
+## Setup NServiceBus.RabbitMQ
+
+- Install virtualization software like VirtualBox
+- Create Win10 VM
```
adamralph: Why the need to use a VM? I've done plenty of development against RabbitMQ and I've never had to set up a VM. I just install it locally.

I can see that this may be a valuable optional way to do it, but if we imply that you *have* to use a VM it feels like quite a barrier to entry.

adamralph:
I think the content in this document is useful, but I think that this and docker.md should be consolidated into a single document showing the various options. We should also have the simplest possible option docuemnted, which is to just install RabbitMQ locally, without a VM. This is the way I have always developed against RabbitMQ.

adamralph:
Closing in favour of #160

PR #143, opened 2016-04-28
Changing namespaces

WojcikMike:
Connect to Particular/PlatformDevelopment#719

PR #145, opened 2016-04-29
Change default namespace in projects

WojcikMike:
Connect to Particular/PlatformDevelopment#719

@Particular/rabbitmq-transport-maintainers please review and merge, it is missing step after we did #143


bording:
Good catch!

PR #146, opened 2016-05-03
Override ReplyTo header when legacy callbacks header is present

### LINKED ISSUE #142: Audits/forwards might be routed to wrong destination
RabbitMQ transport for V6 might have a bug similar to this SQL Server transport one https://github.com/Particular/NServiceBus.SqlServer/issues/202

bording:
Following the sql server fix to the same problem over there (https://github.com/Particular/NServiceBus.SqlServer/pull/221) this removes the legacy callback behavior and instead overrides the ReplyTo header whenever the legacy header is present on an incoming message.

Connects to https://github.com/Particular/PlatformDevelopment/issues/719
Closes #142


johnsimons:
looks good, merge it

Issue #150, opened 2016-05-20
Update to Rabbit client 3.6.2


volak:
They added a new interface property that is not implemented in NSB's client and again broke assembly scanning where this library is concerned.  

The only way to fix is to downgrade to 3.6.1 and add a redirect from 3.6.2 to 3.6.1

bording:
@volak I'm hoping to have an updated release out soon that will be built against 3.6.2.

volak:
@bording Thanks - perhaps it might be a good idea to change the nuget package dependencies to allow rabbit client max 3.6.2 too.  This is the second time in my short use of their client they have made a breaking change in a minor version..

bording:
We have considered doing that, but the downside of that change is we then **have** to release a new package every time they update, even if/when there aren't breaking changes and it would otherwise work with a binding redirect.

Given that they like to break whenever they feel like it, perhaps that might be best approach.

bording:
While preparing the 3.3.0 release, I caught a bug in the new RabbitMQ client, so I skipped releasing it and built 3.3.1 instead.

adamralph:
@volak this has been fixed in https://www.nuget.org/packages/NServiceBus.RabbitMQ/3.3.1.

Thanks very much for raising this issue!

PR #151, opened 2016-05-20
Release 3.3.0

bording:
Connects to #150


bording:
 
For now, I've left the `QueueingBasicConsumer` in place, so we are getting deprecation warnings. Do we want to keep this as is or actually go through with swapping it out? At this point I think I'd prefer waiting until the class is actually removed.

I also have only bumped the min nuspec range and left the max alone. I think I prefer this approach, but what does everyone else think?

SimonCropp:
> I also have only bumped the min nuspec range and left the max alone. I think I prefer this approach, but what does everyone else think?

I still think this is preferable. yes rabbit breaks things in patches, but very rarely does that effect the small surface area of the API we use. So not fixing the patch means we dont need to release in lockstep with rabbit. i think that is an acceptable compromise

src/NServiceBus.RabbitMQ/Connection/RabbitMqConnectionFactory.cs
```diff
@@ -44,10 +44,9 @@ public RabbitMqConnectionFactory(ConnectionConfiguration connectionConfiguration
             };
         }
 
-        public virtual IConnection CreateConnection(string purpose)
+        public virtual IConnection CreateConnection(string connectionName)
```
adamralph: I know this is an edge case, but this is potentially breaking if someone is naming the parameter (same applies to the change to `PersistentConnection.ctor()`). Do we have a policy on this? Personally I'm happy to call it non-breaking.

adamralph:
:+1: LGTM. What's left to do?

src/NServiceBus.RabbitMQ/Connection/RabbitMqConnectionFactory.cs
SimonCropp: What is the compelling reason to make this changes? Does it justify breaking semver?

adamralph: [Every public change is a breaking change](https://blogs.msdn.microsoft.com/ericlippert/2012/01/09/every-public-change-is-a-breaking-change/) so there is always a class of theoretical breakages which need to be excluded from a project's definition of breaking, but I second the challenge. @bording what is the reason for making this change?

bording: Maybe I'm missing something here. `RabbitMqConnectionFactory` isn't public, so this isn't a public-facing change.

adamralph: > they're both internal classes

Well, yes, that changes things :laughing: 

OK, ignore all of the above. Sorry for derailing. :grin:

bording:
> :+1: LGTM. What's left to do?

At this point, it should be ready to release. Feel free to go ahead and do that, otherwise I'll get it out tomorrow.

bording:
While preparing this release, I found a bug in the new connection name feature introduced in the RabbitMQ 3.6.2 client (https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/187), so I'm going to create a 3.3.1 release and re-add the "purpose" client property to surface the value for now.

Issue #149, opened 2016-05-18
Remove ConnectionManager


bording:
Currently, we still have both `ConnectionManager` and `ConnectionFactory` classes. As mentioned in https://github.com/Particular/NServiceBus.RabbitMQ/issues/74#issuecomment-201040218, the scope of what `ConnectionManager` is doing has been decreased. At this point it is only being used to manage the publish connection and passing along the creation of the admin connection to the connection factory:

https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Connection/ConnectionManager.cs

Instead of keeping `ConnectionManager` around just to pass it into `ChannelProvider`, I think it makes sense kill `ConnectionManager` and just pass `ConnectionFactory` around directly.

This would mean that `ChannelProvider` would get a `ConnectionFactory` and create its own publish connection when it needs it, and then would be responsible for closing it when the endpoint is stopping.

This approach works because at that point we would only ever need to be "creating" connections and the two places that keep connections open (`ChannelProvider` and `MessagePump`) can just deal with the created connection directly.

The `MessagePump` was originally designed this way because it needed to create a specific `ConnectionFactory` to pass in a custom scheduler, but I've cleaned that up a bit, and it no longer requires the custom scheduler. 550664404c2563d01096cec25cca283e038eb95d

Because of this, we could decide on an alternate approach. The `MessagePump` could once again get a "managed" connection from `ConnectionManager`, and then it would use that connection to create a channel. 

This would put the `ConnectionManager` back in charge of connection lifetimes.

Currently, as a side effect of the `MessagePump` being responsible for creating a connection, each `MessagePump` instance creates its own connection, so each queue being consumed (main, optional instance queue, satellites) has a separate connection in addition to a separate channel. I was able to take advantage of this and the purpose of each connection is set differently instead of having a generic "consume" purpose: https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs#L74

This is nice when viewing the connections from the management UI, and we'd lose it if we went back to the `ConnectionManager` being in charge.

Thoughts? 

@Particular/rabbitmq-transport-maintainers

andreasohlund:
:heart:

PR #152, opened 2016-05-24
[WIP] Remove connection manager

bording:
Connects to #149


bording:
I think this more or less covers all the work needed for replacing the `ConnectionManager` with `ConnectionFactory`, but I want to leave it [WIP] for now until I've had another pass to make sure.

Also, there are a couple of options for how `ChannelProvider` manages its connection. In the first commit, I have it create the connection in the constructor. In the 3rd commit, I changed that so that it will wait until the first time a channel is requested.

The first approach is simpler, but it does mean we'll be opening a publish channel as soon as the endpoint starts vs. waiting for an actual send to happen.

Thoughts?

adamralph:
@bording I made an attempt at reviewing this this morning but I wanted to get my local dev environment up and running (new workstation) so that has delayed me. I'll continue later today.

WojcikMike:
I prefer the second approach of creating connection when it is needed rather than on start.

src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -20,7 +22,12 @@ public ConfirmsAwareChannel GetPublishChannel()
             {
                 channel?.Dispose();
 
-                channel = new ConfirmsAwareChannel(connectionManager.GetPublishConnection(), usePublisherConfirms);
+                if (connection == null)
+                {
+                    CreateConnection();
```
adamralph: A minor point, but here I'd probably prefer that we inline

```C#
lock (connectionFactory)
{
    connection = connection ?? connectionFactory.CreatePublishConnection();
}
```

to remove the side effect from `CreateConnection`.

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -46,11 +47,18 @@ public ConnectionFactory(ConnectionConfiguration connectionConfiguration)
...
         }
 
+        public IConnection CreatePublishConnection() => CreateConnection("Publish");
+
+        public IConnection CreateAdministrationConnection() => CreateConnection("Administration");
+
         public IConnection CreateConnection(string connectionName)
         {
-            connectionFactory.ClientProperties["connected"] = DateTime.Now.ToString("G");
+            lock (lockObject)
```
adamralph: Can we lock on `connectionFactory` instead, removing the need for `lockObject`?

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -46,11 +47,18 @@ public ConnectionFactory(ConnectionConfiguration connectionConfiguration)
             }
         }
 
+        public IConnection CreatePublishConnection() => CreateConnection("Publish");
```
adamralph: This is a change in behaviour since the `ConnectionManager` used to cache the publish connection. Is that a problem?

adamralph:
@bording I did a pass through the changes. Generally, it looks good, and I agree that the lazy connection creation is better.

I left a few questions inline.

I'm also wondering if the lazy connection creation is an observable change in behaviour rather than being part of the #149 refactoring?

src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
bording: Going to switch to `Lazy<T>` instead.

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
bording: The caching is now happening in the `ChannelProvider` so there's no need for a separate class to do this.

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
bording: I could make this change, but I tend to prefer separate lock objects. You never know when some code is going to do something dumb like `lock (this)` and deadlock themselves.

bording:
> I'm also wondering if the lazy connection creation is an observable change in behaviour rather than being part of the #149 refactoring?

As I already mentioned, this is already the way it works inside the ConnectionManager, it's just that my first commit didn't do it that way, so I moved back to it later.

I've also pushed up some changes to address feedback, or commented on the feedback directly.

Since we want this to go against the release branch instead of develop, I'm going to close and create a new PR.

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
adamralph: Fair enough.

src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
adamralph: :+1:

src/NServiceBus.RabbitMQ/Connection/ConnectionFactory.cs
adamralph: Ah yes, true.

PR #153, opened 2016-05-25
Release 3.3.1

bording:
This works around the bug in the 3.6.2 release of the RabbitMQ.Client where the new connection name setting isn't being set properly so it doesn't show up in the management ui.

@Particular/rabbitmq-transport-maintainers If we want to make a change to the nuspec version range to lock in to 3.6.2, we should make the change here before we release 3.3.1.


SimonCropp:
Re the range lock in. If we are still compat with the lower patch versions i would say dont lock in

adamralph:
The @Particular/rabbitmq-transport-maintainers decided not to lock in for now. We'll see things go for a while with RabbitMQ client releases and then revisit if required.

PR #162, opened 2016-06-15
Add new() constraint on RabbitMQTransportSettingsExtensions.UseRoutingTopology type param

adamralph:
Currently, a runtime exception will be thrown if the type being used does not have a parameter-less constructor. This moves this to a compile time error.


WojcikMike:
@adamralph some tests have failed

adamralph:
Ah yes, the API approval. I'll fix.

adamralph:
Amended the commit to contain the API change approval.

Also, I relabelled this as tech debt rather than a refactoring, since it's a public API change.

bording:
:+1:

PR #165, opened 2016-06-21
Optimizations

bording:
This includes several rather small changes that are focused around making a few small tweaks to prevent unnecessary memory allocations.

The first commit is also a refactoring that made sense given how the `ChannelProvider` and `ConfirmsAwareChannel` classes are actually the ones that need the `RoutingTopology`, instead of passing it to the` MessageDispatcher` first.

If there are any changes here that you'd prefer to have split into a separate PR, let me know and I can do that.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
+++ b/src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
@@ -6,13 +6,13 @@
     class QueueCreator : ICreateQueues
     {
         readonly ConnectionFactory connectionFactory;
-        readonly IRoutingTopology topology;
+        readonly IRoutingTopology routingTopology;
         readonly bool durableMessagesEnabled;
 
-        public QueueCreator(ConnectionFactory connectionFactory, IRoutingTopology topology, bool durableMessagesEnabled)
+        public QueueCreator(ConnectionFactory connectionFactory, IRoutingTopology routingTopology, bool durableMessagesEnabled)
         {
             this.connectionFactory = connectionFactory;
-            this.topology = topology;
+            this.routingTopology = routingTopology;
             this.durableMessagesEnabled = durableMessagesEnabled;
         }
 
@@ -38,7 +38,7 @@
             {
                 channel.QueueDeclare(receivingAddress, durableMessagesEnabled, false, false, null);
 
-                topology.Initialize(channel, receivingAddress);
+                routingTopology.Initialize(channel, receivingAddress);
             }
         }
     }
--- a/src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
@@ -31,7 +31,7 @@
 
         public string CertPassphrase { get; set; }
 
-        public IDictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
+        public Dictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
 
         public ConnectionConfiguration(ReadOnlySettings settings)
         {
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -20,10 +20,7 @@
             this.messageIdStrategy = messageIdStrategy;
         }
 
-        public string RetrieveMessageId(BasicDeliverEventArgs message)
-        {
-            return messageIdStrategy(message);
-        }
+        public string RetrieveMessageId(BasicDeliverEventArgs message) => messageIdStrategy(message);
 
         public Dictionary<string, string> RetrieveHeaders(BasicDeliverEventArgs message)
         {
@@ -135,7 +132,6 @@
         }
 
         readonly Func<BasicDeliverEventArgs, string> messageIdStrategy;
-
 
         static ILog Logger = LogManager.GetLogger(typeof(MessageConverter));
     }
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -114,17 +114,15 @@
                 return Encoding.UTF8.GetString(bytes);
             }
 
-            var objects = value as IDictionary<string, object>;
-            if (objects != null)
+            var dictionary = value as IDictionary<string, object>;
+            if (dictionary != null)
             {
-                var dict = objects;
-                return String.Join(",", dict.Select(kvp => kvp.Key + "=" + ValueToString(kvp.Value)));
+                return String.Join(",", dictionary.Select(kvp => kvp.Key + "=" + ValueToString(kvp.Value)));
             }
 
-            var list1 = value as IList;
-            if (list1 != null)
+            var list = value as IList;
+            if (list != null)
             {
-                var list = list1;
                 return String.Join(";", list.Cast<object>().Select(o => ValueToString(o)));
             }
 
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -26,7 +26,7 @@
         {
             var properties = message.BasicProperties;
 
-            var headers = DeserializeHeaders(message);
+            var headers = DeserializeHeaders(properties.Headers);
 
             if (properties.IsReplyToPresent())
             {
@@ -83,21 +83,22 @@
             return properties.MessageId;
         }
 
-        static Dictionary<string, string> DeserializeHeaders(BasicDeliverEventArgs message)
+        static Dictionary<string, string> DeserializeHeaders(IDictionary<string, object> headers)
         {
-            if (message.BasicProperties.Headers == null)
+            var deserializedHeaders = new Dictionary<string, string>();
+
+            if (headers != null)
             {
-                return new Dictionary<string, string>();
+                var messageHeaders = headers as Dictionary<string, object>
+                    ?? new Dictionary<string, object>(headers);
+
+                foreach (var header in messageHeaders)
+                {
+                    deserializedHeaders.Add(header.Key, header.Value == null ? null : ValueToString(header.Value));
+                }
             }
 
-            return message.BasicProperties.Headers
-                .ToDictionary(
-                    dictionaryEntry => dictionaryEntry.Key,
-                    dictionaryEntry =>
-                    {
-                        var value = dictionaryEntry.Value;
-                        return dictionaryEntry.Value == null ? null : ValueToString(value);
-                    });
+            return deserializedHeaders;
         }
 
         static string ValueToString(object value)
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -128,7 +128,7 @@
             if (list1 != null)
             {
                 var list = list1;
-                return String.Join(";", list.Cast<object>().Select(ValueToString));
+                return String.Join(";", list.Cast<object>().Select(o => ValueToString(o)));
             }
 
             return null;
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -50,9 +50,7 @@
             this.pipe = pipe;
             this.settings = settings;
 
-            circuitBreaker = new MessagePumpConnectionFailedCircuitBreaker($"'{settings.InputQueue} MessagePump'",
-                timeToWaitBeforeTriggeringCircuitBreaker,
-                ex => criticalError.Raise($"{settings.InputQueue} MessagePump's connection to the broker has failed.", ex));
+            circuitBreaker = new MessagePumpConnectionFailedCircuitBreaker($"'{settings.InputQueue} MessagePump'", timeToWaitBeforeTriggeringCircuitBreaker, criticalError);
 
             exclusiveScheduler = new ConcurrentExclusiveSchedulerPair().ExclusiveScheduler;
 
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePumpConnectionFailedCircuitBreaker.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePumpConnectionFailedCircuitBreaker.cs
@@ -6,10 +6,10 @@ namespace NServiceBus.Transport.RabbitMQ
 
     class MessagePumpConnectionFailedCircuitBreaker : IDisposable
     {
-        public MessagePumpConnectionFailedCircuitBreaker(string name, TimeSpan timeToWaitBeforeTriggering, Action<Exception> triggerAction)
+        public MessagePumpConnectionFailedCircuitBreaker(string name, TimeSpan timeToWaitBeforeTriggering, CriticalError criticalError)
         {
             this.name = name;
-            this.triggerAction = triggerAction;
+            this.criticalError = criticalError;
             this.timeToWaitBeforeTriggering = timeToWaitBeforeTriggering;
 
             timer = new Timer(CircuitBreakerTriggered);
@@ -50,7 +50,7 @@ namespace NServiceBus.Transport.RabbitMQ
             if (Interlocked.Read(ref failureCount) > 0)
             {
                 Logger.WarnFormat("The circuit breaker for {0} will now be triggered", name);
-                triggerAction(lastException);
+                criticalError.Raise($"{name} connection to the broker has failed.", lastException);
             }
         }
 
@@ -59,7 +59,7 @@ namespace NServiceBus.Transport.RabbitMQ
         string name;
         TimeSpan timeToWaitBeforeTriggering;
         Timer timer;
-        Action<Exception> triggerAction;
+        CriticalError criticalError;
         long failureCount;
         Exception lastException;
     }
--- a/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
+++ b/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
@@ -2,7 +2,6 @@
 {
     using System;
     using System.Collections.Concurrent;
-    using System.Linq;
     using global::RabbitMQ.Client;
     using Transports;
 
@@ -99,8 +98,10 @@
                 baseType = typeToProcess.BaseType;
             }
 
-            foreach (var exchangeName in type.GetInterfaces().Select(ExchangeName))
+            foreach (var interfaceType in type.GetInterfaces())
             {
+                var exchangeName = ExchangeName(interfaceType);
+
                 CreateExchange(channel, exchangeName);
                 channel.ExchangeBind(exchangeName, ExchangeName(type), string.Empty);
             }
--- a/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
+++ b/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
@@ -37,10 +37,10 @@
                 // Make handlers for IEvent handle all events whether they extend IEvent or not
                 type = typeof(object);
             }
+
             SetupTypeSubscriptions(channel, type);
             channel.ExchangeBind(subscriberName, ExchangeName(type), string.Empty);
         }
-
 
         public void TeardownSubscription(IModel channel, Type type, string subscriberName)
         {
@@ -74,16 +74,11 @@
 
         public void Initialize(IModel channel, string mainQueue)
         {
-
             CreateExchange(channel, mainQueue);
             channel.QueueBind(mainQueue, mainQueue, string.Empty);
-
         }
 
-        static string ExchangeName(Type type)
-        {
-            return type.Namespace + ":" + type.Name;
-        }
+        static string ExchangeName(Type type) => type.Namespace + ":" + type.Name;
 
         void SetupTypeSubscriptions(IModel channel, Type type)
         {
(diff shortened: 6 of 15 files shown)


src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -74,42 +74,37 @@ public void RawSendInCaseOfFailure(IModel channel, string address, byte[] body,
...
             CreateExchange(channel, mainQueue);
             channel.QueueBind(mainQueue, mainQueue, string.Empty);
-
         }
 
-        static string ExchangeName(Type type)
-        {
-            return type.Namespace + ":" + type.Name;
-        }
+        static string ExchangeName(Type type) => type.Namespace + ":" + type.Name;
```
WojcikMike: For the ease of reading I would go with string interpolation here, but I am fine with leaving it as is.

adamralph: The problem with string interpolation is that, before .net 4.6, there's no nice way to make it culture invariant. In .net 4.6, we can use [FormattableString.Invariant](https://msdn.microsoft.com/en-us/library/system.formattablestring.invariant(v=vs.110).aspx). Whilst we're on .net 4.5 I think we should avoid string interpolation unless we explicitly want to use the current culture.

A micro-optimisation here would be to use `string.Concat()`, since the current use of two `+` operators forces an extra heap allocation.

src/NServiceBus.RabbitMQ/Sending/MessageDispatcher.cs
```diff
@@ -21,14 +20,17 @@ public Task Dispatch(TransportOperations operations, ContextBag context)
 
             try
             {
-                var tasks = new List<Task>(operations.UnicastTransportOperations.Count() + operations.MulticastTransportOperations.Count());
+                var unicastTransportOperations = (List<UnicastTransportOperation>)operations.UnicastTransportOperations;
+                var multicastTransportOperations = (List<MulticastTransportOperation>)operations.MulticastTransportOperations;
```
adamralph: I don't think we should make this cast. Although we know that *currently* core is passing in a `List<T>`, that might change, and the compiler will never tell us we have a problem. I'd be happy with:

```C#
var unicastTransportOperations = operations.UnicastTransportOperations as List<UnicastTransportOperation> ??
  operations.UnicastTransportOperations.ToList();
```

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
```diff
@@ -6,13 +6,13 @@
     class QueueCreator : ICreateQueues
     {
         readonly ConnectionFactory connectionFactory;
-        readonly IRoutingTopology topology;
+        readonly IRoutingTopology routingTopology;
```
adamralph: Minor point, but I would have preferred to have this rename (and the similar one further down) as a separate commit to reduce the noise around the actual change. No need to amend though.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -128,7 +128,7 @@ static string ValueToString(object value)
             if (list1 != null)
             {
                 var list = list1;
-                return String.Join(";", list.Cast<object>().Select(ValueToString));
+                return String.Join(";", list.Cast<object>().Select(o => ValueToString(o)));
```
adamralph: What is the advantage of using the lambda here?

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
SimonCropp: @adamralph given the concatenation of name and namespace can u explain how interpolation would have culture issues?

src/NServiceBus.RabbitMQ/Receiving/MessagePumpConnectionFailedCircuitBreaker.cs
```diff
@@ -50,7 +50,7 @@ void CircuitBreakerTriggered(object state)
             if (Interlocked.Read(ref failureCount) > 0)
             {
                 Logger.WarnFormat("The circuit breaker for {0} will now be triggered", name);
-                triggerAction(lastException);
+                criticalError.Raise($"{name} connection to the broker has failed.", lastException);
```
adamralph: Do we have a policy on string interpolation? See https://github.com/Particular/NServiceBus.RabbitMQ/pull/165#discussion_r68730535

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
adamralph: @SimonCropp in this case there would be no issue so we could indeed use string interpolation.

The slightly unsettling aspect for me is that in C# 5 we'd never use current culture for things like this. In moving to C# 6, in .NET 4.6 we can simply swap all our current concatenations and invariant `string.Format`s for bling strings wrapped in `Invariant()`, but in .NET 4.5 we're forced to make a decision about current vs invariant culture in each case, since bling string = current culture.

It would be less risky to stay away from bling strings in .NET 4.5 but if we feel the benefit of the syntax outweighs the risk then I'm happy to go with that.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -174,74 +174,60 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
                 }
             }
+            catch (AlreadyClosedException ex)
+            {
+                Logger.Warn($"Attempt to acknowledge/reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
+            }
             catch (Exception ex)
             {
-                Logger.Warn($"Error while attempting to process message {messageId}. The message will be rejected.", ex);
+                Logger.Warn($"Error while attempting to process message {messageId}. The message will be requeued.", ex);
```
adamralph: "rejected" -> "requeued"

Is this a change in behaviour?

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -174,74 +174,60 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
 
-                await RejectMessage(channel, message.DeliveryTag, messageId).ConfigureAwait(false);
+                try
+                {
+                    await RejectMessage(channel, message.DeliveryTag).ConfigureAwait(false);
+                }
+                catch (AlreadyClosedException) { }
```
adamralph: We previously didn't swallow this exception. Is this a change in behaviour?

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -174,74 +174,60 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
-                Logger.Warn($"Attempt to reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
-            }
+                var s = (MessageState)messageState;
+
+                s.Channel.BasicReject(s.DeliveryTag, true);
+
+            }, new MessageState { Channel = channel, DeliveryTag = deliveryTag }, CancellationToken.None, TaskCreationOptions.DenyChildAttach, exclusiveScheduler);
         }
 
         public void Dispose()
```
adamralph: In addition to my inline comments, it's difficult to tell if there may be other changes in behaviour in this commit. I think it's probably best to spin off this one into it's own PR.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -32,12 +32,15 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
...
             {
-                properties.Type = message.Headers[NServiceBus.Headers.EnclosedMessageTypes].Split(new[]
+                var index = messageTypesHeader.IndexOf(',');
+
+                if (index > 0)
                 {
-                    ','
-                }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
+                    properties.Type = messageTypesHeader.Substring(0, index);
+                }
```
adamralph: This looks like a change in behaviour. Previously, if there were no comma in the header, we'd set `properties.Type` to be the entire header. Now, we're only setting it if there is at least one comma.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -31,7 +31,7 @@ class ConnectionConfiguration
 
         public string CertPassphrase { get; set; }
 
-        public IDictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
+        public Dictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
```
adamralph: What is the reason for this change?

We gain the specificity of the type, but we lose the ability to change the concrete type.

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -104,7 +104,7 @@ void SetupTypeSubscriptions(IModel channel, Type type)
                 }
             }
 
-            foreach (var exchangeName in type.GetInterfaces().Select(ExchangeName))
+            foreach (var exchangeName in type.GetInterfaces().Select(t => ExchangeName(t)))
```
adamralph: What is the benefit of using the lambda?

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
SimonCropp: irrelevant for an internal type

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
```diff
@@ -68,9 +36,28 @@ static string GetRoutingKey(Type type, string key = "")
...
+            return result;
+        }
+
+        static bool IsClrType(byte[] a1)
+        {
+            IStructuralEquatable structuralEquatable = a1;
+            return structuralEquatable.Equals(MsPublicKeyToken, StructuralComparisons.StructuralEqualityComparer);
+        }
+
+        static bool IsNServiceBusMarkerInterface(Type type) => type == typeof(IMessage) || type == typeof(ICommand) || type == typeof(IEvent);
```
adamralph: Nitpick: in comparison with the original method above, this spills off the side off the diff and introduces a scroll bar. I'm not sure we have any policy on line length but I'd prefer to drop the body of the expression down to the next line.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
adamralph: Thanks, I missed that the class was internal. (I'm used to always seeing explicit `internal`.)

Agreed, concrete type is fine.

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -74,42 +74,37 @@ public void RawSendInCaseOfFailure(IModel channel, string address, byte[] body,
...
-                    baseType = typeToProcess.BaseType;
-                }
+                CreateExchange(channel, ExchangeName(baseType));
+                channel.ExchangeBind(ExchangeName(baseType), ExchangeName(typeToProcess), string.Empty);
+                typeToProcess = baseType;
+                baseType = typeToProcess.BaseType;
             }
 
-            foreach (var exchangeName in type.GetInterfaces().Select(ExchangeName))
+            foreach (var exchangeName in type.GetInterfaces().Select(t => ExchangeName(t)))
```
adamralph: What's the benefit of using the lambda?

adamralph:
@bording I've reviewed all the commits and left comments inline.

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
bording: >A micro-optimisation here would be to use string.Concat(), since the current use of two + operators forces an extra heap allocation.

This already happens automatically: https://ericlippert.com/2013/06/17/string-concatenation-behind-the-scenes-part-one/

For example, here's the IL for `ExchangeName()`:

```
.method private hidebysig static string ExchangeName(class [mscorlib]System.Type 'type') cil managed
{
    .maxstack 8
    L_0000: ldarg.0 
    L_0001: callvirt instance string [mscorlib]System.Type::get_Namespace()
    L_0006: ldstr ":"
    L_000b: ldarg.0 
    L_000c: callvirt instance string [mscorlib]System.Reflection.MemberInfo::get_Name()
    L_0011: call string [mscorlib]System.String::Concat(string, string, string)
    L_0016: ret 
}
```

bording: > . In moving to C# 6, in .NET 4.6 we can simply swap all our current concatenations and invariant string.Formats for bling strings wrapped in Invariant()

Is this a more general statement? I don't recall seeing any `String.Format` with specified cultures anywhere in our codebases.

src/NServiceBus.RabbitMQ/Sending/MessageDispatcher.cs
bording: Yeah, I was thinking this one would be a bit controversial, for the exact reason you pointed out. I didn't just go for it since the goal here is removing allocations and calling `ToList` would allocate.

However, that allocation means core has changed, and an allocation vs. blowing up is definitely preferable. Gives us time to react to the core change without having to have them perfectly in sync.

Will make this change.

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
bording: I agree, and I do typically do renames in a separate commit, but I missed this one somehow. Didn't seem worth trying to go back and split it out once I noticed it, though.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: The advantage here is that since this is a lambda that doesn't close over anything, the compiler generates code that creates and caches the delegate once and reuses it for every call.

This optimization doesn't happen you pass in a method group, so every call allocates a new delegate. There is an open PR on the Roslyn repo to change this behavior.

src/NServiceBus.RabbitMQ/Receiving/MessagePumpConnectionFailedCircuitBreaker.cs
bording: If we do, I'm not aware of one. From what I've seen, if anything it's "String Interpolation: GOOD".

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: No, because the `RejectMessage` method would either have not called `BasicReject` or would have caught the message.

Given the restructuring of the methods, this was needing to not throw an exception that would have previously been caught

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Not really, given the restructuring of the methods and catch blocks. Previously if you got to that catch, something really weird would have happened, so there's a good chance your connection is already lost (which automatically requeues the message), The `RejectMessage` call used to have separate logging for that case, but it doesn't now. I changed the message to indicate that the message would be requeued because A) we call `BasicReject` and it succeeds, or B) the connection is already closed and requeued anyway.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: :+1:  It would be pretty easy to split this off into it's own PR, and I can make sure its split into a couple commits to make it clearer.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
bording: Strictly speaking, yes this is a change, but it's one that I can't see impacting us, since the EnclosedMessageTypes header contains the AssemblyQualifiedName of the type, which is going to have a comma.

I could tweak this a bit to something like:

```csharp
if (index > -1)
{
    properties.Type = messageTypesHeader.Substring(0, index);
}
else
{
     properties.Type = messageTypesHeader
}
```

which would match match for the case you mention. If the string had a leading comma, the behavior would still be different. 

However, given that we're working with AssemblyQualifiedNames, does that matter?

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
bording: The reason for the change is that enumerating a `Dictionary` is allocation-free, but allocating `IDictionary` isn't.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: With indentation, the line is question is 143 characters long, and doesn't even reach half way on my monitor in Visual Studio. :smile: 

I've never been a huge fan of wrapping stuff to multiple lines just because, especially if it's a single expression.

If we were going to enforce some sort of line length I'd personally prefer it to be in relation to how the code looks in VS than how it shows up on GitHub.

bording:
Ok, I've pushed up some new changes, primarily around reacting to the feedback.
 
I've also removed the message pump commit and am going to push that up as a separate PR.

@adamralph Even though some of my responses to your feedback are referencing "outdated commits", please let know if is anything else you have questions about.

I'm not entirely sure what, if anything, it will make sense to squash here, but I have included a couple fixups that would definitely be worth squashing.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -83,10 +83,9 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
             return properties.MessageId;
         }
 
-        static Dictionary<string, string> DeserializeHeaders(BasicDeliverEventArgs message)
+        static Dictionary<string, string> DeserializeHeaders(IDictionary<string, object> headers)
```
adamralph: A good change for sure, but out of curiosity, how does it "Remove LINQ usage"?

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
adamralph: > This already happens automatically

Interesting, I didn't know that! Up to four operands, apparently.

> Is this a more general statement?

Yep, it's a general statement, very much so. Current culture is often (unknowingly) misused and in most cases, it just happens to work because there's nothing that in the formatting that would be affected. In most other projects I work on, we allow code analysis to tell us when we forget to specify the culture.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: I learnt something again! Thanks.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
adamralph: What if someone is doing native integration by sending to RabbitMQ from a non-NSB endpoint? Could this header potentially be missing a comma in that case?

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
adamralph: Same as previous question, and answer given.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: The problem with that is that everyone's VS is different, but everyone's GitHub is the same.

Anyway, we can continue this (hugely compelling) discussion elsewhere. :smile:

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
adamralph: Answered above.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: `ToDictionary` is a LINQ extension method. I suppose a more precise commit message would have been "Remove ToDictionary usage from DeserializeHeaders"

bording: Also, this specific commit is a an additional fixup to be squashed into 0aacc7240952641d5996ac7b4bc90384392b2e7c I realized this morning that I could improve it a bit further by changing the parameter type, hence the fixup commit.

adamralph: I can't see any use of `ToDictionary()` here.

Anyhow, once this is squashed into it's a moot point.

adamralph:
@bording the changes look good to go!

Regarding squashing, I really like the original list of commits so I wouldn't want to squash any more than those. If it's not too much effort, I'd love for the new fix up commits to be squashed into the commit they are fixing up.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
bording: `BasicPropertiesExtensions.Fill` gets called when NSB is sending a message, so native integration isn't relevant here. For this to not have a comma, core would have to fundamentally change the value being stored in this header.

bording:
Ok, all of the fixups have been squashed. I believe this covers all of the comment raised so far.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
adamralph: Ah, OK, makes sense then.

adamralph:
:+1: awesome work @bording!

PR #166, opened 2016-06-29
Simplify Message pump ack and reject methods

bording:
This is the set of changes originally in #165 related to the `MessagePump` class.

I've split them into two commits, so hopefully that makes it easier to follow the changes.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -214,26 +214,33 @@
             }
         }
 
+        class MessageState
+        {
+            public IModel Channel { get; set; }
+
+            public ulong DeliveryTag { get; set; }
+        }
+
         Task AcknowledgeMessage(IModel channel, ulong deliveryTag)
         {
-            var task = new Task(() =>
+            return Task.Factory.StartNew(state =>
             {
-                channel.BasicAck(deliveryTag, false);
-            });
+                var messageState = (MessageState)state;
 
-            task.Start(exclusiveScheduler);
-            return task;
+                messageState.Channel.BasicAck(messageState.DeliveryTag, false);
+
+            }, new MessageState { Channel = channel, DeliveryTag = deliveryTag }, CancellationToken.None, TaskCreationOptions.DenyChildAttach, exclusiveScheduler);
         }
 
         Task RejectMessage(IModel channel, ulong deliveryTag)
         {
-            var task = new Task(() =>
+            return Task.Factory.StartNew(state =>
             {
-                channel.BasicReject(deliveryTag, true);
-            });
+                var messageState = (MessageState)state;
 
-            task.Start(exclusiveScheduler);
-            return task;
+                messageState.Channel.BasicReject(messageState.DeliveryTag, true);
+
+            }, new MessageState { Channel = channel, DeliveryTag = deliveryTag }, CancellationToken.None, TaskCreationOptions.DenyChildAttach, exclusiveScheduler);
         }
 
         public void Dispose()
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -223,24 +223,24 @@
 
         Task AcknowledgeMessage(IModel channel, ulong deliveryTag)
         {
-            return Task.Factory.StartNew(state =>
+            return TaskEx.StartNew(new MessageState { Channel = channel, DeliveryTag = deliveryTag }, state =>
             {
                 var messageState = (MessageState)state;
 
                 messageState.Channel.BasicAck(messageState.DeliveryTag, false);
 
-            }, new MessageState { Channel = channel, DeliveryTag = deliveryTag }, CancellationToken.None, TaskCreationOptions.DenyChildAttach, exclusiveScheduler);
+            }, exclusiveScheduler);
         }
 
         Task RejectMessage(IModel channel, ulong deliveryTag)
         {
-            return Task.Factory.StartNew(state =>
+            return TaskEx.StartNew(new MessageState { Channel = channel, DeliveryTag = deliveryTag }, state =>
             {
                 var messageState = (MessageState)state;
 
                 messageState.Channel.BasicReject(messageState.DeliveryTag, true);
 
-            }, new MessageState { Channel = channel, DeliveryTag = deliveryTag }, CancellationToken.None, TaskCreationOptions.DenyChildAttach, exclusiveScheduler);
+            }, exclusiveScheduler);
         }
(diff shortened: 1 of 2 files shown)


src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -174,74 +174,53 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
                 }
                 else
                 {
-                    await AcknowledgeMessage(channel, message.DeliveryTag, messageId).ConfigureAwait(false);
+                    await AcknowledgeMessage(channel, message.DeliveryTag).ConfigureAwait(false);
                 }
             }
+            catch (AlreadyClosedException ex)
+            {
+                Logger.Warn($"Attempt to acknowledge/reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
```
adamralph: Does it matter that the log message has changed from explicitly say "acknowledge" or "reject" depending on which operation was attempted to "acknowledge/reject"?

If we care about this, it would be fairly simple to lift a version of this catch into each branch of the `if`.

adamralph: ...or set a variable which tells us which operation was attempted (but that feels kludgey).

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -174,74 +174,53 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
             {
-                Logger.Warn($"Error while attempting to process message {messageId}. The message will be rejected.", ex);
+                Logger.Warn($"Error while attempting to process message {messageId}. The message will be requeued.", ex);
 
-                await RejectMessage(channel, message.DeliveryTag, messageId).ConfigureAwait(false);
+                try
+                {
+                    await RejectMessage(channel, message.DeliveryTag).ConfigureAwait(false);
+                }
+                catch (AlreadyClosedException) { }
```
adamralph: After reading [your response to my initial question](https://github.com/Particular/NServiceBus.RabbitMQ/pull/165#discussion_r68841837) about this I now understand why it's OK to swallow this exception here.

However, in the old version of the code, we would have executed this in this case:

```C#
Logger.Warn($"Attempt to reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
```

Is it worth preserving that? We could also preserve the "rejected" instead of the "requeued" in the log message above.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -201,26 +201,33 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+        }
+
         Task AcknowledgeMessage(IModel channel, ulong deliveryTag)
         {
-            var task = new Task(() =>
+            return Task.Factory.StartNew(messageState =>
             {
-                channel.BasicAck(deliveryTag, false);
-            });
+                var s = (MessageState)messageState;
```
adamralph: Silly nitpick, but as a way of avoiding `s`, I'd probably prefer `state` for the lambda param and `messageState` for the local.

adamralph:
@bording I've reviewed the changes and left a couple of comments.

It's a great simplification. The current incarnations of `AcknowledgeMessage()` and `RejectMessage()` are quite nasty.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: I didn't do that originally since I'd need to add additional try blocks around each call, but you're right that preserving the ack/reject distinction in the log would be useful.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Since we need the catch here anyway, adding the logging back seems like an ok thing to do.

bording:
> The current incarnations of AcknowledgeMessage() and RejectMessage() are quite nasty.

I agree. I was never happy with how they turned out initially.

I've pushed up some fixup commits to address your comments, and I also added some helper methods around using `Task.Factory.StartNew` that I think make using it a bit nicer. Let me know what you think!

I'll squash the fixups once you've reviewed the changes.

src/NServiceBus.RabbitMQ/TaskEx.cs
```diff
@@ -1,10 +1,20 @@
...
         //TODO: remove when we update to 4.6 and can use Task.CompletedTask
         public static readonly Task CompletedTask = Task.FromResult(0);
+
+        public static Task StartNew(object state, Action<object> action) => StartNew(state, action, TaskScheduler.Default);
+
+        public static Task StartNew(object state, Action<object> action, TaskScheduler scheduler) => Task.Factory.StartNew(action, state, CancellationToken.None, TaskCreationOptions.DenyChildAttach, scheduler);
+
+        public static Task StartNew(object state, Func<object, Task> function) => StartNew(state, function, TaskScheduler.Default);
+
+        public static Task StartNew(object state, Func<object, Task> function, TaskScheduler scheduler) => Task.Factory.StartNew(function, state, CancellationToken.None, TaskCreationOptions.DenyChildAttach, scheduler).Unwrap();
```
adamralph: I like the simplification this affords in `MessagePump` but I think only the 2nd overload should exist as it's the only one that's required. The 1st, 3rd and 4th overloads aren't used so I'd call YAGNI on them and remove them. If we find the further overloads are needed later it's trivial to add them.

Also, given the method is only used in `MessagePump`, I'd probably say YAGNI on it being internal and make it a private in that class.

BTW - I think this would also work:

```C#
static Task StartNew<T>(T state, Action<T> action, TaskScheduler scheduler) =>
    Task.Factory.StartNew(
        o => action((T)o), state, CancellationToken.None, TaskCreationOptions.DenyChildAttach, scheduler);
```

It saves having to do the cast in the delegate that's passed into the method.

adamralph:
@bording the fix up commits look spot on :ok_hand: 

I left a comment on the new helper method commit.

src/NServiceBus.RabbitMQ/TaskEx.cs
bording: While the generic version does make the casting nicer, it has a closure (over `action`) so it reintroduces a per-call allocation.

bording:
I went ahead and removed the currently unused helper methods, but I decided to keep the one being used in the `TaskEx` class, because I know there are going to be other places I want to be able to call it once this is merged.

I've squashed the fixup commits. Are they any other commits you think should be combined?

adamralph:
Thanks @bording :ok_hand:

PR #169, opened 2016-07-11
Release Hotfix 3.3.2

### LINKED ISSUE #168: Reply-to address not being set correctly
Bug introduced in https://github.com/Particular/NServiceBus.RabbitMQ/commit/8e526cf3d84ded9f2df31a56b8473e10a6d2670e

When the reply-to address is already set in the message headers and differs from the native (transport specific) reply-to address, we are not setting the address in the headers to the native address. Any current value of the address in the headers is redundant and should be ignored.

In the context of a receiving gateway, the reply-to address in the message headers will be the address of the queue which is local to the sending gateway. With the current behaviour, the reply will fail if the endpoints are on different machines or are using different transports.

The warning should also be removed, since there is nothing to warn of when the address in the message headers differs from the native address. It is normal behaviour to ignore the redundant address in the headers and set it to the native address.

### LINKED ISSUE #173: Transport incorrectly includes machine name when displaying an Address value
While the v3 transport uses the `Address` class in a way that only cares about the `Queue` property of the class, the `Machine` property still gets set to the actual name of the machine the endpoint is running on. This means that in any place that uses `Address.ToString` to show the value of this address, it will incorrectly display "Queue@MachineName" instead of just "Queue".

While in some cases this is relatively harmless, for example in the text of a `QueueNotFoundException`, this can cause problems in some scenarios. 

One specific example is when using ServiceControl to retry failed messages and the endpoint attempting the retry is using v4 of transport. If that handler attempts to reply, the reply will fail because the ReplyToAddress header will have the machine name in the queue.

bording:
This addresses the problem described in #168.

Fixes #168 
Fixes #173

### DIFF
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
@@ -21,6 +21,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -29,6 +30,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
--- a/src/NServiceBus.RabbitMQ/Config/RabbitMqTransportFeature.cs
+++ b/src/NServiceBus.RabbitMQ/Config/RabbitMqTransportFeature.cs
@@ -36,6 +36,8 @@
 
         protected override void Configure(FeatureConfigurationContext context, string connectionString)
         {
+            Address.IgnoreMachineName();
+
             var useCallbackReceiver = context.Settings.Get<bool>(UseCallbackReceiverSettingKey);
             var maxConcurrencyForCallbackReceiver = context.Settings.Get<int>(MaxConcurrencyForCallbackReceiver);
             var queueName = GetLocalAddress(context.Settings);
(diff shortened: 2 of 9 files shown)


src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
```diff
@@ -21,6 +21,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
adamralph: I'm not sure about this. I used to do it in other OSS projects but it creates a lot of friction for contributors. This will fail the build and prevent us from merging PR's which have warnings. I prefer to be able to merge a PR with the odd warning or two and fix them immediately afterwards with another trivial PR of my own.

WilliamBZA: Most of our other repos that I've worked in have this flag set to true, (and when I was creating new projects, @SimonCropp reviewed them and asked for this flag to be true as well) so I assumed it was a standard for us. I think it makes sense to have this as a default anyway. :+1: to treating errors as warnings here.

src/NServiceBus.RabbitMQ/Config/RabbitMqTransportFeature.cs
```diff
@@ -36,6 +36,8 @@ protected override string GetLocalAddress(ReadOnlySettings settings)
 
         protected override void Configure(FeatureConfigurationContext context, string connectionString)
         {
+            Address.IgnoreMachineName();
```
adamralph: As you suggested, I think we should have an explicit issue describing the bug that this fixes.

src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
adamralph: If that's the overall consensus then I'll go with it, but I am opposed to it. As I said, I tried this in other OSS projects and the cost of friction for contributors outweighed the benefit.

For closed source projects I would always set this flag.

adamralph:
@bording I've reviewed the changes and left a couple of comments.

src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
adamralph: I just noticed that on release-4.0.0 we have these flags all set, so the point is moot here, since this will just bring master into line.

PR #171, opened 2016-07-11
Use StartNew helper method in ConfirmsAwareChannel

bording:
This updates `ConfirmsAwareChannel` to use the newly added `TaskEx.StartNew` helper method. An additional overload to avoid needing to pass in the default tasks scheduler was added.


adamralph:
:+1: thanks @bording

PR #176, opened 2016-07-13
Release Hotfix 3.3.3

### LINKED ISSUE #175: Non-default port settings are being ignored
The default port for RabbitMQ connections is 5672. Currently, this is the only port that the transport will attempt to connect to, even when a different port has been configured in the connection string. This also prevents AMQPS connections from working, because these connections need to be established on port 5671.

This problem was introduced when we updated to 3.6.2 of the RabbitMQ client library, which is ignoring all port settings and always using 5672.

bording:
This fixes #175 and also makes sure that tests won't hang with open connections.

### DIFF
--- a/packaging/nuget/NServiceBus.RabbitMQ.nuspec
+++ b/packaging/nuget/NServiceBus.RabbitMQ.nuspec
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[5.0.0, 6.0.0)" />
-      <dependency id="RabbitMQ.Client" version="[3.6.2, 3.7.0)" />
+      <dependency id="RabbitMQ.Client" version="[3.6.3, 3.7.0)" />
     </dependencies>
   </metadata>
   <files>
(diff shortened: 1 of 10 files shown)


SimonCropp:
LGTM

SimonCropp:
I assume it is too much effort to try and write a test for the hang with open connections?

bording:
@SimonCropp The connection hanging is largely only a problem when running the tests. f7ab916 fixes the problems for the admin connections, but that usage was in the test project itself. The actual transport code is already handling that correctly.

The larger issue is actually related how the V5 acceptance test framework works. If a test fails, it doesn't try to stop the endpoint, which means that the connections never get `Dispose` called on them. Since the connection threads are regular threads by default, they keep the process from exiting, and that means the TC agents can never clean the build folder.

It seemed easier to make the connection threads background threads, which you could argue they should be regardless, than to try and get the V5 test framework changed.

During normal operation, the bus shutdown process will call `Dispose` and the connections threads won't keep the process from exiting, so this didn't seem like something that would really be testable.

SimonCropp:
@bording thanks for the clarification.

adamralph:
The test hanging change looks good. I checked the agents and there are no nunit processes leftover :tada:

PR #181, opened 2016-07-15
Upgrade to NServiceBus 6.0.0-beta0006

adamralph:
Fixes #174

Issue #174, opened 2016-07-13
Upgrade to NServiceBus 6.0.0-beta0006


adamralph:
This beta contains breaking changes wrt recoverability and routing.

Connects to https://github.com/Particular/V6Launch/issues/69
Connects to https://github.com/Particular/NServiceBus/pull/3956

adamralph:
Currently there is no such package published so to get a WIP PR going we'll have to build core at the head of release-6.0.0 and install from a local source.

adamralph:
As part of this change we *may* choose to change the behaviour of FLR's to perform them locally instead of rejecting the message and letting the queue deliver the message to any one of the potentially competing consumers. Initial WIP will probably preserve the current behaviour until we can thrash out the pros/cons of changing the FLR behaviour.

adamralph:
I took the maintainer prio and other related labels off this - they didn't make sense, it's more than a size S, and it's actually a subtask of https://github.com/Particular/V6Launch/issues/69 (and, strictly speaking, whatever issue(s) resulted in the other breaking changes in beta6)

PR #182, opened 2016-07-19
Refactor message pump

adamralph:
Some idle refactorings which crept into place whilst I was trying to figure out how to tackle #181.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -176,7 +176,7 @@
                 {
                     try
                     {
-                        await channel.RejectMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
+                        await channel.BasicRejectAndRequeue(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                     }
                     catch (AlreadyClosedException ex)
                     {
@@ -187,7 +187,7 @@
                 {
                     try
                     {
-                        await channel.AcknowledgeMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
+                        await channel.BasicAckSingle(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                     }
                     catch (AlreadyClosedException ex)
                     {
@@ -201,7 +201,7 @@
 
                 try
                 {
-                    await channel.RejectMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
+                    await channel.BasicRejectAndRequeue(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                 }
                 catch (AlreadyClosedException ex2)
                 {
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -187,7 +187,7 @@
             }
             catch (Exception ex)
             {
-                Logger.Warn($"Error while attempting to process message {messageId}. The message will be rejected.", ex);
+                Logger.Warn($"Error while attempting to process message {messageId}. The message will be requeued.", ex);
                 await Requeue(message.DeliveryTag, messageId).ConfigureAwait(false);
             }
         }
@@ -210,9 +210,8 @@
             {
                 await consumer.Model.BasicRejectAndRequeue(deliveryTag, exclusiveScheduler).ConfigureAwait(false);
             }
-            catch (AlreadyClosedException ex)
+            catch (AlreadyClosedException)
             {
-                Logger.Warn($"Attempt to reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
             }
         }
 
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -155,37 +155,34 @@
 
             try
             {
-                Dictionary<string, string> headers = null;
-                var pushMessage = false;
+                Dictionary<string, string> headers;
+
                 try
                 {
                     messageId = messageConverter.RetrieveMessageId(message);
                     headers = messageConverter.RetrieveHeaders(message);
-                    pushMessage = true;
                 }
                 catch (Exception ex)
                 {
                     await poisonMessageForwarder.ForwardPoisonMessageToErrorQueue(message, ex, settings.ErrorQueue).ConfigureAwait(false);
-                }
-
-                var rejectMessage = false;
-                if (pushMessage)
-                {
-                    using (var tokenSource = new CancellationTokenSource())
-                    {
-                        var pushContext = new PushContext(messageId, headers, new MemoryStream(message.Body ?? new byte[0]), new TransportTransaction(), tokenSource, new ContextBag());
-                        await pipe(pushContext).ConfigureAwait(false);
-                        rejectMessage = tokenSource.IsCancellationRequested;
-                    }
-                }
-
-                if (rejectMessage)
-                {
-                    await Requeue(message.DeliveryTag, messageId).ConfigureAwait(false);
-                }
-                else
-                {
                     await Acknowledge(message.DeliveryTag, messageId).ConfigureAwait(false);
+
+                    return;
+                }
+
+                using (var tokenSource = new CancellationTokenSource())
+                {
+                    var pushContext = new PushContext(messageId, headers, new MemoryStream(message.Body ?? new byte[0]), new TransportTransaction(), tokenSource, new ContextBag());
+                    await pipe(pushContext).ConfigureAwait(false);
+
+                    if (tokenSource.IsCancellationRequested)
+                    {
+                        await Requeue(message.DeliveryTag, messageId).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        await Acknowledge(message.DeliveryTag, messageId).ConfigureAwait(false);
+                    }
                 }
             }
             catch (Exception ex)
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -176,7 +176,7 @@
                 {
                     try
                     {
-                        await RejectMessage(channel, message.DeliveryTag).ConfigureAwait(false);
+                        await channel.RejectMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                     }
                     catch (AlreadyClosedException ex)
                     {
@@ -187,7 +187,7 @@
                 {
                     try
                     {
-                        await AcknowledgeMessage(channel, message.DeliveryTag).ConfigureAwait(false);
+                        await channel.AcknowledgeMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                     }
                     catch (AlreadyClosedException ex)
                     {
@@ -201,7 +201,7 @@
 
                 try
                 {
-                    await RejectMessage(channel, message.DeliveryTag).ConfigureAwait(false);
+                    await channel.RejectMessage(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
                 }
                 catch (AlreadyClosedException ex2)
                 {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.RabbitMQ/Receiving/ModelExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            public ulong DeliveryTag { get; set; }
+        }
+
+        public static Task BasicAck(this IModel channel, ulong deliveryTag, bool multiple, TaskScheduler scheduler) =>
+            TaskEx.StartNew(
+                new MessageState { Channel = channel, DeliveryTag = deliveryTag },
+                state =>
+                {
+                    var messageState = (MessageState)state;
+                    messageState.Channel.BasicAck(messageState.DeliveryTag, multiple);
```
bording: Using `multiple` directly here means you've re-introduced a closure. We need to add an additional property to `MessageState` to pass it in, too.

Or, we could drop the `multiple` parameter and hard-code the `false` again.

src/NServiceBus.RabbitMQ/Receiving/ModelExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+                },
+                scheduler);
+
+        public static Task BasicReject(this IModel channel, ulong deliveryTag, bool requeue, TaskScheduler scheduler) =>
```
bording: Same problem with `requeue`.

bording:
Other than the above comments, this looks good! I do like how it simplifies the message pump code.

adamralph:
@bording thanks for the review. I just knew I'd missed a closure somewhere! One day I'll learn to spot these myself :wink: 

My thinking around having those Booleans passed into the extensions was that the extensions would then be pure async wrappers around `IModel.BasicAck` and `IModel.BasicReject`.

I can either alter `MessageState` as you suggest, or I can hardcode the values in the extension methods. I don't really mind which. If the latter, perhaps I ought to rename the extensions to something like `BasicAckSingle` and `BasicRejectAndRequeue`?

adamralph:
@bording I've pushed a WIP commit which removes the closures and hardcodes the Booleans. This seems like a more sensible approach than exposing a parameter in our internal extension methods which we will only ever set to one possible value.

If you're happy with the current diff, I propose that I squash commits 2, 3 and 5 into a single commit "refactor: ModelExtensions".

bording:
Your change looks good! Removing the parameter is the approach I prefer, also. Squashing those commits sounds good too.

> I just knew I'd missed a closure somewhere! One day I'll learn to spot these myself 

While I did spot this instance directly, because I was reviewing specifically for it, take a look at https://visualstudiogallery.msdn.microsoft.com/f9b47b93-8675-4ae0-9c52-5da8027c4bb8. It will point these cases out directly in Visual Studio!

adamralph:
@bording thanks for the extension tip. Installed!

I've squashed the commits as discussed.

adamralph:
@bording I added two more commits which I think make `MessagePump.ProcessMessage` easier to read.

The first commit teases out the concurrency throttling using the semaphore into `Consumer_Received` and also extracts the trying of ack/reject with warning logs into methods. This reduces the method from 84 lines to 44 lines.

The second commit simplifies the flow control by replacing the setting of Booleans with the calling of the TryAck/TryReject methods and an early return when we realise the message is poisoned. This further reduces the method length from 44 lines to 37.

For me, these commits leave the method far more readable and it makes it much easier to see what I need to do to solve #174.

Let me know what you think.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+            {
+                await ProcessMessage(eventArgs).ConfigureAwait(false);
+            }
+            finally
+            {
+                semaphore.Release();
+            }
+        }
+
+        private async Task ProcessMessage(BasicDeliverEventArgs message)
```
bording: Need to remove `private` to pass inspections.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
-
-        class MessageState
-        {
-            public IModel Channel { get; set; }
-
-            public ulong DeliveryTag { get; set; }
         }
 
-        Task AcknowledgeMessage(IModel channel, ulong deliveryTag)
+        private async Task TryAcknowledgeMessage(ulong deliveryTag, string messageId)
```
bording: Need to remove `private` to pass inspections.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                await consumer.Model.BasicAckSingle(deliveryTag, exclusiveScheduler).ConfigureAwait(false);
+            }
+            catch (AlreadyClosedException ex)
+            {
+                Logger.Warn($"Attempt to acknowledge message {messageId} failed because the channel was closed. The message will be requeued.", ex);
+            }
         }
 
-        Task RejectMessage(IModel channel, ulong deliveryTag)
+        private async Task TryRejectMessage(ulong deliveryTag, string messageId)
```
bording: Need to remove `private` to pass inspections.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+            finally
+            {
+                semaphore.Release();
+            }
+        }
+
+        private async Task ProcessMessage(BasicDeliverEventArgs message)
+        {
             Dictionary<string, string> headers = null;
             string messageId = null;
```
bording: Formatting: I'd prefer to have an empty line between the declarations and the `try`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
                 {
                     messageId = messageConverter.RetrieveMessageId(message);
                     headers = messageConverter.RetrieveHeaders(message);
-
-                    pushMessage = true;
                 }
                 catch (Exception ex)
                 {
                     await poisonMessageForwarder.ForwardPoisonMessageToErrorQueue(message, ex, settings.ErrorQueue).ConfigureAwait(false);
+                    await TryAcknowledgeMessage(message.DeliveryTag, messageId).ConfigureAwait(false);
```
bording: Formatting: I'd prefer to have an empty line between the `TryAcknowledgeMessage` call and the `return`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -140,107 +139,77 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
-                        Logger.Warn($"Attempt to reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);
-                    }
-                }
-                else
+                using (var tokenSource = new CancellationTokenSource())
                 {
-                    try
+                    var pushContext = new PushContext(messageId, headers, new MemoryStream(message.Body ?? new byte[0]), new TransportTransaction(), tokenSource, new ContextBag());
+                    await pipe(pushContext).ConfigureAwait(false);
+                    if (tokenSource.IsCancellationRequested)
```
bording: Formatting: I'd prefer to have an empty line between the `pipe` call and the `if`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Also, I tend to prefer to leave the "Try*" naming scheme for methods that actually return a bool and have an out parameter, since that is the established pattern for those in the framework.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Same comment on "Try" here as well.

bording:
Formatting/naming nitpicks aside, I think these additional changes are good ones. I never really liked how the pushMessage/rejectMessage control flow worked, so this is definitely an improvement!

I suppose there's a slight additional overhead around introducing new async methods (Ack/Reject) compared to the previous approach, but I'm not sure that's worth worrying about compared to the overall improvement this brings.

adamralph:
@bording I agree that `Try...` methods should return a Boolean. (I think whether or not an `out` parameter is required depends on whether the non-try equivalent would have a return value.) I've removed `Try` from the methods names and also removed `Message` for brevity since I believe the surrounding context tells us that these are acks/rejects of messages. I also decided to rename `Reject` to `Requeue` to better convey the semantics of what the method does. The fact that we have to reject and requeue is a Rabbit detail which I think is better wrapped up inside that method.

I made the other tweaks you suggested and fixed the code inspections. When I squash the commits I'll alter the commit message to remove "Try" from the method names.

Incidentally, it wasn't the redundant modifiers that were causing the inspections to fail (the rule is configured as a warning only). It was the fact that `headers` was declared with too large a scope. The funny thing is that I don't believe these changes affect that. Perhaps there's a bug in the inspections. Anyway, the variable is declared where it should be now.

adamralph:
BTW - the rename of 'Reject' to 'Requeue' reveals issues in our warning log messages, i.e.

> Logger.Warn($"Error while attempting to process message {messageId}. The message will be rejected.", ex);

This ought to say "will be requeued".

> Logger.Warn($"Attempt to reject message {messageId} failed because the channel was closed. The message will be requeued.", ex);

This ought to say "Attempt to requeue" but then we'd have to alter "will be requeued regardless". This then brings into question the log level. If the message is going to be requeued anyway, does it merit a warning or should we reduce to info or debug? Or just remove the message altogether?

bording:
>  (I think whether or not an out parameter is required depends on whether the non-try equivalent would have a return value

Good point. You usually see this in the framework where the methods are returning a value, so that's probably why I associate them.

>Incidentally, it wasn't the redundant modifiers that were causing the inspections to fail (that rule is configured as a warning only). It was the fact that headers was declared with too large a scope.

Ah, I glanced at the new inspection problems, and didn't notice the headers entry, so assumed it was just the unneeded modifier. Not sure why that suddenly was a problem, either.

I'm :+1: for the method name changes. 

Regarding the logging, changing "The message will be rejected." to "The message will be requeued." makes sense to me.

For the warning in the `Requeue` method, I'm inclined to just remove it altogether. If the reject fails because the channel was closed, the message is requeued anyway, so is that really a failure scenario? I originally had the warning there just to be thorough, but it does seem unnecessary. 

Also, the fact that the channel has closed will have already been reported in the logs via the circuitbreaker.

adamralph:
@bording I've squashed as discussed and pushed another commit which makes the logging changes.

A couple of things:

1. When failing to reject and requeue because the channel is closed, we can't reliably say "*will* be requeued" because it may (perhaps almost certainly will) appear in the broker as already in back in the queue at that point. Perhaps another end point has even already consumed it. Perhaps we should say "will have been requeued"? I realise this getting nitpicky, but if someone were to do some forensics on log timestamps it would look odd if we said "*will* be requeued" after it was *was* requeued.
2. Do we consider the change of logging to still count as a refactoring or do we treat this as a first class change of behaviour, i.e. with 4.0 issue describing it. If the latter, we can raise a small issue for it and close it with this PR.

/cc @Particular/rabbitmq-transport-maintainers

bording:
>When failing to reject and requeue because the channel is closed, we can't reliably say "will be requeued" because it may (perhaps almost certainly will) appear in the broker as already in back in the queue at that point.

What if we added another catch block specifically for `AlreadyClosedException`, so we know that this has happened. We can log it with a different messages, and should most likely be able to skip the `Requeue` call.

> Do we consider the change of logging to still count as a refactoring or do we treat this as a first class change of behaviour

Seeing how the structure of the message pump (and the logging that goes along with it) has already changed compared to 3.x, I'd say that changing it again now is safe to consider a refactoring. Beta users would be the only ones that would observe the change.

bording:
We can address the logging concerns later. Merging what we have now.

PR #188, opened 2016-07-26
Refactor message pump

adamralph:
Here's yet another take on simplifying the message pump (or at least making it easier to read).

I've flattened out the methods to reduce the nesting, and also separated the parsing of the ID and headers.

The separation of the parsing ensures we have the "best possible" log message at any time. I.e. when we don't have the message ID, use the delivery tag for logging and when we *do* have the message ID, use that for logging. (Previously the catch over both the ID and header parsing used the message ID regardless).

There is some repetition between the overloads of methods with and without `messageID` but at least the methods are close together and very similar. This refactoring results in more LOC, but IMHO it's more readable. When you start from `Consumer_Received` you can read down the call stack quite easily and each method is relatively simple.

I'm very interested to know what people think about this.

Again, this refactoring came out of trying to figure out how to address #174


src/NServiceBus.RabbitMQ/Sending/PoisonMessageForwarder.cs
```diff
@@ -1,37 +1,25 @@
...
-
         readonly IChannelProvider channelProvider;
 
         public PoisonMessageForwarder(IChannelProvider channelProvider)
         {
             this.channelProvider = channelProvider;
         }
 
-        public Task ForwardPoisonMessageToErrorQueue(BasicDeliverEventArgs message, Exception ex, string errorQueue)
+        public Task ForwardPoisonMessageToErrorQueue(BasicDeliverEventArgs message, string errorQueue)
```
adamralph: The more I look at this the more I think we should drop this class and take a dep on `IChannelProvider` in `MessagePump`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,54 +141,98 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
         {
-            string messageId = null;
+            string messageId;
+            try
+            {
+                messageId = messageConverter.RetrieveMessageId(message);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error($"Failed to retreive ID from poison message. Moving message to queue '{settings.ErrorQueue}'...", ex);
```
adamralph: I went with "Moving" over "Forwarding" since we tend to favour that language, e.g.

> Our message-processing infrastructure must *move* these poison messages to the side...

>...it would make sense to create a single centralized error queue to which all poison messages could be *moved*.

<sup>http://particular.net/blog/but-all-my-errors-are-severe</sup>

> This sort of failure case is best handled by having the messaging infrastructure just *move* the message to an error queue.

<sup>http://udidahan.com/2007/02/15/its-poison-i-tell-you-poison/</sup>

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,54 +141,98 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+            try
+            {
+                await poisonMessageForwarder.ForwardPoisonMessageToErrorQueue(message, settings.ErrorQueue).ConfigureAwait(false);
             }
             catch (Exception ex)
             {
-                Logger.Warn($"Error while attempting to process message {messageId}. The message will be requeued.", ex);
-                await Requeue(message.DeliveryTag, messageId).ConfigureAwait(false);
+                Logger.Error($"Failed to move message{(messageId == null ? "" : " '" + messageId + "'")} to queue '{queue}'. Returning message to original queue...", ex);
+                await ReturnToQueue(message.DeliveryTag).ConfigureAwait(false);
```
danielmarbach: Are you sure this is the right behavior? This could lead to indefinite retries. Would it make more sense to trigger critical error or return to queue and arm the circuit breaker?

bording: The most likely reason for the poison message to fail to move to the error queue would be if the connection to the broker was lost, which means the message will have already been requeued anyway. That also means the circuit breaker would already be armed as well.

Given that, I think keeping the endpoint up and letting it attempt to reconnect to the broker (disarming the circuit breaker on success) and retry the poison message makes more sense than triggering a critical error and shutting it down.

adamralph: Bear in mind that this PR doesn't change any behaviour. It's a pure refactoring. (Logging changes aside.)

src/NServiceBus.RabbitMQ/Sending/PoisonMessageForwarder.cs
bording: On the other hand, I'd like to see if there is a way to move more of the poison message stuff back into this class instead of having it all in the message pump.

Once the message pump has found a poison message, I'd like to be able to just hand it off to something to take care of it.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: That works for me.

bording:
I'm going to need to spend some more time looking at this one, but I've run out of time for today.

Overall, these changes aren't sitting as well with me vs. the changes in #182.  I think it feels like it's been pulled apart a bit too much just for the sake of logging.

adamralph:
@bording take a look at the latest WIP commit.

I've collapsed the two `ProcessMessage` methods into one, and reverted to having a `try` around both the ID and header extraction. This results in less code, but it does mean that the log message can no longer tell us whether it is the ID or the headers that are problematic. Personally, I'm leaning to preferring these extractions to be separated with separate log messages. It seems a shame to conflate the two when we don't have to. Of course, the other way to collapse the two methods into one would be to nest the `try` blocks. I can do that if you prefer.

That only really leaves the `Move()` method. I can see the appeal of

> Once the message pump has found a poison message, I'd like to be able to just hand it off to something to take care of it.

but in order to do that we'd have to pass either the `EventingBasicConsumer` or the `IModel` to that something, since in the Rabbit world our "move to another queue" is actually "raw send to another queue and then ack on the original queue and if the raw send fails, return to original queue". It seems cleaner to keep the "ack on the original queue and if the send fails, return to original queue" inside the message pump, which leaves that something else to only do the raw send to other queue.

If the argument holds of keeping "ack on the original queue and if the send fails, return to original queue" inside the message pump then we could leave the current `poisonMessageForwarder` in place. It's so trivial that I'd prefer to remove it, but you feel it should stay then I can also live with that. I'd probably argue for a rename though, since it's only responsibility is "raw send a message to a specified queue".

If that argument doesn't hold, then we could indeed pass the `IModel` to the `poisonMessageForwarder` and let it do the ack/return as well as the raw send. Again, I'd argue for a rename, since it's responsibility would then be "move a message to specified queue".

I don't think there's really much else that could be collapsed back together. `Acknowledge()` only has one reference, so we could inline that.

adamralph:
@bording I've pushed yet another commit which I think take a more sane approach to exceptions. It splits the ID and header parsing back out, but inlines some other stuff and results in much less method delegation. It also removes `PoisonMessageForwarder` as an experiment.

adamralph:
Inspection error:

> MessagePump.cs (1)
> 159: Value assigned is not used in any execution path

I'm amend the commit.

src/NServiceBus.RabbitMQ/Receiving/ModelExtensions.cs
```diff
@@ -22,13 +23,22 @@ class MessageState
...
 
-        public static Task BasicRejectAndRequeue(this IModel channel, ulong deliveryTag, TaskScheduler scheduler) =>
+        public static Task BasicRejectAndRequeueIfOpen(this IModel channel, ulong deliveryTag, TaskScheduler scheduler) =>
             TaskEx.StartNew(
                 new MessageState { Channel = channel, DeliveryTag = deliveryTag },
                 state =>
                 {
                     var messageState = (MessageState)state;
-                    messageState.Channel.BasicReject(messageState.DeliveryTag, true);
+                    if (channel.IsOpen)
```
bording: This needs to be `if (messageState.Channel.IsOpen)` to avoid capturing `channel`.

While you're there, could you add an empty line before the `if`?

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Just noticed a typo here: should be "retrieve" not "retreive"

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,77 +141,112 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
 
             try
             {
-                Dictionary<string, string> headers;
+                headers = messageConverter.RetrieveHeaders(message);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error($"Failed to retreive headers from poison message {messageId}. Moving message to queue '{settings.ErrorQueue}'...", ex);
+                await Move(message, settings.ErrorQueue).ConfigureAwait(false);
```
bording: This call to `Move` should be passing in `messageId`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,77 +141,112 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+            Dictionary<string, string> headers;
 
             try
             {
-                Dictionary<string, string> headers;
+                headers = messageConverter.RetrieveHeaders(message);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error($"Failed to retreive headers from poison message {messageId}. Moving message to queue '{settings.ErrorQueue}'...", ex);
```
bording: Same typo here.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Also, single quotes around `messageId`

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,77 +141,112 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                finally
+                {
+                    channelProvider.ReturnPublishChannel(channel);
+                }
             }
-            catch (AlreadyClosedException ex)
+            catch (Exception ex)
             {
-                Logger.Warn($"Attempt to acknowledge message {messageId} failed because the channel was closed. The message will be requeued.", ex);
+                Logger.Error($"Failed to move message{(messageId == null ? "" : " '" + messageId + "'")} to queue '{queue}'. Returning message to original queue...", ex);
```
bording: Should this say "failed to move poison message" to make it clearer what type of message we're dealing with at this point?

bording: Also, what about using `{(messageId == null ? "" : $" '{messageId}'")}` instead for the message id bit?

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -141,77 +141,112 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
-        {
             try
             {
-                await consumer.Model.BasicRejectAndRequeue(deliveryTag, exclusiveScheduler).ConfigureAwait(false);
+                await consumer.Model.BasicAckSingle(message.DeliveryTag, exclusiveScheduler).ConfigureAwait(false);
             }
-            catch (AlreadyClosedException)
+            catch (AlreadyClosedException ex)
             {
+                Logger.Warn($"Failed to acknowledge message{(messageId == null ? "" : " '" + messageId + "'")} because the channel was closed. The message was sent to queue '{queue}' but also returned to the original queue.", ex);
```
bording: Same comments for this message as well.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
adamralph: > Should this say "failed to move poison message" to make it clearer what type of message we're dealing with at this point?

It could, but then I'd probably want to call the method `MovePoisonMessage` so that the log message makes sense in the context of the caller.

> Also, what about using {(messageId == null ? "" : $" '{messageId}'")} instead for the message id bit?

I like it! I didn't realise I could do a nested bling string.

bording: > It could, but then I'd probably want to call the method MovePoisonMessage so that the log message makes sense in the context of the caller.

I'm ok with changing the name of the method because I don't see us using it for anything other than poison messages. It calls `RawSendInCaseOfFailure`, so I can't see how we'd use it for anything else.

bording:
Looks good! Once you squash, I'll merge it.

adamralph:
Squashed.

PR #190, opened 2016-07-28
Upgrade to NServiceBus 6.0.0-beta0006

### LINKED ISSUE #174: Upgrade to NServiceBus 6.0.0-beta0006

bording:
This takes what was in #181 and fixes a few additional things, currently left as fixup commits.

It also moves the TransportTests out of the AcceptanceTests project into a new project. The project name follows the new naming convention that we should try and adjust the others to at some point.

Fixes #174


adamralph:
@bording the fixups LGTM.

The changes in "fixup! red: fixed compilation and inspection errors" are not strictly required to "fix compilation and inspection errors" but I guess they are natural changes to make while doing that. (The removal of the null coalescing in the operations variable assignments is a genuine fixup - I should have done that in the original commit.)

In the last commit (adding the new test project), am I correct in understanding that it consists of only the following?

* Add the project
* Install the transports source tests
* Add `ConfigureRabbitMQTransportInfrastructure`

If so, and if I were to be really nitpicky, I'd ask for the addition of `ConfigureRabbitMQTransportInfrastructure` to be in a separate commit but I think it's OK to leave as is.

I have one reservation about the transport tests source package (although please correct me if I'm misunderstanding it) - the introduction of a dependency on the "Transport.UseSpecific" environment variable being set in order to run the tests. This is why I munged the `NServiceBusTransportTest` class in #181. IMHO this approach is not contributor friendly. It adds ceremony to cloning and running the tests and it also means I have to juggle my environment if I want to run the tests for two different transports.

I think, at the least, we should introduce a "how to build" in the readme which describes the required environment but this can be done in a separate PR. More generally, I'm happy for us to go with the standard approach for now but I'd like to revisit this later to reduce contributor friction.

I'll be happy to merge after the fixups are squashed.

bording:
> The changes in "fixup! red: fixed compilation and inspection errors" are not strictly required to "fix compilation and inspection errors" but I guess they are natural changes to make while doing that. 

Yeah, I was on the fence as to whether I should keep that as a fixup or a separate commit. Happy to keep it separate if you prefer.

> In the last commit (adding the new test project), am I correct in understanding that it consists of only the following?

"Add the project" did also include things like add the appropriate references, etc.

>  have one reservation about the transport tests source package (although please correct me if I'm misunderstanding it) - the introduction of a dependency on the "Transport.UseSpecific" environment variable being set in order to run the tests. 

I'm not really happy with that either, but to me the appropriate place to fix it is upstream. I consider everything in the App_Packages folders 'Not To Be Touched" unless there is a temporary workaround required until the fix comes in from a package update.

The acceptance test framework has some more involved logic around choosing a transport, so you only have to go to that environment variable for more advanced scenarios. I think I'll try and raise a PR to bring that over to the transport tests as well.

In the mean time, I'd be ok with making a change that avoids needing the environment variable set.

bording:
Also, I just remembered that we need to bump the nuspec version.

adamralph:
> Yeah, I was on the fence as to whether I should keep that as a fixup or a separate commit. Happy to keep it separate if you prefer.

Up to you, but I don't think it's worth the effort. I'll merge either way.

> I'm not really happy with that either, but to me the appropriate place to fix it is upstream. I consider everything in the App_Packages folders 'Not To Be Touched" unless there is a temporary workaround required until the fix comes in from a package update.

Agreed.

> In the mean time, I'd be ok with making a change that avoids needing the environment variable set.

I don't think we have to do it now. It's probably more important to get this merged and get the new beta package ready to go out.

> Also, I just remembered that we need to bump the nuspec version.

Good spot. Once that's done, and the current commits are squashed, I'm happy to merge.

bording:
Squashed and tests are green, though TC hasn't yet been updated to run the new TransportTests project.

adamralph:
Thanks @bording!

I hacked the build configs to make the new tests run:

* `1. Build`
  * General Settings
    * Artifact paths
      * Added `src/NServiceBus.Transport.RabbitMQ.TransportTests/bin/Release => TransportTests`
* `2. Run acceptance tests`
  * Build Steps
    * Added `NUnitV3 - Transport Tests`
  * Dependencies
    * Artifact Dependencies
      * `... 1. Build` 
        * Artifacts Paths
          * Added `TransportTests => .\TransportTests`
  * Parameters
    * Environment Variables (env.)
      * Added `env.Transport.UseSpecific = RabbitMQTransport`

I guess most of this can probably be pushed up into the build templates. /cc @Particular/deployment-maintainers @Particular/nservicebus-maintainers 

BTW - one thing I noticed (although I didn't want it to hold up the PR) is that the new project has a different naming convention to the others. I think we should probably rename it `NServiceBus.RabbitMQ.TransportTests` to be consistent. (Note that we'd have to update `1. Build` -> General Settings -> Artifact paths.)

bording:
@adamralph I intentionally chose the new naming convention, because that fits with the new [component naming guidelines](https://github.com/Particular/PlatformDevelopment/blob/master/designprinciples/nservicebus/component-naming.md).

To make things consistent, we should be looking to move everything else to match instead!

adamralph:
@bording :+1: makes sense. Let's change the other projects.

Note that in TC we'll have to change the project name since it's used in a few places, e.g. `src/%system.teamcity.projectName%.AcceptanceTests/bin/release => AcceptanceTests`. Perhaps some other settings will have to be tweaked too.

Issue #184, opened 2016-07-20
Messages may be silently lost when using DirectTopology and Publisher Confirms are disabled


adamralph:
Spun off from #135.

When using DirectTopology and Publisher Confirms are disabled, if a message is sent before the receiver has set up it's queue, the message will currently be lost silently. By subscribing to the BasicReturn event, we can log a message in this case.

We plan to implement this feature in 4.0. The 3.x codebase does not lend itself to implementing this feature without a significant amount of rework.

adamralph:
@Particular/rabbitmq-transport-maintainers please review the description for correctness.

udidahan:
I wouldn't call this a feature - if anything it's a fix for the bug of messages being lost silently.

adamralph:
Fixed in #195

mikeminutillo:
Should there be a "known issues" list for version 3.x of Rabbit Transport where we can list this?

adamralph:
@mikeminutillo good idea.

Do you know if there is already a place for this in docs? Or would it be a new page under http://docs.particular.net/nservicebus/rabbitmq/ ? /cc @Particular/docs-maintainers

Issue #185, opened 2016-07-20
Provide programmatic configuration of Publisher Confirms


adamralph:

andreasohlund:
Is the idea to add a extension on `SendOptions`?

```

var options = new SendOptions();

options.EnablePublisherConfirms();

await context.Send(new Message(),options);

```

or similar?

bording:
It can't be enabled/disabled a per-message basis. Currently the only way to configure it is via a property on the connection string. Being able to control it via a [setting extension](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs) is what we're looking to add.

adamralph:
Closed by #196

adamralph:
Re-opening since docs are not updated.

WojcikMike:
Doco PR opened, now we need to wait for this feature to get to nuget so that snippets would build

adamralph:
In that case, shall we close this now, since there's nothing more we can do until we've released RC?

WojcikMike:
Yeah, I think so

PR #195, opened 2016-08-30
Moving BasicReturn event subscription to happen no matter what the us…

WojcikMike:
fixes: #184

We can try to write automated test for this case but we need to disable publisher confirms if we implement #185 

cc: @Particular/rabbitmq-transport-maintainers please review and merge


WojcikMike:
ping: @bording or @adamralph to have a look

bording:
Pushed up a minor tweak to appease my code-formatting OCD. Will merge once the tests are green again.

PR #196, opened 2016-08-30
Create code api for publisher confirms

### LINKED ISSUE #185: Provide programmatic configuration of Publisher Confirms

WojcikMike:
fixes #185

@Particular/rabbitmq-transport-maintainers I made a stab at it. I assumed that:
* Code API override connection string setting
* If the override happen we log Info in the logger

### DIFF
--- a/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
@@ -46,6 +46,19 @@
                 throw new Exception("Invalid connection string. 'host' value must be supplied. e.g: \"host=myServer\"");
             }
 
+            if (settings.HasSetting(SettingsKeys.UsePublisherConfirms))
+            {
+                var publisherCofirms = settings.Get<bool>(SettingsKeys.UsePublisherConfirms);
+                if (this.ContainsKey("UsePublisherConfirms"))
+                {
+                    var message = string.Concat("The 'UsePublisherConfirms' connection string option was overriden by code api setting. Value set is ", publisherCofirms);
+
+                    Logger.Info(message);
+                }
+
+                connectionConfiguration.UsePublisherConfirms = publisherCofirms;
+            }
+
             if (ContainsKey("dequeuetimeout"))
             {
                 var message = "The 'DequeueTimeout' connection string option has been removed. Consult the documentation for further information.";
--- a/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
@@ -46,19 +46,6 @@
                 throw new Exception("Invalid connection string. 'host' value must be supplied. e.g: \"host=myServer\"");
             }
 
-            if (settings.HasSetting(SettingsKeys.UsePublisherConfirms))
-            {
-                var publisherCofirms = settings.Get<bool>(SettingsKeys.UsePublisherConfirms);
-                if (this.ContainsKey("UsePublisherConfirms"))
-                {
-                    var message = string.Concat("The 'UsePublisherConfirms' connection string option was overriden by code api setting. Value set is ", publisherCofirms);
-
-                    Logger.Info(message);
-                }
-
-                connectionConfiguration.UsePublisherConfirms = publisherCofirms;
-            }
-
             if (ContainsKey("dequeuetimeout"))
             {
                 var message = "The 'DequeueTimeout' connection string option has been removed. Consult the documentation for further information.";
@@ -80,6 +67,15 @@
             if (ContainsKey("prefetchcount"))
             {
                 var message = "The 'PrefetchCount' connection string option has been removed. Use 'EndpointConfiguration.LimitMessageProcessingConcurrencyTo' instead.";
+
+                Logger.Error(message);
+
+                throw new NotSupportedException(message);
+            }
+
+            if (ContainsKey("UsePublisherConfirms"))
+            {
+                var message = "The 'UsePublisherConfirms' connection string option has been removed. Use 'EndpointConfiguration.UsePublisherConfirms' instead.";
 
                 Logger.Error(message);
 
--- a/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
@@ -66,12 +66,12 @@
         }
 
         /// <summary>
-        /// Overrides the default usePublisherCofirms or value set in connection string.
+        /// Overrides the default usePublisherConfirms or value set in connection string.
         /// </summary>
         /// <param name="transportExtensions"></param>
         /// <param name="publisherConfirms">Controls if publisher confirms should be used.</param>
         /// <returns></returns>
-        public static TransportExtensions<RabbitMQTransport> SetPublisherConfirms(this TransportExtensions<RabbitMQTransport> transportExtensions, bool publisherConfirms)
+        public static TransportExtensions<RabbitMQTransport> UsePublisherConfirms(this TransportExtensions<RabbitMQTransport> transportExtensions, bool publisherConfirms)
         {
             transportExtensions.GetSettings().Set(SettingsKeys.UsePublisherConfirms, publisherConfirms);
             return transportExtensions;
--- a/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
@@ -64,5 +64,17 @@
             transportExtensions.GetSettings().Set(SettingsKeys.TimeToWaitBeforeTriggeringCircuitBreaker, waitTime);
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Overrides the default usePublisherCofirms or value set in connection string.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        /// <param name="publisherConfirms">Controls if publisher confirms should be used.</param>
+        /// <returns></returns>
+        public static TransportExtensions<RabbitMQTransport> SetPublisherConfirms(this TransportExtensions<RabbitMQTransport> transportExtensions, bool publisherConfirms)
+        {
+            transportExtensions.GetSettings().Set(SettingsKeys.UsePublisherConfirms, publisherConfirms);
+            return transportExtensions;
+        }
     }
 }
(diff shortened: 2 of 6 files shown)


WojcikMike:
ping: @bording or @adamralph to have a look

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -64,5 +64,17 @@ public static TransportExtensions<RabbitMQTransport> TimeToWaitBeforeTriggeringC
...
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Overrides the default usePublisherCofirms or value set in connection string.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        /// <param name="publisherConfirms">Controls if publisher confirms should be used.</param>
+        /// <returns></returns>
+        public static TransportExtensions<RabbitMQTransport> SetPublisherConfirms(this TransportExtensions<RabbitMQTransport> transportExtensions, bool publisherConfirms)
```
bording: I'm thinking the name of this method should be `UsePublisherConfirms` instead of `SetPublisherConfirms`.

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -64,5 +64,17 @@ public static TransportExtensions<RabbitMQTransport> TimeToWaitBeforeTriggeringC
             transportExtensions.GetSettings().Set(SettingsKeys.TimeToWaitBeforeTriggeringCircuitBreaker, waitTime);
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Overrides the default usePublisherCofirms or value set in connection string.
```
bording: Typo here: "cofirms"

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
```diff
@@ -46,6 +46,19 @@ public ConnectionConfiguration Parse(string connectionString)
                 throw new Exception("Invalid connection string. 'host' value must be supplied. e.g: \"host=myServer\"");
             }
 
+            if (settings.HasSetting(SettingsKeys.UsePublisherConfirms))
```
bording: While the `ConnectionStringParser` does have access to the settings (really only to pass it to `ConnectionConfiguration`, which in turn only needs it for the endpoint name) doing this check here feels wrong to me. Overall, it feels like RabbitMQTransportInfrastructure would be a better place for this...

bording:
Do we want to keep the connection string option around as an option? For the other options that were in the  connection string previously, I'm throwing an exception pointing them to the code-based option instead.

I think doing that again would be a decent idea. It prevents us from having to dig around in the settings collection in the ConnectionStringParser. It also means we don't have two ways of setting the same value, and then having to choose precedence if both are set, etc.

WojcikMike:
@bording your concerns have been addressed. Regarding if we should remove it from connection string. I definitelly see value in that. So my latest change include removing that. Let me know what you think. If we will follow that route I will create a doco PR to change it.

WojcikMike:
@adamralph as @bording is away for few days please have a look

adamralph:
@WojcikMike sure, will do.

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
```diff
@@ -73,6 +73,15 @@ public ConnectionConfiguration Parse(string connectionString)
                 throw new NotSupportedException(message);
             }
 
+            if (ContainsKey("UsePublisherConfirms"))
+            {
+                var message = "The 'UsePublisherConfirms' connection string option has been removed. Use 'EndpointConfiguration.UsePublisherConfirms' instead.";
+
+                Logger.Error(message);
+
+                throw new NotSupportedException(message);
```
adamralph: I'm not sure I agree with this pattern in general, i.e. log an error and then throw an exception. I think we should just throw an exception. However, it matches the other cases so it's fine for the purposes of this PR.

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -64,5 +64,17 @@ public static TransportExtensions<RabbitMQTransport> TimeToWaitBeforeTriggeringC
...
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Overrides the default usePublisherConfirms or value set in connection string.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        /// <param name="publisherConfirms">Controls if publisher confirms should be used.</param>
+        /// <returns></returns>
+        public static TransportExtensions<RabbitMQTransport> UsePublisherConfirms(this TransportExtensions<RabbitMQTransport> transportExtensions, bool publisherConfirms)
```
adamralph: Do we have a precedent for how Boolean transport extension settings are handled?

Specifically, I'm wondering if `UsePublisherConfirms()` (parameterless) would be preferable to `UsePublisherConfirms(true)`. Would there ever be a reason to call `UsePublisherConfirms(false)`?

Come to think of it, what is the default when the setting is not specified? True or false?

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -25,13 +25,24 @@ public RabbitMQTransportInfrastructure(SettingsHolder settings, string connectio
...
             RequireOutboxConsent = false;
         }
 
+        static void SetUsePublisherConfirmsIfPresent(SettingsHolder settings, ConnectionConfiguration connectionConfiguration)
+        {
+            if (settings.HasSetting(SettingsKeys.UsePublisherConfirms))
+            {
+                var publisherCofirms = settings.Get<bool>(SettingsKeys.UsePublisherConfirms);
+                connectionConfiguration.UsePublisherConfirms = publisherCofirms;
+            }
```
adamralph: It would be nice if we didn't have to do all this and could do something like:

```C#
connectionConfiguration.UsePublisherConfirms = settings.Get<bool>(SettingsKeys.UsePublisherConfirms, false)
```

I.e. if not present, use the second parameter (in this case `false`) as the default. But I guess that would require a change in core?

adamralph:
@WojcikMike I took a look at the changes and left a few comments.

This is the only one currently stopping me from merging: https://github.com/Particular/NServiceBus.RabbitMQ/pull/196#discussion_r77519089

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
WojcikMike: @adamralph I like this question a lot. 
I spent few minutes looking for precedents in other transports and below are my findings:
I found 2 examples of extension method without a parameter and 1 with :)
 
 - ASQ without parameter:
https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/release-7.0.0/src/Transport/Config/AzureStorageTransportExtensions.cs#L75-L79
 - ASB without parameter:
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Config/ExtensionMethods/AzureServiceBusTransportExtensions.cs#L100-L104
 - ASB with parameter:
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Config/ExtensionMethods/AzureServiceBusTransportExtensions.cs#L55-L59

My take is. At the moment default value is true. However as default value can change (even that in this case it is highly unlikely), putting UsePublisherConfirms(true) keep you on the safe side. 

What is more to make this method useful without parameter we would have to negate it to something like: DisablePublisherConfirms(). I would put my vote for leaving it with value.

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
WojcikMike: I think that you are right, however I would first check for all the keys and throw exception with all violations. As currently when one would have 2 deprecated options in connection string, he would run the app, learn about 1st parameter not being supported, fix it. Later run it again and now get an exception about the 2nd paramter. So maybe we sould check all first and later throw  message including all not supported parameters. 

But that could be a separate PR. What do you think?

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
WojcikMike: @timbussmann can you help on this. I noticed that we have:
SettingsHolder that contains Default values and overrides. Thanks to that you can then use
settings.GetOrDefault that returns default if the override is not present.

However in Rabbit we set the defaults not in the Settings but in here:
https://github.com/Particular/NServiceBus.RabbitMQ/blob/5f0164ca89c828c860b49ba1bc44c9b03dda93be/src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs#L36-L51

Should we set Defaults on settings? Or was it created for a different purpose? That would simplify the code above

WojcikMike:
Thanks @adamralph as always your comments were very insightful, I responded to every one of them

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
adamralph: Nice research. If I look at the examples you mention, the ones which do not have a parameter are of the form "use X for Y" and the one with a parameter is more of the form "use Z". I think our case here is the latter, so I agree that we should keep the parameter.

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
adamralph: >  I would first check for all the keys and throw exception with all violations

:+1: 

> that could be a separate PR

:+1:

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
timbussmann: not having the full context, but there is some important and very weird thing about the default values on the settings holder:

* If you do a Get, it already returns values configured as default values when no other value has been set.
* If you do a GetOrDefault, it will return `default(T)` when no custom or "default" value has been found.

So you could already write:

`settings.GetOrDefault<bool>(SettingsKeys.UsePublisherConfirms)` which will return `false` if no value has been configured.

But on the other side, yes, the SettingsHolder class is intended and available to be used for storing settings so you don't necessarily need to run your own.

WojcikMike: So It might be a good oportunity for another refactoring PR to move default values into SettingsHolder as defaults. But for now I would say that the discussed code make sense. What do you think @adamralph ?

adamralph: Yeah, for the purposes of this PR, I think the code is good to go.

I do think that there is scope for some good refactoring here, either in core, the transport, or both.

adamralph:
Thanks @WojcikMike !

PR #197, opened 2016-08-31
[WIP] Refactored logical address

SzymonPobiega:


bording:
Might just make sense to change this to `logicalAddress.ToString()` instead since we're not doing anything special here.

SzymonPobiega:
I was actually thinking about removing it altogether from `LogicalAddress` to enforce transport implementors to write the logic. The idea behind overriding `ToString` in `LogicalAddress` was for logging reasons but after all the changes in the routing `LogicalAddress` ended up being a very short-lived thing which is never logged anywhere.

SzymonPobiega:
Closing this one as the changes in the core are going to be postponed till next major.

PR #204, opened 2016-09-06
[WIP] Sending to non existing queue

WojcikMike:
Connects to #135 
A set of test to verify the behavior when sending to non existing queue using publisher confirms true/false and conventional/direct routing topology

For now I created tests to verify the behavior.

### DIFF
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
@@ -309,7 +309,11 @@
     <Compile Include="ConfigureEndpointRabbitMQTransport.cs" />
     <Compile Include="TaskEx.cs" />
     <Compile Include="When_receiving_a_reply_that_contains_a_legacy_callback_header.cs" />
+    <Compile Include="When_sending_to_non_existing_queue_using_conventional_without_publisher_confirms.cs" />
+    <Compile Include="When_sending_to_non_existing_queue_using_conventional_routing_and_publisher_confirms.cs" />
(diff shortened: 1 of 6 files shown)


WojcikMike:
Ok I understand what is happening now. On the 2 tests: with pending retries (no matter the routing topology), There are 2 events called: 
Basic_Return - this one would set exception
Basic_Ack - this one set the results as success
This is following the [doco](https://www.rabbitmq.com/confirms.html):
> For unroutable messages, the broker will issue a confirm once the exchange verifies a message won't route to any queue (returns an empty list of queues). If the message is also published as mandatory, the basic.return is sent to the client before basic.ack

Both events start new Task and try to set result on TaskCompletionSource. If return does that first than we have exception, if ack get it first then we get success as a result.

cc: @Particular/rabbitmq-transport-maintainers

WojcikMike:
@bording I know that once you tried to read my comment, So let me try to simplify it:

Both handlers for: Basic_Return and Basic_Ack spin a Task that tries to set the result on TaskCompletionSource. Basic_Return task try to set exception, while the other set's it as success.

Depending which task is executed first, this is the result of the TaskCompletionSource. So we have a race condition currently.

bording:
> Depending which task is executed first, this is the result of the TaskCompletionSource. So we have a race condition currently.

Except that according to RabbitMQ docs, the basic return message should always be issued first, so will this actually a race that would lose?

WojcikMike:
> Except that according to RabbitMQ docs, the basic return message should always be issued first,

That is true, But message handler for basic return creates a task, The task might or might not be exectuted at that time. Straight after that Basic ack is called which also creates a task. When running unit tests, it can happen that Task run from basic ack sets result first (as succesfull), which cause the exception to not being passed.

adamralph:
Closing, so we can delete the release-4.0.0 branch. The PR branch will still be available and can be rebased onto develop.

PR #208, opened 2016-09-08
Update GitVersionTask to 3.6.3

bording:
Also includes recent GitVersion,yml change that other repos are using to have GitVersion tag builds on release branches as RC.


adamralph:
Thanks @bording !

PR #216, opened 2016-09-13
moved assembly to net451 folder in output nupkg

adamralph:


adamralph:
Will GitVersion be OK if we merge this?

SimonCropp:
why r u targeting direct to master? shouldnt this be a release branch first?

adamralph:
@SimonCropp I don't think we have to release it, since the RabbitMQ.Client already restricts installation to  projects targeting .NET 4.5.1 or later. That means this change is really only a "correctness" change. We could delay until we release the next 3.x version, so long as we remember to do it. In that case we would indeed do it on a release branch.

Would going straight to master upset GitVersion?

SimonCropp:
> Would going straight to master upset GitVersion?

is should work if u move the tag. but u r breaking gitflow

adamralph:
OK, in that case I'm happy to hold off. I'll raise a 3.x milestone which we can alter to 3.4.1 or 3.5.0 if and when we need to make a release and I'll raise an issue on it so we remember to do this.

PR #218, opened 2016-09-13
Add RabbitMQTransportSettingsExtensions.PrefetchCountPerMessageProcessor

### LINKED ISSUE #198: Expose prefetch count API and optimise default
There still seem to be conversations happening about prefetch count being sub-optimal in 4.0 (NSB 6.0) vs. that used in 3.0 (NSB 5.x).

I couldn't find an existing issue which captures this concern so I raised this one. Perhaps things have moved along in my absence to make this issue redundant.

WojcikMike:
fixes #198 

cc: @Particular/rabbitmq-transport-maintainers


adamralph:
@bording let's discuss later.

bording:
Looking at this, I think I'd prefer to take a slightly different approach. Instead of exposing a setting that we take in as a multiplier, why don't we just use 3 * concurrency as the default value, but if you specify a non-zero value via settings we override our default valueand use that directly instead.

That way we can just call the setting 'PrefetchCount', and in the docs we describe that we default to 3 * concurrency.

We could also enforce that the specified count needs to be >= concurrency, since setting prefetch to lower than the concurrency value doesn't make much sense.

PR #223, opened 2016-09-14
refactor ValueToString method, remove LINQ

adamralph:


adamralph:
Paired with @WojcikMike

PR #226, opened 2016-09-14
Prefetch count settings changes

bording:
First, I modified things to a PrefetchCount setting that overrides the calculation.

As a separate commit, I then exposed the multiplier as a setting that can also be controlled, but if we don't want do do this now, I can always drop the commit.

What do you think? I think the name works, and with the intellisense description of what the PrefetchCount method does, I didn't think it needed to renamed when the PrefetchMultiplier method is also around.

Connects to #198

### DIFF
--- a/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
@@ -66,7 +66,7 @@
 
             if (ContainsKey("prefetchcount"))
             {
-                var message = "The 'PrefetchCount' connection string option has been removed. Use 'EndpointConfiguration.LimitMessageProcessingConcurrencyTo' and 'EndpointConfiguration.PrefetchCountPerMessageProcessor' instead.";
+                var message = "The 'PrefetchCount' connection string option has been removed. Use 'EndpointConfiguration.UseTransport<RabbitMQTransport>().PrefetchCount' instead.";
 
                 Logger.Error(message);
 
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -22,6 +22,7 @@
         readonly IChannelProvider channelProvider;
         readonly QueuePurger queuePurger;
         readonly TimeSpan timeToWaitBeforeTriggeringCircuitBreaker;
+        readonly int prefetchMultiplier;
         readonly ushort overriddenPrefetchCount;
 
         // Init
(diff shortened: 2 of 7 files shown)


src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
```diff
@@ -66,7 +66,7 @@ public ConnectionConfiguration Parse(string connectionString)
 
             if (ContainsKey("prefetchcount"))
             {
-                var message = "The 'PrefetchCount' connection string option has been removed. Use 'EndpointConfiguration.LimitMessageProcessingConcurrencyTo' and 'EndpointConfiguration.PrefetchCountPerMessageProcessor' instead.";
+                var message = "The 'PrefetchCount' connection string option has been removed. Use 'TransportExtensions<RabbitMQTransport>.PrefetchCount' instead.";
```
bording: I'm not sure what I think about having to refer to the class we use for the extension methods here, but it seemed less awkward that trying to say something about EndpointConfiguration.UseTransport<RabbitMQTransport>.PrefetchCount.

What do you think, should we go with that instead? More verbose, but possibly clearer.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -79,7 +81,25 @@ public void Start(PushRuntimeSettings limitations)
             connection = connectionFactory.CreateConnection($"{settings.InputQueue} MessagePump");
 
             var channel = connection.CreateModel();
-            channel.BasicQos(0, (ushort)Math.Min(ushort.MaxValue, Convert.ToUInt16(limitations.MaxConcurrency) * prefetchCountPerMessageProcessor), false);
+
+            long prefetchCount;
```
bording: Went with a long here to make sure we don't overflow when doing the multiplier calculation. Sure, not likely to be an actual problem, but seemed easy enough to make sure.

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
adamralph: :thu

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -79,7 +79,26 @@ public void Start(PushRuntimeSettings limitations)
...
+                    Logger.Warn($"The specified prefetch count '{prefetchCount}' is smaller than the specified maximum concurrency '{maxConcurrency}'. The maximum concurrency value will be used as the prefetch count instead.");
+                    prefetchCount = maxConcurrency;
+                }
+            }
+            else
+            {
+                prefetchCount = (long)maxConcurrency * prefetchMultiplier;
+            }
+
+            channel.BasicQos(0, (ushort)prefetchCount, false);
```
adamralph: Wont the cast to `ushort` throw an exception if the value is > `ushort.MaxValue`?

This is why we originally had `(ushort)Math.Min(ushort.MaxValue,...`.

bording: By default this cast is `unchecked` so it just uses the relevant bits in the cast. If the value is > `ushort.MaxValue`,  then it will just end up being `ushort.MaxValue`.

If we could have negative values in `prefetchCount` then that would produce some weird overflows, but all the input values are checked against 0 and negative.

src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
bording: I'll push up a change with this.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: And of course, you're right about needing something extra here, since the cast won't always return `ushort.MaxValue`.

Will fix!

PR #227, opened 2016-09-14
Fix UsePublisherConfirms connection string error message

bording:
While working on #226 I realized that the publisherconfirms message was referring to the wrong thing also.

I went ahead and used the more verbose message that I mentioned in https://github.com/Particular/NServiceBus.RabbitMQ/pull/226#r78827464.

Connects to #185


WojcikMike:
I am almost sure that we have this problem in at least 1 more message. I wrote this comment knowing that this is not entirely accurate because it was used in at least 1 more place in this repo.

WojcikMike:
The thing that should change with that change is 
ObsoleteAppSettings comments related to circuit breaker time

PR #240, opened 2016-09-30
add csx build script

### LINKED ISSUE #237: CSX build script
As a result of spike #220

adamralph:
Fixes #237

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,0 +1,28 @@
+:: the windows shell, so amazing
+
+:: options
+@echo Off
+cd %~dp0
+setlocal
+
+:: determine cache dir
+set NUGET_CACHE_DIR=%LocalAppData%\.nuget\v3.4.4
+
+:: download nuget to cache dir
+set NUGET_URL=https://dist.nuget.org/win-x86-commandline/v3.4.4/NuGet.exe
+if not exist %NUGET_CACHE_DIR%\NuGet.exe (
+  if not exist %NUGET_CACHE_DIR% md %NUGET_CACHE_DIR%
+  echo Downloading '%NUGET_URL%'' to '%NUGET_CACHE_DIR%\NuGet.exe'...
+  @powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest '%NUGET_URL%' -OutFile '%NUGET_CACHE_DIR%\NuGet.exe'"
+)
+
+:: copy nuget locally
+if not exist src\.nuget\NuGet.exe (
+  copy %NUGET_CACHE_DIR%\NuGet.exe src\.nuget\NuGet.exe > nul
+)
+
+:: restore packages
+src\.nuget\NuGet.exe restore .\src\NServiceBus.RabbitMQ.sln -MSBuildVersion 14
+
+:: run script
+"%ProgramFiles(x86)%\MSBuild\14.0\Bin\csi.exe" build.csx %*
--- a/build.csx
+++ b/build.csx
@@ -1,0 +1,63 @@
+#r "System.IO.Compression.FileSystem.dll"
+#load "scripts/cmd.csx"
+#load "scripts/download.csx"
+#load "scripts/inspect.csx"
+#load "src/packages/simple-targets-csx.5.0.0/simple-targets.csx"
+
+using System;
+using System.IO;
+using System.IO.Compression;
+using static SimpleTargets;
+
+// locations
+var resharperCltUrl = "https://download.jetbrains.com/resharper/JetBrains.ReSharper.CommandLineTools.2016.2.20160913.100041.zip";
+var resharperCltPath = $"{Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)}/.resharper/JetBrains.ReSharper.CommandLineTools.2016.2.20160913.100041.zip";
+var inspectCodePath = "./.resharper/inspectcode.exe";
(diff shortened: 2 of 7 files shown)


SimonCropp:
I thought the agreement was to not do this?

bording:
@SimonCropp The current plan is to keep it for local convenience, but not use it for CI.

SimonCropp:
Source code repository are not a way for u to share stuff that is not required for the build. It pollutes the history. Based on this approach any developer should be able to add in any custom script they need. Would you accept a pull request from the community with a custom build script that is their unique preference. 

Share this script in another manner

andreasohlund:
>Share this script in another manner

Such a tool would be nice for our other components as well

We could use a default convention to "run all the stuff that ends with `tests`"

Eg:

`>choco install csxbuilder`

`>csxbuilderconfig --global testrunner "NUnit:latest"`

`>csxbuilderconfig --global testfilter "**Tests.dll"`


`>csxbuilder build`

Thoughts?

build.csx
```diff
@@ -0,0 +1,102 @@
...
+        DependOn = new[] { "download-resharper-clt" },
+        Do = () => { if (!File.Exists(inspectCodePath)) ZipFile.ExtractToDirectory(resharperCltPath, Path.GetDirectoryName(inspectCodePath)); },
+    });
+
+targets.Add(
+    "inspect",
+    new Target
+    {
+        DependOn = new[] { "build", "unzip-resharper-clt", },
+        Do = () => Cmd(inspectCodePath, $"--profile={dotSettings} {solution}"),
```
adamralph: It turns out that this doesn't have the desired effect. The exit code is always zero regardless of which types of violations are found is to output the reports to disk and then inspect those which are marked with "error" to see if there is anything which will fail on the CI server. I'll continue to investigate, but it seems like we may have to write some code to do that, since it is taken care of by something built into TeamCity when run on the CI server.

adamralph:
@Particular/rabbitmq-transport-maintainers this should be fully functional now. I added a target to inspect the output of `inspectcode.exe` and fail the build if any errors were found. The output looks something like this:

image

*(truncated)*

image

*(truncated)*

image

build.cmd
```diff
@@ -0,0 +1,29 @@
...
+:: download nuget to cache dir
+set NUGET_URL=https://dist.nuget.org/win-x86-commandline/v3.4.4/NuGet.exe
+if not exist %NUGET_CACHE_DIR%\NuGet.exe (
+  if not exist %NUGET_CACHE_DIR% md %NUGET_CACHE_DIR%
+  echo Downloading '%NUGET_URL%'' to '%NUGET_CACHE_DIR%\NuGet.exe'...
+  @powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest '%NUGET_URL%' -OutFile '%NUGET_CACHE_DIR%\NuGet.exe'"
+)
+
+:: copy nuget locally
+if not exist .nuget\NuGet.exe (
```
bording: Why create a root .nuget folder when there is already one in src?

bording: Also, not sure I see the benefit of downloading nuget.exe to an AppData folder and having to then copy it to the project folder. Why not just check and download it directly where it will be used?

build.csx
```diff
@@ -0,0 +1,66 @@
...
+#load "src/packages/simple-targets-csx.5.0.0/simple-targets.csx"
+
+using System;
+using System.IO;
+using System.IO.Compression;
+using static SimpleTargets;
+
+// locations
+var resharperCltUrl = "https://download.jetbrains.com/resharper/JetBrains.ReSharper.CommandLineTools.2016.2.20160913.100041.zip";
+var resharperCltPath = $"{Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)}/.resharper/JetBrains.ReSharper.CommandLineTools.2016.2.20160913.100041.zip";
```
bording: Don't really like this being saved forever in AppData vs maybe a temp folder?

build.csx
```diff
@@ -0,0 +1,66 @@
...
+    () => { if (!File.Exists(resharperCltPath)) Download(resharperCltUrl, resharperCltPath); });
+
+targets.Add(
+    "unzip-resharper-clt",
+    DependsOn("download-resharper-clt"),
+    () => { if (!File.Exists(inspectCodePath)) ZipFile.ExtractToDirectory(resharperCltPath, Path.GetDirectoryName(inspectCodePath)); });
+
+targets.Add(
+    "run-inspectcode",
+    DependsOn("build", "unzip-resharper-clt"),
```
bording: Could this dependency chain be based on the existence of inspectcode.exe in such a way that if it's missing it would download on demand and unzip, vs having to keep the zip around all the time now?

build.csx
```diff
@@ -0,0 +1,66 @@
...
+        if (!Inspect($"{solution}.inspections.xml"))
+        {
+            Console.WriteLine("Inspection failed!");
+            Environment.Exit(1);
+        }
+    });
+
+targets.Add("reset-broker", () => Cmd("powershell.exe", resetBroker));
+
+targets.Add("unit-test", DependsOn("build", "reset-broker"), () => Cmd(nunit, unitTests));
```
bording: Having reset-broker as a dependency breaks things for me, because I always run RabbitMQ in a VM, so the Reset-RabbitMQ.ps1 script doesn't work locally.

adamralph:
@bording the latest two commits should address the issues we discussed.

I'm happy to squash all 3 commits before merging.

bording:
👍 LGTM

adamralph:
Squashed.

PR #242, opened 2016-10-05
Prepare release branch for merging to master

bording:


adamralph:
Do we have to update the repo standards before the merge to master? If not, I'd prefer to do it in develop instead and restrict this PR to only what is *absolutely necessary* for RTM (the update to NSB 6.0 RTM).

Having said that, it's probably a purist argument only since the repo standards commit does not touch the code, but only the .DotSettings file.

adamralph:
Never mind, I just saw the discussion about the acceptance test source code changes which require the new DotSettings file.

Issue #234, opened 2016-09-20
Native timeouts


adamralph:
Spun off from https://github.com/Particular/NServiceBus/issues/4085

Currently, I've no idea if this is feasible with RabbitMQ.

WojcikMike:
An idea presented by @bording would be to implement using:
https://github.com/rabbitmq/rabbitmq-delayed-message-exchange
https://www.rabbitmq.com/blog/2015/04/16/scheduling-messages-with-rabbitmq/

But at this time it is experimental plugin. So there is a risk associated with that action. However we could spike it, and see how it works

adamralph:
If the plug-in ever comes out of experimental status, but remains a plug-in, we'd also have to decide on our policy regarding the requirement of plug-ins for our features.

bording:
I've come up with an idea that doesn't require that plugin. I hope to have a spike ready soon to try out.

adamralph:
Meeting scheduled for Thu 20th Oct 13:30 UTC.

MarcinHoppe:
@Particular/rabbitmq-transport-maintainers Have you reached a conclusion on this topic?

adamralph:
@MarcinHoppe we had our meeting yesterday. I will write up the results on this issue today, as soon as I have time.

adamralph:
In summary, we are considering two approaches:

### Method 1 "Queue per delay value"

This method creates a queue with a time-to-live (TTL) equal to the delay value for a given message. The message is sent to that queue and when it expires it is sent to the destination queue.

### Method 2 "Queue per delay bit"

This method creates 32 queues each with a TTL representing a bit in the largest possible delay value (32 bits gives a maxiumum range of ~136 years). A given message is sent to the queue representing the most significant bit set to 1 in its delay value and when it expires it is sent to the queue representing the next less significant bit set to 1. When it expires from the last queue it is sent to the destination queue.
--
We will consider which, if any approach to take, and then raise another issue to get it done.

I'm also closing the spikes now as they have served their purpose.

PR #245, opened 2016-10-08
[Spike] - Native timeout - queue per delay

bording:
Here is my spike for native delayed delivery/timeouts, implementing the idea I mentioned in #234.

Take a look, and let me know what you think!

//cc @SeanFeldman


src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -76,6 +77,26 @@ public void Initialize(IModel channel, string mainQueue)
             channel.QueueBind(mainQueue, mainQueue, string.Empty);
         }
 
+        public string SetupDelay(IModel channel, long delay)
```
SeanFeldman: > `SetupDelay`

`SetupDelayTopology`?

src/NServiceBus.RabbitMQ/Routing/DirectRoutingTopology.cs
```diff
@@ -45,6 +46,22 @@ public void Initialize(IModel channel, string main)
             //nothing needs to be done for direct routing
         }
 
+        public string SetupDelay(IModel channel, long delay)
```
SeanFeldman: Same as above

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -20,9 +21,20 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
...
 
+            DelayDeliveryWith delayDeliveryWith;
+            DoNotDeliverBefore doNotDeliverBefore;
             DiscardIfNotReceivedBefore timeToBeReceived;
+            long delay = 0;
 
-            if (TryGet(deliveryConstraints, out timeToBeReceived) && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
+            if (TryGet(deliveryConstraints, out delayDeliveryWith))
+            {
+                delay = Convert.ToInt64(Math.Round(delayDeliveryWith.Delay.TotalSeconds) * 1000);
```
SeanFeldman: This is where we'll allow user precision to be used to have less of precision?

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -20,9 +21,20 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
...
+            }
+            else if (TryGet(deliveryConstraints, out doNotDeliverBefore))
+            {
+                delay = Convert.ToInt64(Math.Round((doNotDeliverBefore.At - DateTime.UtcNow).TotalSeconds) * 1000);
+            }
+            else if (TryGet(deliveryConstraints, out timeToBeReceived) && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
```
SeanFeldman: What about the case of TTBR and delayed delivery that is larger than TTBR?

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
bording: Yeah, if we decide to expose that as a feature, this is where it would be.

src/NServiceBus.RabbitMQ/Receiving/DelayedReceiver.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using global::RabbitMQ.Client;
+    using global::RabbitMQ.Client.Events;
+
+    class DelayedReceiver
```
bording: I think `DelayedMessageReceiver` sounds decent. At this point I'm not sure this will even survive as a separate class vs. possibly trying to refactor the message pump code since these are are so similar.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
SeanFeldman: Perhaps leave a note to let others know we've discussed this edge case?

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
bording: At the moment I'm just considering these three options mutually exclusive, with delay taking precedence over TTBR.

The other interesting this is that both DelayDeliveryWith and DoNotDeliverBefore can be set on a message, and you just have to pick one or the other to take precedence over the other. How did you handle it in ASB?

bording:
As @SeanFeldman mentioned above, one of the ideas here is that while the `x-message-ttl` property accepts milliseconds, I'm currently rounding the incoming delay values to the nearest second, because it didn't seem like having millisecond precision was worth the potential of all the additional queues it would create.

This could be exposed as a setting, letting the user choose the level of precision, choosing between precision vs. number of "buckets" for the timeouts. For example, we could support rounding to the nearest minute, or hour instead.

Alternatively, we could also do some sort of scaling precision, where smaller values do get per-millisecond or per-second precision, but when you start looking at days, we could scale to the nearest hour instead.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
SeanFeldman: In ASB we look at `DeliveryConstaint` on operation is `DelayedDeliveryConstraint` and use `Time.UtcNow() + delay` for scheduled time. Otherwise, it's a `DoNotDeliverBefore` and we use the delay value as scheduled time. We also set `TimeToLive` on the message regardless of the delayed delivery.  ASB will expire the message and won't deliver it if TTBR has elapsed.

SeanFeldman:
To add to what @bording said, reason we've decided to round to the nearest seconds is because delayed delivery with extremely fine grained precision over a messaging system makes no sense. By dropping milliseconds, we'll generate significantly less number of exchanges/queues.

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -76,6 +77,26 @@ public void Initialize(IModel channel, string mainQueue)
...
+            channel.QueueDeclare("delay-triggered", useDurableExchanges, false, false, null);
+            channel.QueueBind("delay-triggered", "delay-triggered", "");
+
+            var address = $"delay-{delay}";
+
+            channel.ExchangeDeclare(address, "fanout", useDurableExchanges);
+
+            var arguments = new Dictionary<string, object>();
+            arguments.Add("x-message-ttl", delay);
+            arguments.Add("x-dead-letter-exchange", "delay-triggered");
```
WojcikMike: So basically we create a queue, we set TTL and allow the messages to be discarded, as dead letter queue setting the delay-triggered queue from which we pick messages and send to its original destination. 

That is crafty @bording  :clap:

src/NServiceBus.RabbitMQ/Receiving/DelayedReceiver.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (connection.IsOpen)
+            {
+                connection.Close();
+            }
+        }
+
+        async void Consumer_Received(object sender, BasicDeliverEventArgs eventArgs)
+        {
+            eventArgs.BasicProperties.Headers.Remove("x-death");
```
tmasternak: Do we want to check if the reason for `TTL expired` and not `maxlen` reached (asking based on Dead-Lettered Messages in https://www.rabbitmq.com/dlx.html)?

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -76,6 +77,26 @@ public void Initialize(IModel channel, string mainQueue)
...
+            channel.QueueDeclare(address, useDurableExchanges, false, false, arguments);
```
tmasternak: Are we setting up a queue per each delay value passed? Is it possible to have a single queue and use per-message TTL?

src/NServiceBus.RabbitMQ/Receiving/DelayedReceiver.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            var address = Encoding.UTF8.GetString((byte[])eventArgs.BasicProperties.Headers[delayedDestinationHeader]);
+
+            var channel = channelProvider.GetPublishChannel();
+
+            try
+            {
+                await channel.RawSendInCaseOfFailure(address, eventArgs.Body, eventArgs.BasicProperties).ConfigureAwait(false);
```
tmasternak: Would we want to pass the message through recoverability? I'm not sure it makes sense in case of Rabbit.

src/NServiceBus.RabbitMQ/Receiving/DelayedReceiver.cs
bording: We don't ever set a queue length limit, so for this to be a reason, it would mean the end users have applied this setting on their own. 

I suppose we could be super-defensive here, but I'm not sure there is enough benefit.

SeanFeldman: Out of curiosity, what's a `maxlen` for Rabbit? Would it be possible to have so many timeouts to exceed the max queue size?

SeanFeldman: I.e. if we get `maxlen` that would only be because user has specified that on the queue (read as "changed the default")? In that case, throwing an exception and moving on? 😄

bording:
I've pushed up a change that removes the need for `DelayedReceiver` altogether. Instead, the `delay-triggered` exchange becomes a header exchange, and each endpoint queue gets bound to it based on the value of the DelayedDestination header.

This means that delayed messages make it all the way back to the final destination queue completely through the broker, and no queue consumer is needed! This should eliminate the narrow possibility of the DelayedReceiver generating a duplicate message by failing to ack after sending.

adamralph:
See https://github.com/Particular/NServiceBus.RabbitMQ/issues/234#issuecomment-255379904

SeanFeldman:
@adamralph perhaps worth mentioning the decision to go with spike #2 and an option of ops tooling if needed.

PR #246, opened 2016-10-12
[Spike] - Native timeouts - queue per delay bit

tmasternak:
Here is a spike for idea that we had together with @SzymonPobiega showing alternative approach to implementing native delayed delivery in RabbitMq.

### Idea
The idea is based on the fact that we can chain header exchanges. First of all we create a chain structure (with minor tweaks it could be shared by all endpoints running on a given broker) that looks like this:

image

It consists of 10 stages and at each stage we have two queues `wait` and `no-wait` that have TTL set to 2^n or 0 seconds. The message which is sent to the upper most level travels through the chain and at each stage goes to a waiting or no-waiting queue which depends on the header set (`delay-level-x`).

Before the message is sent we encode delay in binary format e.g. 11 -> 0000001011 and encode that into headers. So for a given example headers `delay-level-0`, `delay-level-1` and `delay-level-3` would be set to `1` and others to `0. 

### Results
The PR includes a sample console app that does delayed deliveries. It was not extensively tested but looks promising. Here is a sample run. It shows the requested delay and the actual one achieved during execution.

image

Connects to #234


bording:
So, this is a really interesting idea, though I'm mixed on what I think about it.

It does ensure there are a finite number of exchanges/queues created to manage timeouts, so that's cool. I also like how this lets the broker handle moving the messages from queue to queue, so there shouldn't be a chance of introducing duplicates like there would be if you had to consume the message and send it to a new queue.

On the other hand, I wonder if a finite number of queues is really something that needs to be optimized for. In a real system, how many unique timeout values would you run into? How many queues would the queue-per-timeout approach actually end up with?

From an Ops perspective, this is also more conceptually difficult to understand and reason about if you wanted to see what timeouts you have pending and when they're going to expire vs. being able to look at the contents of a single queue.

The simpler approach also has the benefit of effectively having no upper bound on the maximum delay length, while you need to add a lot of extra delay levels just to potentially support higher ranges. You'd need 25 levels just to be able to support a delay of a year.

SzymonPobiega:
@bording well, the benefit here is that the number of queues is static and it does not change the topology while running. This is I think quite interesting from ops point of view. 

When it comes to estimating how many timeouts do I have and when they are going to expire, ops can use the `2*n^queue_level` formula to calculate the upper bound value of when the timeouts in that queue are going to expire. 

With 32 levels we can get enough range to not worry about it. This means 64 queues and 32 exchanges.

tmasternak:
As a side note. It looks that it's possible bind [exchange-to-exchange](https://www.rabbitmq.com/e2e.html) which would remove a need for `no-wait` queues.

bording:
> As a side note. It looks that it's possible bind exchange-to-exchange which would remove a need for no-wait queues.

Having a level be a single exchange and queue is much better, so I think taking advantage of that would be required!

tmasternak:
The change with single queue, exchange is already in.

SzymonPobiega:
I did some rough performance checks and it seems like rabbit is able to pump 10K delayed messages through 10 exchanges in around 5-7 seconds on my machine. Hard to get the exact number because I don't know how to hook up to the last message queued to `triggered` queue. 

I think that for now such a rough value is enough (until we figure out a better way to measure it).

bording:
I've rebased this against the latest changes in my branch, so now the `delay-triggered` exchange directly routes things back to the original queue.

This also ends up fixing a problem where `delay-triggered` wasn't being declared before attempting to bind to it, which was causing test failures.

adamralph:
See https://github.com/Particular/NServiceBus.RabbitMQ/issues/234#issuecomment-255379904

PR #247, opened 2016-10-14
[SPIKE] Native automatic command routing

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/980

Adds another toplogy that enables configuration-free command routing
- Commands are routed to the endpoints that have handlers for them
- Command inheritence is not supported :-(
- When an endpoint is re-deployed without a handler for a given command, the routing is automatically adjusted to not route that command any more
- When there is an endpoint that "claimed" a given command, no other endpoint can define a route for it, preventing commands to be routed to multiple endpoints
- When there is no endpoint handling a given command, the sender gets an exception when sending (preventing message loss)
### Topology

native command routing

There is a global `Commands` exchange that is used to publish all commands. Each handling endpoints creates its own `Commands-{endpointName}` exchange which is used to configure its routing. When the endpoints starts up, it creates a "detour" exchange which allows it to drop and re-create the  `Commands-{endpointName}` without missing a message. Re-creating this exchange drops all the bindings thus if a given message is no longer handled, the binding is removed. The `{commandType}` exchange is used to ensure that only one endpoint can handle a given type of command. When creating it, the endpoint sets the `alternate-exchange` property to its `/dev/null` fanout exchange. When another endpoint wants to create an exchange for the same command type using its `/dev/null` fanout exchange, it gets an exception.
### TODO
- [x] Change the subscription logic to also go through additional endpoint to allow graceful unsubscribing when an endpoint no longer has a handler for a given event
- [ ] Automatically re-route commands to proper destinations when a handler has been removed

### DIFF
--- a/src/CommandWireSample/App.config
+++ b/src/CommandWireSample/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" />
+    </startup>
+</configuration>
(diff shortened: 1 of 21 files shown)


janovesk:
How about replies that need to go directly to a specific endpoint instance, @SzymonPobiega ?

SzymonPobiega:
@janovesk I don't touch the replies here. They go directly to the input queue via the default exchange I believe.

SzymonPobiega:
Got down to 7 ATs failing. Raised an issue about core being not ready for multicast native command routing: https://github.com/Particular/PlatformDevelopment/issues/980

SzymonPobiega:
@Particular/rabbitmq-transport-maintainers I think we can safely close this. The spike proved it is possible so if we at some point decide this would be useful then we dig this out.

PR #249, opened 2016-10-18
Use single admin connection to create queues

bording:
This changes the `QueueCreator` class to open a single administration connection to do its work, instead of opening a separate connection & channel for each queue.


WojcikMike:
Good change, LGTM

Issue #248, opened 2016-10-15
Declare queues inside routing topologies for more flexibility in custom routing topologies


volak:
Here at this line https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs#L38

Queues are being declared before routing topology is initialized.  I just started messing with rabbitmq-sharding https://github.com/rabbitmq/rabbitmq-sharding - one of the requirements to use is to not declare the main queue, it handles sharding instances of the queues across nodes.  Meaning that with the NSB transport like it is its not compatible with this plugin

Not sure if its a big deal to move queue declare into possible user-land or not - but thats up to you guys

---

### Plan of Attack
- [x] make changes - #263
- [x] add docs - https://github.com/Particular/docs.particular.net/pull/2247

adamralph:
@volak thanks for raising this. I'll have to investigate a little before giving an opinion.

volak:
All in all the IRoutingTopology extension is nice, but its very limited.  For instance we can't use any constructor arguments - which is a pretty big limitation if you need to do something fancy with other services for routing

bording:
@volak Moving the `QueueDeclare` into the `IRoutingToplogy` implementations is something I've been contemplating recently. However, since that would be a behavior break for any existing implementations, we'd need to release a new major version to be able to make that change, so it's not something we can do immediately.

bording:
>All in all the IRoutingTopology extension is nice, but its very limited. For instance we can't use any constructor arguments - which is a pretty big limitation if you need to do something fancy with other services for routing

@volak Take a look at #250 and let me know if that fixes up things regarding `IRoutingTopology` limitations.

volak:
@bording Looks good!

adamralph:
@volak to what degree does NServiceBus.RabbitMQ 4.1.0 (containing #250) address your concerns?

adamralph:
@volak after some discussion, we decided that the original idea you raised has merit regardless of the change we've already made to allow non-default constructors (#250).

We're now treating this as a feature which we would like to implement. Currently we believe it will have to be a breaking change so will require a major version bump, but we're investigating ways in which we can avoid the breakage and deprecate instead, which would allow a minor version bump. If the latter is possible, then the feature will probably be delivered sooner.

volak:
[...]

PR #250, opened 2016-10-18
Create IRoutingTopology with factory delegate

bording:
This adds a `UseRoutingTopology` overload that accepts a factory delegate in order to allow creating instances with a non-default constructor.

The original `UseRoutingTopology` method and `UseDirectRoutingTopology` have been reimplemented using the new factory approach. The original `UseRoutingTopology` method should be obsoleted on the next major release.

Since both `ConventionalRoutingTopology` and `DirectRoutingTopology` receive `useDurableExchanges` in their constructors, the factory delegate also passes this in. External implementations should now be at feature parity with our internal topologies.

The routing topology creation  in `RabbitMQTransportInfrastructure` has been updated and simplified as part of this as well.

This addresses the concern raised in https://github.com/Particular/NServiceBus.RabbitMQ/issues/248#issuecomment-254238054 by @volak.


bording:
@adamralph I made a change to the factory parameter description which attempts to explain what the parameter is for.

Also, if we decide we can go ahead and do it now, I'll add the obsolete to the old method.

adamralph:
@bording thanks, the docs look good.

It seems that we can indeed obsolete the old method now with a warning, so if you add that then I'll be 👍 to merge this.

bording:
I pushed up a couple more changes. First, I realized that the new method doesn't need to be generic, so I fixed that.

The obsolete attribute I added generates the following message:

'RabbitMQTransportSettingsExtensions.UseRoutingTopology<T>(TransportExtensions<RabbitMQTransport>) is obsolete: 'Use `RabbitMQTransportSettingsExtensions.UseRoutingTopology(TransportExtensions<RabbitMQTransport> transportExtensions, Func<bool, IRoutingTopology>)` instead. Will be treated as an error from version 5.0.0. Will be removed in version 5.0.0.'

adamralph:
@bording 👍 is it worth squashing the commits before merging?

bording:
@adamralph I have squashed everything into a single commit.

FYI, one minor tweak I did include as part of this squashed commit is I changed the `RemoveInVersion` value of the obsolete attribute to 6.0 instead of 5.0. After having some other conversations around obsoletes with the core maintainers, it seems more correct to me to go ahead and have this around in v5 to let us provide an error message, before removing it entirely in v6.

adamralph:
LGTM. Will merge when green.

PR #254, opened 2016-10-31
update to RabbitMQ.Client 4.1.1

adamralph:
I was smoke testing the new RabbitMQ.Client so I thought I may as well PR the update so we build against it.


WojcikMike:
LGTM

PR #256, opened 2016-11-11
[SPIKE] move queue declare to topology (breaking)

adamralph:
As a reference point, this is how moving queue declare to the routing topologies could look if we didn't mind the breaking change to `IRoutingTopology`.

Connects to #248.

PR #257, opened 2016-11-11
[SPIKE] move queue declare to topology (non-breaking)

adamralph:
This kind of cool, but also kind of crazy.

Ultimately, the constraint is that it's impossible to version interfaces in a non-breaking way so we need a new interface.

If an old routing topology factory is provided in config, it's immediately wrapped in an adapter to the new interface and the rest of the code base uses the new interface.

`IRoutingTopology2` is a tear jerking name for the new interface, but what's better? `IRoutingTopologyNew`, `IRoutingTopologyAdvanced`, `IRoutingTopologyExt`? All equally repulsive to me.

Connects to #248.


bording:
Instead of having to create an entirely new, all-encompassing routing interface, I was thinking about just creating an additional interface that has the single new method we need. `QueueCreator` could then check and see if the `IRoutingTopology` instance passed in implements the new interface also and can call the new method. Otherwise it falls back to the original logic.

Then, in the next major, we can just update `IRoutingTopology`, and deprecate the newly introduced interface since we won't need it any more.

adamralph:
> an additional interface that has the single new method

I considered this too, but it doesn't allow a deprecation path. I.e. it would have to be a clean break on a major, with no prior deprecation. If we're happy to do that, then I would probably prefer to just make that breaking change straight away and go with the approach proposed in #256.

bording:
>I considered this too, but it doesn't allow a deprecation path. I.e. it would have to be a clean break on a major, with no prior deprecation.

While we did have an obsolete to point the way for the namespace change, I'd say that overall we've already made these kind of changes to the interface going from v3 to v4: 

https://docs.particular.net/nservicebus/upgrades/rabbitmq-3to4#routing-topology-custom-routing-topology

>  then I would probably prefer to just make that breaking change straight away and go with the approach proposed in #256.

I'm not yet convinced this the conclusion to come to regarding this change.

adamralph:
Closing spike.

PR #260, opened 2016-11-12
[WIP] Moving default values into SettingHolder

WojcikMike:
This is an idea how we could set up default values.

cc: @Particular/rabbitmq-transport-maintainers please have a look if it is worth pursuing.

From my end. 

Advantages:
 - all default values are in one place `SettingHolder`
 - That results in one place where we can apply defaults `DefaultConfigurationValues`
 - Removing ConnectionConfiguration caused that we pass only SettingHolder
 - `RabbitMQTransportInfrastructure` is a bit simpler as defaults allowed us to remove some ifs and default values hard coded in this class (PrefetchCount, PrefetchMultiplier etc.)

Disadvantages: 
 - `SettingHolder` use casting to store values (has a Dictionary<string,object> 
 - ConnectionConfiguration gave us strongly typed access to connection parameters, now we need to use `settings.Get(key)` as the values are in dictionary.


src/NServiceBus.RabbitMQ/Configuration/ConnectionStringParser.cs
```diff
@@ -4,42 +4,47 @@
...
+            DefaultConfigurationValues.Apply(settings);
 
             foreach (var key in Keys.Cast<string>())
             {
-                var property = connectionConfigurationType.GetProperty(key, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
-                property?.SetValue(connectionConfiguration, TypeDescriptor.GetConverter(property.PropertyType).ConvertFrom(this[key]));
+                var normalizedKey = string.Format("RabbitMQ.{0}", key);
+                if (settings.HasSetting(normalizedKey))
+                {
+                    settings.Set(normalizedKey, TypeDescriptor.GetConverter(settings.Get(normalizedKey).GetType()).ConvertFrom(this[key]));
```
WojcikMike: THis is a line that I don't particularly like. At the moment one of our setting (CertPassphrase) has default value null => this line was throwing null ref expcetion. For the time being I changed default value to `string.Empty` only to not spend to much time how to solve this piece of a puzzle.

bording:
At this point, I'm 👎 for these changes. I don't think they provide enough value to be worth it, especially considering some of the downsides that are introduced.

WojcikMike:
@adamralph can you express your opinion with regards to if this is worth pursuing? I am unsure.

adamralph:
@WojcikMike right now I would say probably not. I think there are some simplifications/improvements we can make wrt `ConnectionConfiguration` and `ConnectionStringParser` but I'm not sure what the best way is yet. Perhaps it might be a series of less invasive PR's. I'm planning to experiment a little on the back of the change introduced in #277, and I think that will give me a better idea.

I suggest we close this PR but keep the branch for now, as a reference point, and then explore in which direction some smaller changes might take us. Perhaps they will lead back to something similar.

WojcikMike:
PR closed

adamralph:
@WojcikMike is is worth keeping the branch or shall we delete?

WojcikMike:
branch deleted. Thanks for spotting that @adamralph

PR #262, opened 2016-11-14
Fix custom message id strategy endless retry loop

### LINKED ISSUE #261: Custom message Id strategy that returns invalid value will cause endless retry loop
If a custom message id strategy ever returns an invalid message id (null or an empty string), then that causes both the `MessageContext` and `ErrorContext` constructors to throw `ArgumentNullException`s, causing the message pump to reject and retry the message in an endless loop.

While the custom message id strategies are intended to assist in native interop, it gets applied to all incoming messages. This means that any incoming messages sent from another NSB endpoint, or even internal control messages, need to be able to produce a valid message id through the strategy, which may not always be possible. For example, if a native message fails to process and ends up in a delayed retry, the timeout manager control messages could end up in an endless retry loop.

bording:
Fixes #261

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -19,7 +19,17 @@
             this.messageIdStrategy = messageIdStrategy;
         }
 
-        public string RetrieveMessageId(BasicDeliverEventArgs message) => messageIdStrategy(message);
+        public string RetrieveMessageId(BasicDeliverEventArgs message, Dictionary<string, string> headers)
+        {
+            var messageId = messageIdStrategy(message);
+
+            if (string.IsNullOrWhiteSpace(messageId) && !headers.TryGetValue(Headers.MessageId, out messageId))
+            {
+                throw new InvalidOperationException("The message ID strategy did not provide a message ID, and the message does not have an 'NServiceBus.MessageId' header.");
+            }
+
+            return messageId;
+        }
 
         public Dictionary<string, string> RetrieveHeaders(BasicDeliverEventArgs message)
         {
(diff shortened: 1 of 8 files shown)


bording:
Pushed up the changes introduced in #267, since using the header value as the fallback appears to be the better approach.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -19,7 +19,17 @@ public MessageConverter(Func<BasicDeliverEventArgs, string> messageIdStrategy)
...
         }
 
-        public string RetrieveMessageId(BasicDeliverEventArgs message) => messageIdStrategy(message);
+        public string RetrieveMessageId(BasicDeliverEventArgs message, Dictionary<string, string> headers)
+        {
+            var messageId = messageIdStrategy(message);
+
+            if (string.IsNullOrWhiteSpace(messageId) && !headers.TryGetValue(Headers.MessageId, out messageId))
+            {
+                throw new InvalidOperationException("The message id strategy did not provide a message id, and the message does not have an 'NServiceBus.MessageId' header.");
```
bording: The wording here should probably "ID" here instead of "id".

Though that means we have the same problem in the xml comments of `CustomMessageIdStrategy` as well. Maybe go ahead and clean up both now?

bording:
@adamralph Pushed up some wording tweaks. 

Looking at the `CustomMessageIdStrategy` xml comments, the wording there feels awkward. Think we should revise it any further?

adamralph:
@bording the latest two commits look good. I'm happy to merge this once the fixup! commit is squashed.

I agree that the wording around `CustomMessageIdStrategy` isn't great, but I think I'd prefer to deal with it outside this PR.

bording:
Fixup commit squashed.

PR #267, opened 2016-11-15
Alternate fix for custom message id strategy loop

bording:
Building on #262, here's how we could use the NSB message header instead of the default message id strategy.

Changing the processing order does mean that the poison message errors will never have a chance of including a message id, so I went ahead and removed that.

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
@@ -32,6 +32,48 @@
         }
 
         [Test]
+        public void Should_throw_exception_when_no_message_id_is_set()
+        {
+            var message = new BasicDeliverEventArgs
+            {
+                BasicProperties = new BasicProperties()
+            };
+
+            Assert.Throws<InvalidOperationException>(() => converter.RetrieveMessageId(message));
+        }
+
+        [Test]
+        public void Should_throw_exception_when_using_custom_strategy_and_no_message_id_is_returned()
+        {
+            var customConverter = new MessageConverter(args => "");
+
+            var message = new BasicDeliverEventArgs
+            {
+                BasicProperties = new BasicProperties()
+            };
+
+            Assert.Throws<InvalidOperationException>(() => customConverter.RetrieveMessageId(message));
+        }
+
+        [Test]
+        public void Should_fall_back_to_message_id_when_custom_strategy_returns_empty_string()
+        {
+            var customConverter = new MessageConverter(args => "");
+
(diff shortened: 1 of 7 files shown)


src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
```diff
@@ -52,23 +54,24 @@ public void Should_throw_exception_when_using_custom_strategy_and_no_message_id_
...
 
-            Assert.Throws<InvalidOperationException>(() => customConverter.RetrieveMessageId(message));
+            var headers = new Dictionary<string, string>();
+
+            Assert.Throws<InvalidOperationException>(() => customConverter.RetrieveMessageId(message, headers));
         }
 
         [Test]
-        public void Should_fall_back_to_message_id_when_custom_strategy_returns_empty_string()
+        public void Should_fall_back_to_message_id_header_when_custom_strategy_returns_empty_string()
```
bording: Maybe this should say "NServiceBus_message_id_header" instead?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -19,13 +19,13 @@ public MessageConverter(Func<BasicDeliverEventArgs, string> messageIdStrategy)
...
-        public string RetrieveMessageId(BasicDeliverEventArgs message)
+        public string RetrieveMessageId(BasicDeliverEventArgs message, Dictionary<string, string> headers)
         {
             var messageId = messageIdStrategy(message);
 
-            if (string.IsNullOrWhiteSpace(messageId))
+            if (string.IsNullOrWhiteSpace(messageId) && !headers.TryGetValue(Headers.MessageId, out messageId))
             {
-                messageId = DefaultMessageIdStrategy(message);
+                throw new InvalidOperationException("The registered message id strategy did not provide a message id, and the message does not have an 'NServiceBus.MessageId' header.");
```
bording: This error message could probably use a bit of tweaking.

bording:
Pushed up some wording fixup commits.

> What's our strategy for merging? Should this commit be squashed into the last commit from the fix-custom-message-id-strategy branch from #262? And then we merge #262 instead?

Probably best to close this PR, fast forward the #262 branch and clean up the commits to get everything ready.

adamralph:
@bording the fixup commits LGTM. I'm a bit surprised that we use lower case "id" in descriptive text rather than upper case, but if that is the convention then I won't argue with it.

👍 to your proposal for handling the merge.

adamralph:
Superseded by #262

PR #272, opened 2016-11-18
upgrade to R# CTL 2016.3 EAP9

adamralph:
The intermittent bug which causes inspection failure [seems to be fixed in this EAP](https://youtrack.jetbrains.com/issue/RSRP-461241#comment=27-1730891).

I'm not sure if we want to merge this in now or wait for RTM. It won't affect CI of course, but at least it will give us more reliable local inspections.

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -7,11 +7,12 @@
 using System;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using static SimpleTargets;
 
 // locations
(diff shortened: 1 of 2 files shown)


bording:
Oh, the irony is strong, build failed because of inspections again!

adamralph:
ROTFL I saw the red cross in the PR list and came in here and comment on the irony. You beat me to it!

adamralph:
Let's watch develop fail now... 😛

PR #273, opened 2016-11-18
ignore the databus folder... at last

adamralph:
This infuriating thing appears and pollutes my working directory every time I run the tests.

Why I didn't send this PR months ago I'll never know... `ლ(ಠ益ಠლ)`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ run-git.cmd
 src/Chocolatey/Build/*
 src/.nuget/NuGet.exe
 *.sln.inspections.xml
+databus
 
 installer/[F|f]iles
 installer/[C|c]ustom[A|a]ctions


bording:
While this will fix the problem in the short term, the real fix for this is:

1. Fix the tests upstream to use the NUnit work directory instead of the current directory when creating the databus folder.
2. Set the `--work` parameter of `nunit3-console.exe` calls in the build script to be equal to the directory that the test assembly is located.

This will also have the side benefit of making sure the individual `TestResult.xml` files for each test assembly will be preserved after a run of the script instead of being overwritten.

adamralph:
@bording I also get it when I run the tests from inside Visual Studio. Would point 1 fix that?

bording:
@adamralph What are you using to run from inside of VS? If I run any of the databus tests using the NUnit Test Runner from inside of VS, the tests fail because the current directory is set to a VS path in Program Files, which isn't writeable.

adamralph:
@bording I'm using TestDriven.NET. All the tests pass for me (as long as I reset the broker before running).

bording:
So I just noticed something interesting. I added a call to `System.IO.Directory.GetCurrentDirectory()` inside the test, and the value returned changes based on how I opened Visual Studio.

If I open it directly, and then open the solution, the value is:

`C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE`

But if I browse to the solution file in Explorer and double click it, the value is:

`C:\Code\NServiceBus.RabbitMQ\src`

So, the test will fail or pass based on how I opened VS!

Tests should never be depending on the current directory, so I'll definitely open a PR in core to fix this. I'm going to have it use the NUnit work directory instead, which seems to be the correct way of doing this from an NUnit perspective.

adamralph:
Withdrawing this, and waiting on tenterhooks for the core update 😄

PR #275, opened 2016-11-21
isolate tests on a "nsb-rabbitmq-test" virtual host

### LINKED ISSUE #271: Broker isolation for test suites
Currently, by default, we use the default `/` virtual host on `localhost`. This causes some problems, and potential problems:

- The tests may interfere with other endpoints running locally which use the default host
- The tests may be interfered with by other endpoints running locally which use the default host
- We cannot safely teardown all broker configuration that the tests rely on, whether pre-existing before the test runs or not. A non-empty broker seems to cause at least one (possibly intermittent) test failure when running the tests locally. (On the CI server, we drop and recreate the `/` virtual host between each test run.)

We should investigate using a custom virtual host by default, named in a way that is unlikely to clash with any other endpoints.

Building on top of this, we can investigate adding full teardown before/after running our tests. This could be pursued in a separate issue.

adamralph:
fixes #271 

I added an env var on the CI server so that it continues to use the `/` virtual host.

This also helps when running tests from Visual Studio. Before you run a test, you can drop to a command prompt and run `.\build.cmd prepare-virtual-host`. 😎

### DIFF
--- a/scripts/broker.csx
+++ b/scripts/broker.csx
@@ -1,0 +1,108 @@
+#r "System.Data"
+#r "../src/packages/RabbitMQ.Client.4.1.1/lib/net451/RabbitMQ.Client.dll"
+
+using System.Data.Common;
+using System.Net;
+using RabbitMQ.Client;
+
+public void DeleteVirtualHost()
+{
+    try
+    {
+        CreateVirtualHostRequest("DELETE", GetBroker()).GetResponse().Dispose();
+    }
+    catch (WebException ex) when ((ex?.Response as HttpWebResponse)?.StatusCode == HttpStatusCode.NotFound)
+    {
+    }
+}
+
+public void CreateVirtualHost() => CreateVirtualHostRequest("PUT", GetBroker()).GetResponse().Dispose();
+
+public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBroker()).GetResponse().Dispose();
+
+public Broker GetBroker()
+{
+    var connectionStringBuilder = new DbConnectionStringBuilder
+    {
+        ConnectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString")
+    };
+
+    ApplyDefault(connectionStringBuilder, "username", "guest");
+    ApplyDefault(connectionStringBuilder, "password", "guest");
+    ApplyDefault(connectionStringBuilder, "virtualhost", "nsb-rabbitmq-test");
+    ApplyDefault(connectionStringBuilder, "host", "localhost");
+
+    return new Broker
+    {
+        UserName = connectionStringBuilder["username"].ToString(),
+        Password = connectionStringBuilder["password"].ToString(),
+        VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
+        HostName = connectionStringBuilder["host"].ToString(),
+        Port = 15672,
+    };
+}
+
+public void ApplyDefault(DbConnectionStringBuilder builder, string key, string value)
+{
+    if (!builder.ContainsKey(key))
+    {
+        builder.Add(key, value);
+    }
+}
+
+public HttpWebRequest CreateVirtualHostRequest(string method, Broker broker) =>
+    CreateHttpWebRequest(
+        $"http://{broker.HostName}:{broker.Port}/api/vhosts/{Uri.EscapeDataString(broker.VirtualHost)}",
+        method,
+        broker);
+
+public HttpWebRequest CreateUserPermissionRequest(string method, Broker broker)
+{
+    var uriString = $"http://{broker.HostName}:{broker.Port}/api/permissions/{Uri.EscapeDataString(broker.VirtualHost)}/{Uri.EscapeDataString(broker.UserName)}";
+
+    var request = CreateHttpWebRequest(uriString, method, broker);
+
+    var bodyString =
+@"{
+    ""scope""       : ""client"",
+    ""configure""   : "".*"",
+    ""write""       : "".*"",
+    ""read""        : "".*""
+}";
+
+    var bodyBytes = new ASCIIEncoding().GetBytes(bodyString);
+
+    request.ContentLength = bodyBytes.Length;
+
+    using (var stream = request.GetRequestStream())
+    {
+        stream.Write(bodyBytes, 0, bodyBytes.Length);
+    }
+
+    return request;
+}
+
+public HttpWebRequest CreateHttpWebRequest(string uriString, string method, Broker broker)
+{
+    var request = HttpWebRequest.CreateHttp(uriString);
+
+    var encoded = Convert.ToBase64String(Encoding.GetEncoding("ISO-8859-1").GetBytes(broker.UserName + ":" + broker.Password));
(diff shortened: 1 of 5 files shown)


scripts/broker.csx
```diff
@@ -0,0 +1,111 @@
+#r "../src/packages/NServiceBus.6.0.0/lib/net452/NServiceBus.Core.dll"
+#r "../src/packages/NServiceBus.AcceptanceTesting.6.0.0/lib/net452/NServiceBus.AcceptanceTesting.dll"
```
adamralph: Required for [`RunSettings`](https://github.com/Particular/NServiceBus.RabbitMQ/pull/275/files#diff-4aacd7a8808fd077b2ab57dc39c949b1R34).

adamralph:
It seems like the tests themselves are still using the `/` host. I'm investigating...

adamralph:
🍾 A green build at last!

Given all the churn with the fixups, I squashed all the commits and rebased. This should be ready for review now.

The env var on the build server is now corrected to `RabbitMQTransport.ConnectionString`.

scripts/broker.csx
```diff
@@ -0,0 +1,104 @@
+#r "System.Data"
+#r "../src/packages/RabbitMQ.Client.4.1.1/lib/net451/RabbitMQ.Client.dll"
```
bording: Having to reference the file like this is unfortunate, since the next time we update the client we'll have to remember to manually update this.

scripts/broker.csx
```diff
@@ -0,0 +1,104 @@
...
+    {
+        Host = GetValueOrDefault(builder, "host", "localhost"),
+        Port = 15672,
+        VirtualHost = GetValueOrDefault(builder, "virtualhost", "nsb-rabbitmq-test"),
+        Username = GetValueOrDefault(builder, "username", "guest"),
+        Password = GetValueOrDefault(builder, "password", "guest"),
+    };
+}
+
+public static string GetValueOrDefault(DbConnectionStringBuilder builder, string key, string defaulValue)
```
bording: typo: defaulValue

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -78,6 +75,17 @@ ConnectionFactory CreateConnectionFactory()
         return connectionFactory;
     }
 
+    static string GetValueOrDefault(DbConnectionStringBuilder builder, string key, string defaulValue)
```
bording: typo: defaulValue

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
```diff
@@ -75,26 +71,32 @@ static void PurgeQueue(IModel channel, string queue)
...
+            VirtualHost = GetValueOrDefault(builder, "virtualhost", "nsb-rabbitmq-test"),
+            HostName = GetValueOrDefault(builder, "host", "localhost"),
             AutomaticRecoveryEnabled = true,
             UseBackgroundThreadsForIO = true
         };
 
         return connectionFactory;
     }
 
+    static string GetValueOrDefault(DbConnectionStringBuilder builder, string key, string defaulValue)
```
bording: typo: defaulValue

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -57,19 +59,14 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
...
 
         var connectionFactory = new ConnectionFactory
         {
-            UserName = username,
-            Password = password,
-            VirtualHost = virtualHost,
-            HostName = host,
+            UserName = GetValueOrDefault(builder, "username", "guest"),
+            Password = GetValueOrDefault(builder, "password", "guest"),
+            VirtualHost = GetValueOrDefault(builder, "virtualhost", "nsb-rabbitmq-test"),
```
bording: All three spots that use this to set a default value have a problem. If a user has set a connection string in the environment variable that does not contain a VirtualHost, then the tests will run without one (the default of `/`, but the code using this `GetValueOrDefault` setting will use `nsb-rabbitmq-test` instead.

This means that queue purging would be run against a different virtual host than the tests ran on.

adamralph:
@bording thanks or the review. Well spotted wrt to the virtual host disparity between tests and clean up.

I push another fixup! commit which should address all your concerns. I think I probably could have kept the original approach in the csx file for defaults, but I chose to use the same approach as the cs files for consistency. I also tweaked some of the naming in the csx file match that in the cs files.

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -20,12 +20,36 @@ class ConfigureScenariosForRabbitMQTransport : IConfigureSupportedScenariosForTe
...
+
+        if (!connectionStringBuilder.ContainsKey("virtualhost"))
+        {
+            connectionStringBuilder.Add("virtualhost", "nsb-rabbitmq-test");
+        }
+
+        if (!connectionStringBuilder.ContainsKey("host"))
+        {
+            connectionStringBuilder.Add("host", "localhost");
+        }
```
adamralph: If all these `if` blocks are distasteful, I can extract a method and collapse these down to four lines. E.g.

```C#
ApplyDefault(connectionStringBuilder, "username", "guest");
...
```

scripts/broker.csx
bording: At least when simple-targets is updated, you're probably thinking in the context of the build script already, vs a rabbit client update.

Not really anything to be done about it, but something to make sure we remember.

scripts/broker.csx
```diff
@@ -0,0 +1,115 @@
...
+
+public void CreateVirtualHost() => CreateVirtualHostRequest("PUT", GetBroker()).GetResponse().Dispose();
+
+public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBroker()).GetResponse().Dispose();
+
+public Broker GetBroker()
+{
+    var connectionStringBuilder = new DbConnectionStringBuilder
+    {
+        ConnectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString"),
```
bording: I'd prefer not to have the trailing comma here and in the other places where this has been copied.

bording: > If it was many items, I'd prefer to keep the trailing comma on the last line so that all lines have the same format and can be moved around easily.

I can understand that, though personally the comma still bothers me too much to be able to take advantage of that convenience. 😄

adamralph:
@bording I pushed another fixup! to remove the trailing commas.

adamralph:
@bording and another fixup! to introduce the `ApplyDefault` method...

bording:
Looks good, I think this ready to be squashed and merged.

adamralph:
Squashed.

PR #276, opened 2016-11-22
change nunit work folder to test directory

adamralph:


adamralph:
This will conflict with #275 and I'd prefer to get that in first, so marking this as WIP.

adamralph:
Rebased. This should be good to go now.

PR #277, opened 2016-11-22
simplify parameter to ConnectionConfiguration

adamralph:


src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -49,21 +48,20 @@ public void SetUp()
             routingTopology = new ConventionalRoutingTopology(true);
             receivedMessages = new BlockingCollection<IncomingMessage>();
 
-            var settings = new SettingsHolder();
-            settings.Set("NServiceBus.Routing.EndpointName", "endpoint");
+            var endpointName = "NServiceBus.Routing.EndpointName";
```
bording: To match previous behavior, this should be set to "endpoint" instead.

However, I think an even better option would be to remove `endpointName` completely and pass `ReceiverQueue` directly into the parser instead. That actually matches the real behavior better anyway.

src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionStringParserTests.cs
```diff
@@ -12,19 +11,18 @@ public class ConnectionStringParserTests
...
 
-        SettingsHolder settings;
+        string endpointName;
 
         [SetUp]
         public void Setup()
         {
-            settings = new SettingsHolder();
-            settings.Set("NServiceBus.Routing.EndpointName", "endpoint");
+            endpointName= "NServiceBus.Routing.EndpointName";
```
bording: This should be set to "endpoint" to match the previous behavior.

Instead, `endpointName` can become a const string, and we can drop the setup method entirely.

src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationTests.cs
```diff
@@ -13,10 +12,7 @@ public class ConnectionConfigurationTests
         [SetUp]
         public void Setup()
         {
-            var settings = new SettingsHolder();
-            settings.Set("NServiceBus.Routing.EndpointName", "endpoint");
-
-            defaults = new ConnectionConfiguration(settings);
+            defaults = new ConnectionConfiguration("NServiceBus.Routing.EndpointName");
```
bording: This should be using "endpoint" to match the previous behavior.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
adamralph: Oops, I used the wrong name! Will fix.

I'll make the `ReceiverQueue` change in a separate commit.

src/NServiceBus.RabbitMQ.Tests/ConnectionString/ConnectionStringParserTests.cs
adamralph: 👍 will fix the string and send the const change in a separate commit.

adamralph:
@bording correct string fixup! and further simplification pushed.

bording:
LGTM

Mind rebasing against latest develop before we merge?

adamralph:
Sure, will do.

adamralph:
Rebased.

adamralph:
Rebased again. (And remembered to squash this time!)

PR #278, opened 2016-11-22
Update build & test dependencies

bording:


bording:
By updating the API approvals stuff, this PR should have the side effect of immunizing us from the R# inspections bug, because the files that would fail inspections are now either gone or cleaned up!

PR #280, opened 2016-11-23
Refactor the resetting of queues and exchanges in RabbitMQContext

adamralph:

PR #281, opened 2016-11-23
Fix waiting for messages with RabbitMqContext

adamralph:
Unless I'm missing something, this is currently fundamentally flawed. If the timeout is exceeded, the current method returns a null and the calling test just assumes that return value isn't null and de-references it which would result in an NRE. This PR switches to explicit methods designed for each use case.

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
-    using System.Diagnostics;
     using System.Linq;
     using System.Threading.Tasks;
     using NUnit.Framework;
@@ -102,6 +101,6 @@
         BlockingCollection<IncomingMessage> receivedMessages;
         ConventionalRoutingTopology routingTopology;
 
-        static readonly TimeSpan incomingMessageTimeout = TimeSpan.FromSeconds(Debugger.IsAttached ? 600 : 1);
+        static readonly TimeSpan incomingMessageTimeout = TimeSpan.FromSeconds(1);
     }
 }
--- a/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
@@ -69,20 +69,25 @@
             channelProvider?.Dispose();
         }
 
-        protected IncomingMessage WaitForMessage()
+        protected bool TryWaitForMessageReceipt()
         {
-            var waitTime = TimeSpan.FromSeconds(1);
-
-            if (Debugger.IsAttached)
-            {
-                waitTime = TimeSpan.FromMinutes(10);
-            }
-
             IncomingMessage message;
-            receivedMessages.TryTake(out message, waitTime);
+            return TryReceiveMessage(out message, incomingMessageTimeout);
+        }
+
+        protected IncomingMessage ReceiveMessage()
+        {
+            IncomingMessage message;
+            if (!TryReceiveMessage(out message, incomingMessageTimeout))
+            {
+                throw new TimeoutException($"The message did not arrive within {incomingMessageTimeout.TotalSeconds} seconds.");
+            }
 
             return message;
         }
+
+        bool TryReceiveMessage(out IncomingMessage message, TimeSpan timeout) =>
+            receivedMessages.TryTake(out message, timeout);
 
         protected virtual IEnumerable<string> AdditionalReceiverQueues => Enumerable.Empty<string>();
 
@@ -96,5 +101,7 @@
         ChannelProvider channelProvider;
         BlockingCollection<IncomingMessage> receivedMessages;
         ConventionalRoutingTopology routingTopology;
+
+        static readonly TimeSpan incomingMessageTimeout = TimeSpan.FromSeconds(Debugger.IsAttached ? 600 : 1);
     }
 }
--- a/src/NServiceBus.RabbitMQ.Tests/When_consuming_messages.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/When_consuming_messages.cs
@@ -18,9 +18,9 @@
 
             await messageDispatcher.Dispatch(transportOperations, new TransportTransaction(), new ContextBag());
 
-            var received = WaitForMessage();
+            var receivedMessage = ReceiveMessage();
 
-            Assert.AreEqual(message.MessageId, received.MessageId);
+            Assert.AreEqual(message.MessageId, receivedMessage.MessageId);
         }
 
         [Test]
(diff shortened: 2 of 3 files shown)


adamralph:
Marked as WIP because it's based on the branch for #280. Once #280 is merged, I'll reform this branch to be just the last commit, based on develop.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -99,30 +69,39 @@ public void TearDown()
...
-        BlockingCollection<IncomingMessage> receivedMessages;
-
-        protected ConventionalRoutingTopology routingTopology;
         protected SubscriptionManager subscriptionManager;
+
+        private ChannelProvider channelProvider;
+        private BlockingCollection<IncomingMessage> receivedMessages;
+        private ConventionalRoutingTopology routingTopology;
+
+        private static readonly TimeSpan incomingMessageTimeout = TimeSpan.FromSeconds(Debugger.IsAttached ? 600 : 1);
```
bording: While I know this was here before, I'm not sure I see the value in extending the timeout time when the debugger is attached. In fact, when I was recently debugging some tests, this was more of a hindrance, because the test execution would just hang here.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -99,30 +69,39 @@ public void TearDown()
             channelProvider?.Dispose();
         }
 
-        protected IncomingMessage WaitForMessage()
+        protected bool TryWaitForMessageReceipt()
```
bording: This doesn't really feel like a Try* message to me, since the return value is the actual thing we care about, vs it being the side channel of it was successful or not. Perhaps something like `GetMessageReceivedResult` or even `WasMessageReceived`, though that feels more like property name than a method name.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -99,30 +69,39 @@ public void TearDown()
...
         protected ConnectionFactory connectionFactory;
-        private ChannelProvider channelProvider;
         protected MessagePump messagePump;
-        BlockingCollection<IncomingMessage> receivedMessages;
-
-        protected ConventionalRoutingTopology routingTopology;
         protected SubscriptionManager subscriptionManager;
+
+        private ChannelProvider channelProvider;
+        private BlockingCollection<IncomingMessage> receivedMessages;
```
bording: Given that BlockingCollection blocks, I wonder if we should replace this with something non-blocking now that we have async stuff everywhere. For example, we could use `BufferBlock<T>` from the [TPL Dataflow](https://www.nuget.org/packages/Microsoft.Tpl.Dataflow/) library instead.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
adamralph: The way I looked at it was: a method which waited for message receipt and timeout, throwing a `TimeOutException`, would be called `WaitForMessageReceipt`, and the method which tries to do that, and returns a `bool` to indicate whether the attempt succeeded would then be `TryWaitForMessageReceipt`. This follows the pattern of methods like `Get` and `TryGet`.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -99,30 +69,39 @@ public void TearDown()
...
             }
 
-            IncomingMessage message;
-            receivedMessages.TryTake(out message, waitTime);
-
             return message;
         }
 
+        private bool TryReceiveMessage(out IncomingMessage message, TimeSpan timeout) =>
+            receivedMessages.TryTake(out message, incomingMessageTimeout);
```
adamralph: Oops, yes, will fix. The reason that I wanted the timeout to be passed into this method, is that `ReceiveMessage()` throws an exception which mentions the timeout period, and it felt a little ugly spreading the knowledge of that field across caller and callee.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
adamralph: 👍 I like it, but perhaps in a follow up PR?

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
adamralph: Yep, I also questioned this but decided to leave it in place for now. I don't think we need it though, so I'll add another commit to remove it. We can always edit the value when we're debugging, if we need to.

adamralph:
@bording I pushed a fixup! plus another commit to remove the debugging timeout.

Still WIP for now, since I'm fairly sure it will need altering after other PR's have gone in.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
bording: I always think of the Try pattern as requiring an out parameter, though I suppose it strictly speaking doesn't need one. Is there an example in the BCL that doesn't?

So, in this scenario, to fit the pattern, the non-try method signature would be `void WaitForMessageReceipt()` and since the return type here is `void` the Try version would be `out void` which is meaningless, so you get `bool TryWaitForMessageReceipt()` instead.

Ok, I'm good with this then.

bording:
Once #280 is in, I'll want to glance over the revised PR, but otherwise I think this is ready to go.

adamralph:
Squashed and rebased. Should be good to go now.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -69,21 +68,26 @@ public void TearDown()
...
+                throw new TimeoutException($"The message did not arrive within {incomingMessageTimeout.TotalSeconds} seconds.");
             }
 
-            IncomingMessage message;
-            receivedMessages.TryTake(out message, waitTime);
-
             return message;
         }
 
+        private bool TryReceiveMessage(out IncomingMessage message, TimeSpan timeout) =>
```
bording: Looks like there's still a `private` here to be removed.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -96,5 +100,7 @@ protected IncomingMessage WaitForMessage()
         ChannelProvider channelProvider;
         BlockingCollection<IncomingMessage> receivedMessages;
         ConventionalRoutingTopology routingTopology;
+
+        private static readonly TimeSpan incomingMessageTimeout = TimeSpan.FromSeconds(1);
```
bording: And one more.

adamralph:
@bording well spotted. I pushed a fixup!

bording:
Looks good, should be ready for squashing.

PR #282, opened 2016-11-23
trim README.md and point to docs

adamralph:
My thinking here is that there's no point doing a weak repetition of docs in the README.md. Better to just point to the actual docs.

BTW - the maintainer group was never accessible publicly, so no point having it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,5 @@
 # RabbitMQ Transport for NServiceBus
 
-Install this to enable NServiceBus to facilitate messaging over RabbitMQ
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,24 +1,5 @@
 # RabbitMQ Transport for NServiceBus
 
-Install this to enable NServiceBus to facilitate messaging over RabbitMQ
+The [NServiceBus.RabbitMQ nuget package](https://www.nuget.org/packages/NServiceBus.RabbitMQ) provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/).
```
bording: Should be "NuGet" instead of "nuget"

adamralph:
@bording I pushed a fixup! to correct the casing.

adamralph:
squashed and rebased

Issue #283, opened 2016-11-23
Poison messages ack-ed and not forwarded to the error queue


tmasternak:
## Description
In version 3 of the transport whenever headers for a message being received cannot be properly [deserialized](https://github.com/Particular/NServiceBus.RabbitMQ/blob/3.5.0/src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs#L195) it gets ack-ed on the broker and not moved to the `error` queue.

I'm not sure what are the scenarios in which the header deserialization can fail but that could potentially lead to message loss.

adamralph:
Thanks for raising @tmasternak. FYI @bording is going to spike a potential fix for this.

adamralph:
For the 3.1.1, 3.2.1, 3.3.4 and 3.4.1 backports, do we need to title and label each PR as a bug, on each appropriate milestone?

adamralph:
I updated the description of this issue to add some more context.

PR #284, opened 2016-11-23
[SPIKE] Move poison messages to the error queue

bording:
This approach should get us able to solve #283, and make sure that poison messages get moved into the error queue.

If we do want to go forward with this, I'll close this PR and push up these changes to a hotfix branch.

### DIFF
--- a/src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
+++ b/src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
@@ -13,11 +13,12 @@
 
     class RabbitMqDequeueStrategy : IDequeueMessages, IDisposable
     {
-        public RabbitMqDequeueStrategy(IManageRabbitMqConnections connectionManager, RepeatedFailuresOverTimeCircuitBreaker circuitBreaker, ReceiveOptions receiveOptions)
+        public RabbitMqDequeueStrategy(IManageRabbitMqConnections connectionManager, RepeatedFailuresOverTimeCircuitBreaker circuitBreaker, ReceiveOptions receiveOptions, string errorQueue)
         {
             this.connectionManager = connectionManager;
             this.circuitBreaker = circuitBreaker;
             this.receiveOptions = receiveOptions;
+            this.errorQueue = errorQueue;
         }
 
         public void Init(Address address, TransactionSettings transactionSettings, Func<TransportMessage, bool> tryProcessMessage, Action<TransportMessage, Exception> endProcessMessage)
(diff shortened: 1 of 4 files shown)


src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
```diff
@@ -280,7 +279,26 @@ void Purge()
...
+        bool MovePoisonMessage(BasicDeliverEventArgs message, string queue)
+        {
+            var connection = connectionManager.GetPublishConnection();
+            var success = true;
+
+            using (var channel = connection.CreateModel())
+            {
+                try
+                {
+                    channel.BasicPublish("", queue, false, message.BasicProperties, message.Body);
```
adamralph: In 4.x, we seem to be passing `true` for `mandatory`. Is there a reason we're passing `false` here?

bording: In 3.x we don't actually have a way to check for `basic.return` messages, so there really isn't a point to setting it to true.

adamralph: I see, makes sense.

adamralph:
@bording I think this approach works, so I'm 👍 to convert into a non-spike, which actually fixes #283. Is there anything you think needs tidying? Nothing really leaps out at me. There's a bit of munging of `using` statements mixed into the commit, but I don't think that's a big deal.

src/NServiceBus.RabbitMQ/RabbitMqDequeueStrategy.cs
```diff
@@ -280,7 +279,26 @@ void Purge()
             }
         }
 
+        bool MovePoisonMessage(BasicDeliverEventArgs message, string queue)
+        {
+            var connection = connectionManager.GetPublishConnection();
+            var success = true;
```
adamralph: Perhaps it would be safer to set this to false, and only set it to true as the last statement in the `try` block? Or we could even remove it and do `return true` and `return false` in the appropriate places.

bording:
> @bording I think this approach works, so I'm 👍 to convert into a non-spike, which actually fixes #283. Is there anything you think needs tidying? Nothing really leaps out at me. There's a bit of munging of using statements mixed into the commit, but I don't think that's a big deal.

With the commit that changes how `success` gets set, I think this is ready to go. I'll go ahead and open a hotfix PR with the contents of this branch squashed.

adamralph:
#289 has been merged, and I confirmed that that branch was identical to this one.

PR #285, opened 2016-11-23
Fix inspection error in ApiGenerator.cs

bording:
This should get rid of the last inspection error that shows up when the R# inspections mess up on the CI server after updating to PublicApiGenerator 5.0.0.

I normally would try to avoid touching files in App_Packages, but this problem is too annoying!


adamralph:
Is there no way to tell inspectcode to ignore certain files/folders? Perhaps in https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ.sln.DotSettings ?

adamralph:
Oh, I just noticed this: https://github.com/Particular/NServiceBus.RabbitMQ/blob/3be3d9af4950ac2edef7fdd209720aed6ccc325f/src/NServiceBus.RabbitMQ.sln.DotSettings#L10

So I guess this is the bug? Sometimes it ignores that setting?

bording:
> So I guess this is the bug? Sometimes it ignores that setting?

Yes, I think that is the bug. Based on that setting, it's supposed to ignore files in App_Packages folders, but sometimes it doesn't.

PR #286, opened 2016-11-25
Revert "Fix inspection error in ApiGenerator.cs"

adamralph:
This reverts commit 86a3131ceabc7bd31d59e0f6a8978c43bf8e9d1b.

CI has been upgraded to 2016.3.20161124.94154-eap10 so we should be able to reset ApiGenerator back to the packaged version, which I think is better.

I'll run the build several times to make sure that we don't see the inspectcode bug in eap10.


adamralph:
I've run the build 8 times and I can't get inspectcode to fail, so I think it's safe to say that we can merge this, if we find it preferable to leave the source file unchanged, which I think makes it easier when upgrading to the next version.

PR #287, opened 2016-11-25
upgrade to R# CLT 2016.3.20161124.94154-eap10

adamralph:


bording:
If I'm reading the script correctly, there's nothing that would make sure that I'm using the new version in my local repo if there is already an older version of inspectcode.exe extracted into the local .resharper folder. Should have some sort of version number in the folder path?

adamralph:
@bording yes, that is true. The same applies for NuGet.exe, which is managed by `build.cmd`.

I did consider versioned local paths for each when initially writing the script, but decided against it. I think that decision was influenced by two things, a) I `git clean` my clone so regularly that any version of each never survives very long and b) `build.cmd` was largely copied from one of the MS repos (I can't remember which) (the original doesn't even version the `%LocalAppData%\.nuget` folder which I think is quite bad).

I guess we could add versioning to these local folders if you think it would help.

adamralph:
@bording I added another commit to add versioning to the local path.

I'll do the same for `Nuget.exe` but in another PR. There are also a couple of other things I'd like to tidy in `build.cmd` at the same time.

PR #288, opened 2016-11-25
add tasks for vscode

adamralph:
I'm using vscode more and more for lightweight changes, and this would be an awesome addition. I realise it's a tooling preference, but it shouldn't interfere with anything else.

This is what it provides:

image

(See the output window at the bottom, showing the results of the previous run.)


adamralph:
Withdrawing and will solve with local tooling for now.

PR #294, opened 2016-11-29
Protect against nulls in BasicPropertiesExtensions.Fill

### LINKED ISSUE #239: Message missing CorrelationId can cause retries to loop endlessly
@andreasohlund told me to open an issue here as we've been debugging a failed message that was retried from Service Pulse that got stuck in an endless loop, totally ignoring all recoverability settings.

It seems that if you dispatch a message from a 'Satellite' that a regular `IHandleMessages<...>` handles with  an exception, all the recoverability settings is applied like it should, like immediate retry/delayed retries. That message gets pushed to the error queue and picked up by ServicePulse.

If you THEN retry this message from ServicePulse and it once again fails in the handler, but now it just keeps retrying immediately forever and fills the logs blazingly fast ;)

After investigation I found that if a message is dispatched from a Satellite and you don't manually add `messageContext.Headers.Add(Headers.CorrelationId, messageContext.MessageId);` this issue is happening if the message fails somewhere and then retried from ServicePulse.

I can provide code samples where this is reproduced if needed.

bording:
Fixes #239

### DIFF
--- a/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
@@ -5,8 +5,8 @@
     using System.Globalization;
     using System.Linq;
     using System.Text;
-    using global::RabbitMQ.Client;
     using DeliveryConstraints;
+    using global::RabbitMQ.Client;
     using Performance.TimeToBeReceived;
 
     static class BasicPropertiesExtensions
--- a/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
@@ -18,12 +18,6 @@
                 properties.MessageId = message.MessageId;
             }
 
-            string correlationId;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -13,15 +13,18 @@ static class BasicPropertiesExtensions
     {
         public static void Fill(this IBasicProperties properties, OutgoingMessage message, List<DeliveryConstraint> deliveryConstraints)
         {
-            properties.MessageId = message.MessageId;
+            if (message.MessageId != null)
+            {
+                properties.MessageId = message.MessageId;
```
adamralph: I was initially confused as to why we need to do this, but then I decompiled the source and the side effect revealed itself: 

```C#
// RabbitMQ.Client.Framing.BasicProperties
public override string MessageId
{
    get
    {
        return this.m_messageId;
    }
    set
    {
        this.m_messageId_present = true;
        this.m_messageId = value;
    }
}
```

Note that I can't link to the source because the class is autogenerated 😖

PR #297, opened 2016-11-30
[WIP] make all tests projects use the same vhost

adamralph:
This fixes one problem mentioned in #295.

All tests projects should now behave in the same way, with or without the `RabbitMQTransport.ConnectionString` env var set.

Issue #295, opened 2016-11-29
Virtual Host isolation in tests has some problems


bording:
With the changes made in #275 to make the tests run on a separate virtual host, we have some inconsistent behavior. The Tests project handles the connection string differently from the other two tests projects.

If there is a connection string in the environment variable, the Tests project uses it exactly as is, but the other two projects (and the build script) will modify the value before using it.

For example, I have the following connection string set: "host=rabbit". For the Tests project the virtual host for the tests ends up being the default `/`, but for the other two projects, the lack of a default specified in the environment variable means that those tests run in the `nsb-rabbitmq-test` virtual host.

I'm thinking that we need to not be modifying the connection string specified in the environment variable, aligning all the behavior to how it is in the Tests project. Right now, if I try to run an acceptance test from inside VS, I get a failure because the `nsb-rabbitmq-test` virtual host doesn't exist yet. 

Another somewhat related issue is that if I were to specify `/` as my virtual host explicitly, then the tests would all work correctly from inside VS, but if I ran them through the build script, the build script would be deleting and recreating the default virtual host. While this appears to be allowed, it seems a bit weird to me.

adamralph:
I agree that we need to get rid of the inconsistencies.

IMHO all the tests should use `nsb-rabbitmq-test` by default and the guidance for running the tests within Visual Studio should be:

- Run `.\build.cmd prepare-virtual-host`
- Run the tests

adamralph:
Regardless of what the default host should be, I've had a refactoring floating around in my head for the past week or so which would simplify all of this. I'll have a play around and send a PR.

adamralph:
☝️ #297 is not the refactoring, it merely does the simplest thing possible to remove the inconsistency.

bording:
@adamralph Looking at the change introduced in #297, the more I think I'd prefer to go in the other direction. Here's what I propose:

- If no connection string is defined, stick to the default values as defined by `ConnectionConfiguration`.
- If a connection string is defined, then use it exactly as specified by the user.
- The build scripts virtual host tasks should check to see if the specified connection string has an explicit virtual host set (maybe even look specifically for `nsb-rabbitmq-test`) and only try to delete/create if it exists. Otherwise it should be a no-op.

This would mean that someone who clones the repo would immediately be able to run tests from Visual Studio if they had a newly configured rabbit install on their machine, and wouldn't have to first invoke the build script to make everything work.

[...]

PR #299, opened 2016-11-30
Rewrite connection string handling

### LINKED ISSUE #298: Report all invalid values in a connection string
Currently, if two bad values are passed, e.g. "host=abc;UseTls=**notabool**;requestedHeartbeat=**notaheartbeat**" we get an exception message like this:

> notabool is not a valid value for Boolean.

adamralph:
fixes #298


src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -62,14 +149,51 @@ void SetDefaultClientProperties(string endpointName)
...
+                { "machine_name", hostname },
+                { "user", userName },
+                { "endpoint_name", endpointName },
+            };
+
+            if (invalidOptionsMessage.Length > 0)
+            {
+                var message = invalidOptionsMessage.ToString().TrimEnd('\r', '\n');
+
+                Logger.Error(message);
```
adamralph: I'd love to get rid of this. We're throwing the exception up the stack so logging the message here feels inappropriate. By this logic we should log every exception message before we throw it. Doesn't make sense to me.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,145 @@
...
 
-        public Dictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
+        public string CertPath { get; }
 
-        public ConnectionConfiguration(string endpointName)
+        public string CertPassphrase { get; }
+
+        public IReadOnlyDictionary<string, string> ClientProperties { get; }
+
+        private ConnectionConfiguration(
```
bording: Drop the `private`.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,145 @@
...
-        public string CertPassphrase { get; set; }
+        public bool UseTls { get; }
 
-        public Dictionary<string, object> ClientProperties { get; } = new Dictionary<string, object>();
+        public string CertPath { get; }
 
-        public ConnectionConfiguration(string endpointName)
+        public string CertPassphrase { get; }
+
+        public IReadOnlyDictionary<string, string> ClientProperties { get; }
```
bording: Given that this is an internal class, I'd prefer to drop the interface and the readonly, and keep this as a Dictionary instead.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,145 @@
...
+            var port = GetValue(dictionary, "port", int.TryParse, useTls ? defaultTlsPort : defaultPort, invalidOptionsMessage);
+            var virtualHost = GetValue(dictionary, "virtualHost", "/");
+            var userName = GetValue(dictionary, "userName", "guest");
+            var password = GetValue(dictionary, "password", "guest");
+            var requestedHeartbeat = GetValue(dictionary, "requestedHeartbeat", ushort.TryParse, defaultRequestedHeartbeat, invalidOptionsMessage);
+            var retryDelay = GetValue(dictionary, "retryDelay", TimeSpan.TryParse, defaultRetryDelay, invalidOptionsMessage);
+            var certPath = GetValue(dictionary, "certPath", defaultCertPath);
+            var certPassPhrase = GetValue(dictionary, "certPassphrase", defaultCertPassphrase);
+
+            var host = default(string);
```
bording: Formatting: line break here please.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,145 @@
...
+                var hostsAndPorts = value.Split(',');
+
+                if (hostsAndPorts.Length > 1)
+                {
+                    invalidOptionsMessage.AppendLine("Multiple hosts are no longer supported. If using RabbitMQ in a cluster, consider using a load balancer to represent the nodes as a single host.");
+                }
+
+                var parts = hostsAndPorts[0].Split(':');
+                host = parts.ElementAt(0);
+                if (host.Length == 0)
```
bording: Formatting, line break before the `if` please

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,145 @@
...
+                {
+                    invalidOptionsMessage.AppendLine("Empty host name in 'host' connection string option.");
+                }
+
+                if (parts.Length > 1 && !int.TryParse(parts[1], out port))
```
bording: I'll need to study this section a bit more compared to the old `ParseHosts` method, but it feels like there might be a logic change here aside from the change to using `TryParse`.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -62,14 +149,51 @@ void SetDefaultClientProperties(string endpointName)
...
+                Logger.Error(message);
+
+                throw new NotSupportedException(message);
+            }
+
+            return new ConnectionConfiguration(
+                host, port, virtualHost, userName, password, requestedHeartbeat, retryDelay, useTls, certPath, certPassPhrase, clientProperties);
+        }
+
+        static string GetValue(IDictionary<string, string> dictionary, string key, string @default)
```
bording: I'd prefer to drop the interface here as well.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -62,14 +149,51 @@ void SetDefaultClientProperties(string endpointName)
...
+        {
+            string value;
+            return dictionary.TryGetValue(key, out value) ? value : @default;
         }
+
+        static T GetValue<T>(IDictionary<string, string> dictionary, string key, Convert<T> convert, T @default, StringBuilder invalidOptionsMessage)
```
bording: I'd prefer to drop the interface here as well.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
bording: Could we change the name of the default parameter to `defaultValue` to get rid of needing the `@` syntax?

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
bording: Same comment about the parameter here.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
adamralph: Yeah, come to think of it, I agree. I'll send a fixup!

The commit that added this has the message "made ConnectionConfiguration immutable" but when I squash I'll alter the message, since it won't be immutable anymore. That said, I'm not sure if we want to keep the micro commits or not before merging.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
adamralph: Done, but seems like GitHub doesn't realise it.

adamralph:
@bording the series of fixup!s should address all your comments.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
bording: The exception will end up getting logged anyway, right? If so, then yes this separate logging call is redundant.

adamralph: Exactly, or if not, it will go all the way up the stack and kill the process as unhandled. Regardless, logging it is not concern of this class. I'll add another commit to remove this.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
bording: Spent more time looking at this, and this looks fine. In fact, it looks like it handles even more parsing scenarios than the previous code did, so 👍.

src/NServiceBus.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -2,58 +2,143 @@
...
+            }
+
+            if (dictionary.ContainsKey("usepublisherconfirms"))
+            {
+                invalidOptionsMessage.AppendLine("The 'UsePublisherConfirms' connection string option has been removed. Use 'EndpointConfiguration.UseTransport<RabbitMQTransport>().UsePublisherConfirms' instead.");
+            }
+
+            var useTls = GetValue(dictionary, "useTls", bool.TryParse, defaultUseTls, invalidOptionsMessage);
+            var port = GetValue(dictionary, "port", int.TryParse, useTls ? defaultTlsPort : defaultPort, invalidOptionsMessage);
+            var virtualHost = GetValue(dictionary, "virtualHost", "/");
```
adamralph: Oops, no, just missed them in my excitement. Will fix.

adamralph:
@bording I pushed a fixup! to add the missing constants.

bording:
@adamralph I think this is looking good. 

Given that we're still discussing #295 / #297 do you want to base this on develop instead and go ahead and get it in, or wait until we have decided on how to handle the virtual host stuff?

adamralph:
@bording there is indeed no need for this branch to be based on the branch for #297 so I've rebased this branch on to develop and squashed the fixup! commits. The rebase was far from trivial so you might want to have another glance at the diff.

bording:
@adamralph Do you want to squash this any further, or would your prefer to merge with all the intermediate commits?

adamralph:
@bording I'm not really sure. Each commit is self contained (i.e. none of them are effectively fixup!s for previous commits) but they are very granular. One thing we could do is squash the refactoring commits together but keep the test/functional changes separate like so:

```
refactor: merge ConnectionStringParser into ConnectionConfiguration
green: test that host port takes precedence over separate port
refactor: move connection string validation to before configuration construction
red: report all invalid values in a connection string
green: report all invalid values in a connection string
refactor: ConnectionConfiguration internals
remove redundant logging
```

PR #300, opened 2016-11-30
Release Hotfix 3.1.1

### LINKED ISSUE #304: Poison messages ack-ed and not forwarded to the error queue
This is a backport of #283 for the 3.1 support branch.

bording:
Fixes #304

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,5 @@
+assembly-versioning-scheme: Major
+
+branches:
+  develop:
+    tag: alpha
--- a/GitVersionConfig.yaml
+++ b/GitVersionConfig.yaml
@@ -1,1 +1,0 @@
-assembly-versioning-scheme: Major
--- a/src/NServiceBus.RabbitMQ.Tests/ClusteringTests/ClusteredTestContext.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/ClusteringTests/ClusteredTestContext.cs
@@ -138,7 +138,7 @@
 
             publishChannel.Close();
             publishChannel.Dispose();
-            
+
             connectionManager.Dispose();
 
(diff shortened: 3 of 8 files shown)


adamralph:
I wonder if this branch has an old buggy version of gitversion, or bad configuration for it, since it seems to think it's version 4.3.0.

adamralph:
Indeed, [this commit](https://github.com/Particular/NServiceBus.RabbitMQ/commit/e9987b7d3ed25142900341f1357daf9e5e97f9d6) went in post 3.2.0.

bording:
> Indeed, this commit went in post 3.2.0.

Yeah, I was just looking at that. I'll have to update GitVersion on this branch and #301.

bording:
So, the tests are still failing, but now its from a -100 error:

```
Exception encountered unloading AppDomain
Agent Process was terminated successfully after error.
Process exited with code -100
```

I've tried rerunning them several times, and they appear to be consistently failing. This is happening on #301 as well. I suspect this is because `support-3.1` and `support-3.2` don't have the fixes that are in the 3.3.3 and later around cleaning up connections and preventing RabbitMQ open connection threads from keeping the process open (f7ab916958dcc5980ea463b8b2dcc55ffaf4c136 and d89bb8bc1bcac6d83d6658be30caf1bdedf81da5).

Seems like trying f7ab916958dcc5980ea463b8b2dcc55ffaf4c136 first makes sense it since only changes test code and wouldn't be a change the transport assembly behavior.

adamralph:
> Seems like trying f7ab916958dcc5980ea463b8b2dcc55ffaf4c136 first makes sense it since only changes test code and wouldn't be a change to the transport assembly behavior.

👍 sounds like a good course of action.

adamralph:
🎉

PR #310, opened 2016-12-05
[WIP] Update NServiceBus 6.1.0

danielmarbach:
Relates to https://github.com/Particular/NServiceBus/pull/4377


danielmarbach:
@Particular/rabbitmq-transport-maintainers once 6.1 is released it should be enough to update this PR.

adamralph:
Thanks @danielmarbach

timbussmann:
@Particular/rabbitmq-transport-maintainers 6.1 is now officially released. Feel free to pick up this PR and update it to the final 6.1 package. It's up to you whether you decide to only update the test projects or if you update the main projects build dependency to 6.1 too. In case of the second option, please be careful to update nuspec dependencies only if you make use of newly available 6.1 APIs.

adamralph:
IMHO we should update the test projects only. When the main projects are updated, the nuspec dependency should be updated too.

timbussmann:
@adamralph we discussed this and had different points of view which all had some good arguments. In the end I think this is up to the specific maintainers how they want to handle this. Personally I prefer the approach to only update the tests too as it completely eliminates the risk of accidentally relying on a 6.1 API (which is very unlikely though).

adamralph:
replaced by #316

PR #313, opened 2016-12-08
Implement new test connection strategy

adamralph:
~Currently based on the branch for #299, so still WIP.~

### DIFF
--- a/scripts/broker.csx
+++ b/scripts/broker.csx
@@ -20,32 +20,44 @@ public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBrok
 
 public Broker GetBroker()
 {
+    var connectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString");
+
+    if (string.IsNullOrEmpty(connectionString))
+    {
+        throw new Exception("The 'RabbitMQTransport.ConnectionString' environment variable is not set.");
+    }
+
     var connectionStringBuilder = new DbConnectionStringBuilder
     {
-        ConnectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString")
+        ConnectionString = connectionString
     };
 
-    ApplyDefault(connectionStringBuilder, "username", "guest");
-    ApplyDefault(connectionStringBuilder, "password", "guest");
-    ApplyDefault(connectionStringBuilder, "virtualhost", "nsb-rabbitmq-test");
-    ApplyDefault(connectionStringBuilder, "host", "localhost");
+    string hostName;
+    
+    object value;
+    if (connectionStringBuilder.TryGetValue("host", out value))
+    {
+        hostName = value.ToString();
+    }
+    else
+    {
+        throw new Exception("The connection string doesn't contain a value for 'host'.");
+    }
 
     return new Broker
     {
-        UserName = connectionStringBuilder["username"].ToString(),
-        Password = connectionStringBuilder["password"].ToString(),
-        VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
-        HostName = connectionStringBuilder["host"].ToString(),
+        UserName = connectionStringBuilder.GetOrDefault("username", "guest"),
+        Password = connectionStringBuilder.GetOrDefault("password", "guest"),
+        VirtualHost = connectionStringBuilder.GetOrDefault("virtualhost", "/"),
+        HostName = hostName,
         Port = 15672,
     };
 }
 
-public void ApplyDefault(DbConnectionStringBuilder builder, string key, string value)
+public static string GetOrDefault(this DbConnectionStringBuilder builder, string key, string defaultValue)
 {
-    if (!builder.ContainsKey(key))
-    {
-        builder.Add(key, value);
-    }
+    object value;
+    return builder.TryGetValue(key, out value) ? value.ToString() : defaultValue;
 }
 
 public HttpWebRequest CreateVirtualHostRequest(string method, Broker broker) =>
--- a/scripts/broker.csx
+++ b/scripts/broker.csx
@@ -27,10 +27,7 @@ public Broker GetBroker()
         throw new Exception("The 'RabbitMQTransport.ConnectionString' environment variable is not set.");
     }
 
-    var connectionStringBuilder = new DbConnectionStringBuilder
-    {
-        ConnectionString = connectionString
-    };
+    var connectionStringBuilder = new DbConnectionStringBuilder { ConnectionString = connectionString };
 
     string hostName;
     
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
@@ -31,10 +31,7 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
             throw new Exception("The 'RabbitMQTransport.ConnectionString' environment variable is not set.");
         }
 
-        connectionStringBuilder = new DbConnectionStringBuilder
-        {
-            ConnectionString = connectionString
-        };
+        connectionStringBuilder = new DbConnectionStringBuilder { ConnectionString = connectionString };
 
         configuration.UseTransport<RabbitMQTransport>().ConnectionString(connectionStringBuilder.ConnectionString);
 
@@ -82,11 +79,9 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
             throw new Exception("The connection string doesn't contain a value for 'host'.");
         }
 
-        connectionFactory.ClientProperties["purpose"] = "Test Queue Purger";
-
         var queues = await GetQueues(connectionFactory);
 
-        using (var connection = connectionFactory.CreateConnection())
+        using (var connection = connectionFactory.CreateConnection("Test Queue Purger"))
         using (var channel = connection.CreateModel())
         {
             foreach (var queue in queues)
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
@@ -24,34 +24,64 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
 
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
+        var connectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString");
+
+        if (string.IsNullOrEmpty(connectionString))
+        {
+            throw new Exception("The 'RabbitMQTransport.ConnectionString' environment variable is not set.");
+        }
+
         connectionStringBuilder = new DbConnectionStringBuilder
         {
-            ConnectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString")
+            ConnectionString = connectionString
         };
-
-        ApplyDefault(connectionStringBuilder, "username", "guest");
-        ApplyDefault(connectionStringBuilder, "password", "guest");
-        ApplyDefault(connectionStringBuilder, "virtualhost", "nsb-rabbitmq-test");
-        ApplyDefault(connectionStringBuilder, "host", "localhost");
 
         configuration.UseTransport<RabbitMQTransport>().ConnectionString(connectionStringBuilder.ConnectionString);
 
         return TaskEx.CompletedTask;
     }
 
-    static void ApplyDefault(DbConnectionStringBuilder builder, string key, string value)
-    {
-        if (!builder.ContainsKey(key))
-        {
-            builder.Add(key, value);
-        }
-    }
-
     public Task Cleanup() => PurgeQueues();
 
     async Task PurgeQueues()
     {
-        var connectionFactory = CreateConnectionFactory();
+        if (connectionStringBuilder == null)
+        {
+            return;
+        }
+
+        var connectionFactory = new ConnectionFactory
+        {
+            AutomaticRecoveryEnabled = true
+        };
+
+        object value;
+
+        if (connectionStringBuilder.TryGetValue("username", out value))
+        {
+            connectionFactory.UserName = value.ToString();
+        }
+
+        if (connectionStringBuilder.TryGetValue("password", out value))
+        {
+            connectionFactory.Password = value.ToString();
+        }
+
+        if (connectionStringBuilder.TryGetValue("virtualhost", out value))
+        {
+            connectionFactory.VirtualHost = value.ToString();
+        }
+
+        if (connectionStringBuilder.TryGetValue("host", out value))
+        {
+            connectionFactory.HostName = value.ToString();
+        }
+        else
+        {
+            throw new Exception("The connection string doesn't contain a value for 'host'.");
+        }
+
+        connectionFactory.ClientProperties["purpose"] = "Test Queue Purger";
 
         var queues = await GetQueues(connectionFactory);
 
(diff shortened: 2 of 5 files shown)


adamralph:
This is ready for review now.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
```diff
@@ -88,14 +82,32 @@ ConnectionFactory CreateConnectionFactory()
...
         {
-            UserName = connectionStringBuilder["username"].ToString(),
-            Password = connectionStringBuilder["password"].ToString(),
-            VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
-            HostName = connectionStringBuilder["host"].ToString(),
             AutomaticRecoveryEnabled = true,
             UseBackgroundThreadsForIO = true
         };
 
+        object value;
```
bording: The code here can use some of the same suggested changes as the acceptance tests, but we've got another problem here. Unlike the acceptance tests, the `Cleanup` method in the transport tests appears to be still be called even if an exception is thrown in the `Configure` method. That means that `connectionStringBuilder` can be null. We also can't guarantee we have a fully formed connection string either, since this code still runs even when `ConnectionConfiguration.Create` fails to validate it.

We should probably look into making an upstream change in the transport tests to change this behavior, but in the mean time, we should add something to `Configure` to make sure it ran to completion that we can check in `Cleanup`.

scripts/broker.csx
```diff
@@ -20,15 +21,26 @@ public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBrok
 
 public Broker GetBroker()
 {
+    var connectionString = Environment.GetEnvironmentVariable("RabbitMQTransport.ConnectionString");
+
+    if (string.IsNullOrEmpty(connectionString))
+    {
+        throw new Exception("The `RabbitMQTransport.ConnectionString` environment variable is not set.");
```
adamralph: Oops, that's a mistake. Will fix.

scripts/broker.csx
```diff
@@ -20,15 +21,26 @@ public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBrok
...
 
-    ApplyDefault(connectionStringBuilder, "username", "guest");
-    ApplyDefault(connectionStringBuilder, "password", "guest");
-    ApplyDefault(connectionStringBuilder, "virtualhost", "nsb-rabbitmq-test");
-    ApplyDefault(connectionStringBuilder, "host", "localhost");
+    var missingKeys = new[] { "username", "password", "virtualhost", "host" }
+        .Where(key => !connectionStringBuilder.ContainsKey(key))
+        .ToList();
+
+    if (missingKeys.Any())
```
adamralph: Yep, the reason I did that was that the user has to be explicitly added to the new virtual host, but I guess there's no reason we can't fall back to defaults any way, similar to the tests. I'll change.

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -76,13 +70,31 @@ ConnectionFactory CreateConnectionFactory()
...
-            VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
-            HostName = connectionStringBuilder["host"].ToString(),
             AutomaticRecoveryEnabled = true
         };
 
+        object value;
+
+        if (connectionStringBuilder.TryGetValue("username", out value))
+        {
+            connectionFactory.UserName = value?.ToString();
```
adamralph: If the value is missing from the connection string, then `connectionStringBuilder.TryGetValue` will return `false` so this statement will never execute. Note that we have to use `TryGetValue` since `Get`, or the indexer will throw an exception if the value is not present.

adamralph: The reason it works right now is that [RabbitMQ.Client.ConnectionFactory has the same defaults](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/63aeeb8d0b7908e3dcbd87905b5f37515ed6e2f6/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L126-L142).

This begs the question of whether we are unnecessarily repeating those defaults in `ConnectionConfiguration`.

One could argue that it insulates us from RabbitMQ.Client changing their defaults, but I'd argue that that would be such an enormous breaking change for the community that they'll never do it, and we also watch their releases with a close eye so if the ever did it, I think we'd spot the problem very early.

bording: So yeah, I completely forgot about the `ConnectionFactory` defaults. For many of the values, `ConnectionConfiguration` has the same default values as `ConnectionFactory`, so this is fine here. 

Looking at the `?.` operators, I'm wondering how likely it would be to actually get a null value out of the connection builder at this point, but I don't suppose it hurts anything.

So, at this point, I'd say this looks good and disregard my comments around defaults!

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -76,13 +70,31 @@ ConnectionFactory CreateConnectionFactory()
...
+        }
+
+        if (connectionStringBuilder.TryGetValue("password", out value))
+        {
+            connectionFactory.Password = value?.ToString();
```
bording: Disregard (see https://github.com/Particular/NServiceBus.RabbitMQ/pull/313#discussion_r92018622)

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -76,13 +70,31 @@ ConnectionFactory CreateConnectionFactory()
...
+        }
+
+        if (connectionStringBuilder.TryGetValue("virtualhost", out value))
+        {
+            connectionFactory.VirtualHost = value?.ToString();
```
bording: Disregard (see https://github.com/Particular/NServiceBus.RabbitMQ/pull/313#discussion_r92018622)

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -76,13 +70,31 @@ ConnectionFactory CreateConnectionFactory()
...
+        }
+
+        if (connectionStringBuilder.TryGetValue("host", out value))
+        {
+            connectionFactory.HostName = value?.ToString();
```
bording: Disregard (see https://github.com/Particular/NServiceBus.RabbitMQ/pull/313#discussion_r92018622)

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
bording: Avoiding the NRE is good, but I also think from a general correctness standpoint, we shouldn't be trying to execute the cleanup purging if we failed in setup and didn't actually run a test, so 👍 for changing behavior here.

adamralph:
@bording I pushed a fixup! to take care of the exception messages and the connection defaults in the build script.

I also added a minor refactoring commit.

adamralph:
@bording I pushed another fixup! to make the queue purging conditional on the `connectionStringBuilder` being set. If this object is set, then the methods have everything the need to execute, so I think this is probably the safest way to do this, and is insulated against other changes to the setup method. I added this in the acceptance tests too for safety, since there's nothing obvious there which tells us that the method is skipped of the setup fails, and we're also insulated against that changing in the future to match the transport tests behaviour. Nothing was required in the Tests project since NUnit's Setup and Teardown are used there, and Teardown will not be called if Setup fails.

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -24,34 +24,57 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
...
+        }
+
+        connectionFactory.ClientProperties["purpose"] = "Test Queue Purger";
```
bording: This line can be removed and the "Test Queue Purger" can be added to the `CreateConnection` call on line 81.

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -24,34 +24,57 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
...
         {
-            builder.Add(key, value);
+            return;
         }
-    }
 
-    public Task Cleanup() => PurgeQueues();
+        var connectionFactory = new ConnectionFactory
+        {
+            AutomaticRecoveryEnabled = true
```
bording: Should add `UseBackgroundThreadsForIO = true` here as well for consistency.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
```diff
@@ -55,7 +46,38 @@ public Task Cleanup()
...
+        {
+            AutomaticRecoveryEnabled = true,
+            UseBackgroundThreadsForIO = true
+        };
+
+        object value;
+
+        if (connectionStringBuilder.TryGetValue("username", out value))
+        {
+            connectionFactory.UserName = value?.ToString();
```
bording: I took a quick look at the DbConnectionStringBuilder code [here](https://github.com/Microsoft/referencesource/blob/master/System.Data/System/Data/Common/DbConnectionStringBuilder.cs), and it looks like there's no way for value to be null if it successfully retrieved a value. The stored value also always gets stored as a string, so we could just cast instead of call `ToString`.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
```diff
@@ -55,7 +46,38 @@ public Task Cleanup()
 
     void PurgeQueues()
     {
-        var connectionFactory = CreateConnectionFactory();
+        if (connectionStringBuilder == null)
```
bording: I'm not sure this is enough of a check to avoid all problems. If the connection string exists, but it fails validation at line 28, then `Cleanup` will still be called when the test didn't run, and we had some problem with the connection string we'll try to repeat here.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
```diff
@@ -55,7 +46,38 @@ public Task Cleanup()
...
+
+        if (connectionStringBuilder.TryGetValue("virtualhost", out value))
+        {
+            connectionFactory.VirtualHost = value?.ToString();
+        }
+
+        if (connectionStringBuilder.TryGetValue("host", out value))
+        {
+            connectionFactory.HostName = value?.ToString();
+        }
```
bording: For consistency in how we handle the default values for the transport, it seems like we should throw if we don't have a hosthame, since that would already be considered an incorrect connection string by the transport.

scripts/broker.csx
```diff
@@ -20,32 +20,29 @@ public void AddUserToVirtualHost() => CreateUserPermissionRequest("PUT", GetBrok
...
     return new Broker
     {
-        UserName = connectionStringBuilder["username"].ToString(),
-        Password = connectionStringBuilder["password"].ToString(),
-        VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
-        HostName = connectionStringBuilder["host"].ToString(),
+        UserName = connectionStringBuilder.GetOrDefault("username", "guest"),
+        Password = connectionStringBuilder.GetOrDefault("password", "guest"),
+        VirtualHost = connectionStringBuilder.GetOrDefault("virtualhost", "/"),
+        HostName = connectionStringBuilder.GetOrDefault("host", "localhost"),
```
bording: As I commented on the Transport tests, not sure we should be using localhost as a default here, since we don't do that in `ConnectionConfiguration`.

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
adamralph: I don't see an obvious way of doing that in line 81.

I can see that `CreateConnection()` has an overload which takes a `string clientProvidedName` so perhaps that's what you mean, but there's nothing in the API which indicates that that is the same as `ClientProperties["purpose"]`.

BTW - this is creeping the scope for this PR, since the current code looks like this:

```C#
var connectionFactory = new ConnectionFactory
{
    UserName = connectionStringBuilder["username"].ToString(),
    Password = connectionStringBuilder["password"].ToString(),
    VirtualHost = connectionStringBuilder["virtualhost"].ToString(),
    HostName = connectionStringBuilder["host"].ToString(),
    AutomaticRecoveryEnabled = true
};

connectionFactory.ClientProperties["purpose"] = "Test Queue Purger";
```

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
adamralph: This is the same as the current code, but I don't mind adding another commit to change this.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
adamralph: Something about this whole pattern doesn't sit right with me but I can't quite put my finger on it.

Anyway, I'll make a change to ensure that `Cleanup` only runs if `Initialize` succeeded.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
adamralph: I don't mind removing the `?` since it makes sense that the value can never be null if it exists (there's no way to represent a null value in the connection string).

I think I'd rather leave the `ToString` calls. `string.ToString()` just returns the current instance, so it's cheaper than a cast and it doesn't rely on knowledge of the internals of `DbConnectionStringBuilder`.

src/NServiceBus.Transport.RabbitMQ.TransportTests/ConfigureRabbitMQTransportInfrastructure.cs
adamralph: 👍 will fix.

scripts/broker.csx
adamralph: 👍 will fix.

adamralph:
@bording I've pushed two more commits (and rebased):

- change acceptance tests to UseBackgroundThreadsForIO
- a fixup! which
 - throws an exception if the connection string does not specify a host
 - removes the redundant Elvis operator when setting `ConnectionFactory` values
 - suppresses `Cleanup` for the transport tests when initialisation failed

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -24,34 +24,62 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
...
-            builder.Add(key, value);
+            return;
         }
-    }
 
-    public Task Cleanup() => PurgeQueues();
+        var connectionFactory = new ConnectionFactory
+        {
+            AutomaticRecoveryEnabled = true,
+            UseBackgroundThreadsForIO = true,
```
bording: The trailing comma here is inconsistent with the same code in the transport tests version. I know we disagree on if it should be there or not, so lets at least make it consistent!

src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -24,34 +24,62 @@ class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
...
+        if (connectionStringBuilder.TryGetValue("host", out value))
+        {
+            connectionFactory.HostName = value.ToString();
+        }
+        else
+        {
+            throw new Exception("The connection string doesn't contain a value for 'host'.");
+        }
+
+        connectionFactory.ClientProperties["purpose"] = "Test Queue Purger";
```
bording: This can be moved into the `CreateConnection` call on line 86, like it is done in the transport tests.

PR #314, opened 2016-12-09
isolate build packages and add versioning for local nuget.exe

adamralph:
Some ideas around how to clean up/improve the build command. Looking for opinion on this.

Switching to `-Verbosity quiet` when restoring the build packages gets rid of the annoying 

> All packages listed in packages.config are already installed.

...every time the build command is run, even when running something like `.\build.cmd -?`

However, I think it's better to have the regular NuGet.exe output when restoring the solution packages, and isolating the build packages from the solution packages allows us to do this.

Local versioning of NuGet.exe is something we've discussed before, and could be split into a separate PR  (probably ought to be a separate commit).

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -5,8 +5,9 @@
 cd %~dp0
 setlocal
 
-:: determine cache dir
+:: determine dirs
 set NUGET_CACHE_DIR=%LocalAppData%\.nuget\v3.4.4
+set NUGET_LOCAL_DIR=src\.nuget\v3.4.4
 
 :: download nuget to cache dir
 set NUGET_URL=https://dist.nuget.org/win-x86-commandline/v3.4.4/NuGet.exe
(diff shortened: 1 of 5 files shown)


build.cmd
```diff
@@ -17,12 +18,13 @@ if not exist %NUGET_CACHE_DIR%\NuGet.exe (
...
-if not exist src\.nuget\NuGet.exe (
-  copy %NUGET_CACHE_DIR%\NuGet.exe src\.nuget\NuGet.exe > nul
+if not exist %NUGET_LOCAL_DIR% md %NUGET_LOCAL_DIR%
+if not exist %NUGET_LOCAL_DIR%\NuGet.exe (
+  copy %NUGET_CACHE_DIR%\NuGet.exe %NUGET_LOCAL_DIR%\NuGet.exe > nul
 )
 
 :: restore packages
-src\.nuget\NuGet.exe restore .\src\NServiceBus.RabbitMQ.sln -MSBuildVersion 14
+%NUGET_LOCAL_DIR%\NuGet.exe restore .\packages.config -PackagesDirectory ./src/packages -MSBuildVersion 14 -Verbosity quiet
```
bording: I assume the new separate packages.config is to avoid restoring the items that are only used by TeamCity? Seems reasonable, but I'm wondering if there might be a better location/filename for the new file that makes it clearer it is for pulling down packages that the build script needs?

adamralph: I did in fact toy with having it named `build-packages.config` but I reverted that before sending the PR. Perhaps that's better?

bording: Either that, or perhaps move it into the scripts folder?

adamralph:
@bording is anything else required before merge?

bording:
At this point, the only thing would be to rebase against latest develop, if you've got a chance to do that. Otherwise I can't go ahead and merge as-is.

adamralph:
Rebased.

PR #316, opened 2016-12-13
[WIP] upgrade tests to NSB 6.1.0

adamralph:


adamralph:
OK, so we're down to the single test failure on `When_using_a_received_message_for_timeout.Timeout_should_be_received_after_expiration`

adamralph:
I'm seeing this in the output:

```
Moving message '0a855c20-c08f-4ca2-a2d0-a6db011d12c9' to the error queue 'error' because processing failed due to an exception: System.Exception: Saga can't be completed as it was updated by another process.
   at NServiceBus.InMemorySagaPersister.<>c__DisplayClass1_0.<Complete>b__0() in Z:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Persistence\InMemory\SagaPersister\InMemorySagaPersister.cs:line 29
   at NServiceBus.InMemoryTransaction.Commit() in Z:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransaction.cs:line 15
   at NServiceBus.InMemorySynchronizedStorageSession.CompleteAsync() in Z:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransactionalStorageSession.cs:line 31
   at NServiceBus.LoadHandlersConnector.<Invoke>d__1.MoveNext() in Z:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Pipeline\Incoming\LoadHandlersConnector.cs:line 49
```

adamralph:
@Particular/nservicebus-maintainers any ideas what might be causing this?

bording:
I've looked into this a bit, and there's definitely a problem here, but either the test itself is incorrect, or the in-memory saga persister changes have some problems.

If run this same test on MSMQ from the core solution, the test's Done condition stops running the test before in `InMemorySagaPersister.Complete` is even called, which is where the error here is coming from.

If I comment out the `TestContext.TimeoutReceived = true;` line, I end up seeing the exact same error on MSMQ as I do here.

Given that 6.1.0 has a new `InMemorySagaPersister` implementation, I believe we're seeing a problem with it that was being masked by the Done condition and not having any Asserts in this test.

@Particular/nservicebus-maintainers @Scooletz This appears to be something we need to follow up on.

timbussmann:
sounds like a normal race condition as the requested timeout is very short and will not go through the timeout manager persistence.

@adamralph @bording can you verify whether the tests succeeds if you increase the timeout delay to something > 5 seconds or if you limit concurrency to 1?

bording:
@timbussmann I tried the changing the concurrency, and that didn't make a difference, as we expected it might not based on our conversation.

Looking at the test a bit more, I'm not even sure what behavior it's trying to validate. It has a `Handle` and `Timeout` method for the same type, `StartSagaMessage`, so even though it's requesting a timeout in `Handle`, it doesn't matter because the initial message that starts the saga also invokes `Timeout`. That means the Done condition is triggered and the endpoint is shut down before the requested timeout comes back. This is why this test passes on MSMQ, it gets stopped before the timeout is processed.

I tried changing the timeout to 5 seconds, and then test did pass, but the test runs for only about 600ms before it passes, so that's clearly not right either. All this is doing is making RabbitMQ pass the test in the same way MSMQ is. By commenting out the `TestContext.TimeoutReceived = true;` line I get a better picture of what's actually going on here.

For the initial message that starts the saga, the method invocation sequence looks like this:

- `InMemorySagaPersister.Get(string ...)` overload
- `Handle`
- `InMemorySagaPersister.Save`
- `InMemorySagaPersister.Get(string ...)` overload
- `Timeout`

Note that even though `Timeout` contains `MarkAsComplete`, `InMemorySagaPersister.Complete` isn't called here. Not sure why that is.

When the timeout expires, this invocation is the problem, because it ends up doing the following:

- `InMemorySagaPersister.Get(Guid ...)` overload
- `Handle`
- `InMemorySagaPersister.Update`
- `InMemorySagaPersister.Get(Guid ...)` overload
- `Timeout`
- `InMemorySagaPersister.Complete` -> throws exception because update has been called.

This time `Complete` is called after `Timeout` and things go boom.

I also went back and looked at why this was passing with core 6.0, and the above sequence still occurs, but `InMemorySagaPersister.Complete` didn't throw an exception if called after an update like it does in 6.1.

So, from what I can tell, at the very least, the test seems broken in that the only way it can pass is if it stops executing before the timeout it requests actually gets processed. The bigger question: is the above behavior correct? If so, should core be preventing something here? Retrying this message doesn't seem like it would fix anything, because that just means `Handle` and `Timeout` will be called again, setting up the same failure.

andreasohlund:
I think the test is valid, this looks like a race condition/bug in the inmemory persister? Completing a saga from a timeout handler should be a supported scenario.

andreasohlund:
This is confirmed to be a core bug: 

https://github.com/Particular/NServiceBus/issues/4074#issuecomment-267259813

andreasohlund:
This turned out to be a core patch, see:

https://github.com/Particular/NServiceBus/issues/4397

bording:
Superseded by #318, which contains the core fix for the problem test in this PR.

PR #321, opened 2016-12-17
address inspectcode warnings

adamralph:
After these changes, only "suggestions" and warnings for the obsoletes are left:

image

I guess the obsolete warnings can be removed with some special Resharper incantation in a comment but not being a Resharper user I don't know what this should be.


bording:
Honestly, looking at that stuff, I'm not really sure any of it needs to be changed.

adamralph:
Generally I agree. However, I would like to find *some* way of reducing the warnings to zero. Whilst it's OK for us to review the current warnings and choose to ignore them, we should have a reliable way of recognising when new warnings appear.

adamralph:
Discussed in @Particular/rabbitmq-transport-maintainers sync. I will split this PR into the obvious "these should be fixed" violations, and those which are questionable.

PR #325, opened 2017-01-03
add a section on RabbitMQ.Client updates

adamralph:


MAINTAINERS.md
```diff
@@ -19,3 +19,7 @@ The maintainers [watch](https://github.com/Particular/NServiceBus.RabbitMQ/watch
 - A pull request created by a maintainer must be merged by *another* maintainer. No "self-merges".
 - The pull request must be made from a branch which is a straight line of commits from `develop`. There must be no merges in the branch history since the commit on `develop`.
   - The branch does not have to be based on the latest commit in `develop` but this is preferable, where practical.
+
+## RabbitMQ.Client Updates
+
+In order to reduce the burden of backporting patches to many versions of the package, when a new version of RabbitMQ.Client is released with breaking changes, we will update to that version in a patch release (hotfix on `master`).
```
bording: Should this say something like "if possible" to account for the case of a breaking change that requires us to make changes that would require more than a patch bump?

adamralph: 👍 I'll reword it.

adamralph: Hmmm... thinking this through a little more, I can't envisage any scenario where we'd have to make a minor bump in order to upgrade the client. I.e. we'd never *add a feature* in order to cater for a breaking change in the client.

I.e. if the client breaking change is not exposed via our API, then we'll make the necessary internal changes to call the new API in the client and then bump the patch version. If the breaking change *is* exposed via our API, we'd have to bump the major version.

adamralph: Anyhow, I'll reword this to at least allow the possibility of bumping more than the patch version, which I guess should be enough for this PR to be merged. We can investigate exactly what version bumps we may have to make separately, or even just as the need arises (i.e. next time we can't just bump the patch).

bording: For example, what if there was a change to any of the types we expose in our API? For example, the `IModel` interface that we expose on our `IRoutingTopology` interface. In that case, wouldn't we need to have more than a patch bump?

bording:
Ah, just realized I merged this without getting the fixup squashed first. Oh well. 😢

adamralph:
😱 never mind 😉 

@bording did you want a response to this question?

> For example, what if there was a change to any of the types we expose in our API? For example, the IModel interface that we expose on our IRoutingTopology interface. In that case, wouldn't we need to have more than a patch bump?

It's largely echoing what I said in my comment:

> If the breaking change is exposed via our API, we'd have to bump the major version.

bording:
@adamralph I originally took your comments as "why do we need to change the text", but reading back over them, I'm not sure why I thought that, so just consider my comment as agreeing with you. 😄

PR #326, opened 2017-01-14
Spelling corrected in transport transaction field

danielmarbach:
@Particular/rabbitmq-transport-maintainers please review and merge


bording:
Thanks for the correction @danielmarbach! How did you notice that?

danielmarbach:
I was looking into code which was copy-pasted from this transport which had the exact same spelling mistake :)

PR #329, opened 2017-01-27
Native delayed message delivery

bording:
## Things to-do before merging

- Run through the following test scenarios
  - [x] Spin up NSB v5 and NSB v6 endpoints. Send a delayed message from v6 to v5 and verify it's working
  - [x] Scenario (v6 to v6) where one endpoint disables non-native timeouts
  - [x] Scenario (v6 to v6)  where all endpoints disable non-native timeouts
  - [x] [More test scenarios](https://github.com/Particular/NServiceBus.RabbitMQ/pull/329#issuecomment-277240134)
- [x] Document native delays (PR raised: Particular/docs.particular.net/pull/2451)

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/APIApprovals.Approve.approved.txt
+++ b/src/NServiceBus.RabbitMQ.Tests/APIApprovals.Approve.approved.txt
@@ -15,6 +15,7 @@ namespace NServiceBus
             "mentedException. Will be removed in version 5.0.0.", true)]
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> CallbackReceiverMaxConcurrency(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, int maxConcurrency) { }
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> CustomMessageIdStrategy(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, System.Func<RabbitMQ.Client.Events.BasicDeliverEventArgs, string> customIdStrategy) { }
+        public static NServiceBus.Transport.RabbitMQ.DelayedDeliverySettings DelayedDelivery(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions) { }
         [System.ObsoleteAttribute("Replaced by NServiceBus.Callbacks package. The member currently throws a NotImple" +
             "mentedException. Will be removed in version 5.0.0.", true)]
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> DisableCallbackReceiver(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions) { }
@@ -38,6 +39,11 @@ namespace NServiceBus
 }
 namespace NServiceBus.Transport.RabbitMQ
 {
+    public class DelayedDeliverySettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
+    {
+        public NServiceBus.Transport.RabbitMQ.DelayedDeliverySettings AllEndpointsSupportDelayedDelivery() { }
+        public NServiceBus.Transport.RabbitMQ.DelayedDeliverySettings DisableTimeoutManager() { }
+    }
     public interface IDeclareQueues
     {
         void DeclareAndInitialize(RabbitMQ.Client.IModel channel, System.Collections.Generic.IEnumerable<string> receivingAddresses, System.Collections.Generic.IEnumerable<string> sendingAddresses);
@@ -50,6 +56,10 @@ namespace NServiceBus.Transport.RabbitMQ
         void Send(RabbitMQ.Client.IModel channel, string address, NServiceBus.Transport.OutgoingMessage message, RabbitMQ.Client.IBasicProperties properties);
         void SetupSubscription(RabbitMQ.Client.IModel channel, System.Type type, string subscriberName);
         void TeardownSubscription(RabbitMQ.Client.IModel channel, System.Type type, string subscriberName);
+    }
+    public interface ISupportDelayedDelivery
+    {
+        void BindToDelayInfrastructure(RabbitMQ.Client.IModel channel, string address, string deliveryExchange, string routingKey);
     }
 }
 namespace NServiceBus.Transports.RabbitMQ
--- a/src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
+++ b/src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
@@ -21,6 +21,8 @@
             using (var connection = connectionFactory.CreateAdministrationConnection())
             using (var channel = connection.CreateModel())
             {
+                DelayInfrastructure.Build(channel);
+
                 var queueDeclaringTopology = routingTopology as IDeclareQueues;
 
                 if (queueDeclaringTopology != null)
@@ -37,6 +39,16 @@
                     foreach (var sendingAddress in queueBindings.SendingAddresses)
                     {
                         CreateQueueIfNecessary(channel, sendingAddress);
+                    }
+                }
+
+                var delayTopology = routingTopology as ISupportDelayedDelivery;
+
+                if (delayTopology != null)
+                {
+                    foreach (var receivingAddress in queueBindings.ReceivingAddresses)
+                    {
+                        delayTopology.BindToDelayInfrastructure(channel, receivingAddress, DelayInfrastructure.DeliveryExchange, DelayInfrastructure.BindingKey(receivingAddress));
                     }
                 }
             }
--- a/src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs
@@ -1,0 +1,36 @@
+﻿namespace NServiceBus.Transport.RabbitMQ
+{
+    using Configuration.AdvanceExtensibility;
+    using Settings;
+
+    /// <summary>
+    /// The delayed delivery settings.
+    /// </summary>
+    public class DelayedDeliverySettings : ExposeSettings
+    {
+        internal DelayedDeliverySettings(SettingsHolder settings) : base(settings) { }
+
+        /// <summary>
+        /// Disables the timeout manager for this endpoint.
+        /// <para>
+        /// The timeout manager can be disabled once all preexisting timeouts stored in the persistence for this endpoint have expired.
+        /// </para>
+        /// </summary>
+        public DelayedDeliverySettings DisableTimeoutManager()
+        {
+            this.GetSettings().Set(SettingsKeys.DisableTimeoutManager, true);
+
+            return this;
+        }
+
+        /// <summary>
+        /// Tells the endpoint that all other endpoints can manage their own bindings to the delay infrastructure.
+        /// </summary>
+        public DelayedDeliverySettings AllEndpointsSupportDelayedDelivery()
+        {
+            this.GetSettings().Set(SettingsKeys.AllEndpointsSupportDelayedDelivery, true);
+
+            return this;
+        }
+    }
+}
--- a/src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
+++ b/src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
@@ -10,12 +10,16 @@ namespace NServiceBus.Transport.RabbitMQ
 
     class ConfirmsAwareChannel : IDisposable
     {
-        public ConfirmsAwareChannel(IConnection connection, IRoutingTopology routingTopology, bool usePublisherConfirms)
+        public ConfirmsAwareChannel(IConnection connection, IRoutingTopology routingTopology, bool usePublisherConfirms, bool allEndpointsSupportDelayedDelivery)
         {
             channel = connection.CreateModel();
             channel.BasicReturn += Channel_BasicReturn;
 
             this.routingTopology = routingTopology;
+
+            delayTopology = routingTopology as ISupportDelayedDelivery;
+            this.allEndpointsSupportDelayedDelivery = allEndpointsSupportDelayedDelivery;
+
             this.usePublisherConfirms = usePublisherConfirms;
 
             if (usePublisherConfirms)
@@ -50,7 +54,23 @@ namespace NServiceBus.Transport.RabbitMQ
                 task = TaskEx.CompletedTask;
             }
 
-            routingTopology.Send(channel, address, message, properties);
+            object delayValue;
+            if (properties.Headers.TryGetValue(DelayInfrastructure.DelayHeader, out delayValue) && delayTopology != null)
+            {
+                int startingDelayLevel;
+                var routingKey = DelayInfrastructure.CalculateRoutingKey((int)delayValue, address, out startingDelayLevel);
+
+                if (!allEndpointsSupportDelayedDelivery)
+                {
+                    delayTopology.BindToDelayInfrastructure(channel, address, DelayInfrastructure.DeliveryExchange, DelayInfrastructure.BindingKey(address));
+                }
+
+                channel.BasicPublish(DelayInfrastructure.LevelName(startingDelayLevel), routingKey, true, properties, message.Body);
+            }
+            else
+            {
+                routingTopology.Send(channel, address, message, properties);
+            }
 
             return task;
         }
@@ -225,6 +245,8 @@ namespace NServiceBus.Transport.RabbitMQ
 
         IModel channel;
         readonly IRoutingTopology routingTopology;
+        readonly ISupportDelayedDelivery delayTopology;
+        readonly bool allEndpointsSupportDelayedDelivery;
         readonly bool usePublisherConfirms;
         readonly ConcurrentDictionary<ulong, TaskCompletionSource<bool>> messages;
 
--- a/src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
+++ b/src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
@@ -5,6 +5,7 @@
     using System.Security.Cryptography.X509Certificates;
     using System.Text;
     using System.Threading.Tasks;
+    using DelayedDelivery;
     using global::RabbitMQ.Client.Events;
     using Janitor;
     using Performance.TimeToBeReceived;
@@ -18,6 +19,8 @@
         readonly ConnectionFactory connectionFactory;
         readonly ChannelProvider channelProvider;
         IRoutingTopology routingTopology;
+        readonly bool routingTopologySupportsDelayedDelivery;
+        readonly bool disableTimeoutManager;
 
         public RabbitMQTransportInfrastructure(SettingsHolder settings, string connectionString)
         {
@@ -31,18 +34,44 @@
 
             routingTopology = CreateRoutingTopology();
 
+            routingTopologySupportsDelayedDelivery = routingTopology is ISupportDelayedDelivery;
+            settings.Set(SettingsKeys.RoutingTopologySupportsDelayedDelivery, routingTopologySupportsDelayedDelivery);
+
             bool usePublisherConfirms;
             if (!settings.TryGet(SettingsKeys.UsePublisherConfirms, out usePublisherConfirms))
             {
                 usePublisherConfirms = true;
             }
 
-            channelProvider = new ChannelProvider(connectionFactory, routingTopology, usePublisherConfirms);
+            settings.TryGet(SettingsKeys.DisableTimeoutManager, out disableTimeoutManager);
+
+            bool allEndpointsSupportDelayedDelivery;
+            settings.TryGet(SettingsKeys.AllEndpointsSupportDelayedDelivery, out allEndpointsSupportDelayedDelivery);
+
+            channelProvider = new ChannelProvider(connectionFactory, routingTopology, usePublisherConfirms, allEndpointsSupportDelayedDelivery);
 
             RequireOutboxConsent = false;
         }
 
-        public override IEnumerable<Type> DeliveryConstraints => new[] { typeof(DiscardIfNotReceivedBefore), typeof(NonDurableDelivery) };
+        public override IEnumerable<Type> DeliveryConstraints
+        {
+            get
+            {
+                var constraints = new List<Type>
+                {
+                    typeof(DiscardIfNotReceivedBefore),
+                    typeof(NonDurableDelivery)
+                };
+
+                if (disableTimeoutManager)
+                {
+                    constraints.Add(typeof(DoNotDeliverBefore));
+                    constraints.Add(typeof(DelayDeliveryWith));
+                }
+
+                return constraints;
+            }
+        }
 
         public override OutboundRoutingPolicy OutboundRoutingPolicy => new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
 
@@ -62,7 +91,7 @@
         {
             return new TransportSendInfrastructure(
                 () => new MessageDispatcher(channelProvider),
-                () => Task.FromResult(StartupCheckResult.Success));
+                () => Task.FromResult(DelayInfrastructure.CheckForInvalidSetting(routingTopologySupportsDelayedDelivery, disableTimeoutManager)));
         }
 
         public override TransportSubscriptionInfrastructure ConfigureSubscriptionInfrastructure()
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -33,36 +33,44 @@
         public Dictionary<string, string> RetrieveHeaders(BasicDeliverEventArgs message)
         {
             var properties = message.BasicProperties;
+            var messageHeaders = properties.Headers;
 
-            var headers = DeserializeHeaders(properties.Headers);
+            if (messageHeaders != null)
+            {
+                //These headers need to be removed so that they won't be copied to an outgoing message if this message gets forwarded
+                messageHeaders.Remove(DelayInfrastructure.DelayHeader);
+                messageHeaders.Remove(DelayInfrastructure.DeadLetteredMessageHeader);
+            }
+
+            var deserializedHeaders = DeserializeHeaders(messageHeaders);
 
             if (properties.IsReplyToPresent())
             {
-                headers[Headers.ReplyToAddress] = properties.ReplyTo;
+                deserializedHeaders[Headers.ReplyToAddress] = properties.ReplyTo;
             }
 
             if (properties.IsCorrelationIdPresent())
             {
-                headers[Headers.CorrelationId] = properties.CorrelationId;
+                deserializedHeaders[Headers.CorrelationId] = properties.CorrelationId;
             }
 
             //When doing native interop we only require the type to be set the "fullName" of the message
-            if (!headers.ContainsKey(Headers.EnclosedMessageTypes) && properties.IsTypePresent())
+            if (!deserializedHeaders.ContainsKey(Headers.EnclosedMessageTypes) && properties.IsTypePresent())
             {
-                headers[Headers.EnclosedMessageTypes] = properties.Type;
+                deserializedHeaders[Headers.EnclosedMessageTypes] = properties.Type;
             }
 
             if (properties.IsDeliveryModePresent())
             {
-                headers[Headers.NonDurableMessage] = (properties.DeliveryMode == 1).ToString();
+                deserializedHeaders[Headers.NonDurableMessage] = (properties.DeliveryMode == 1).ToString();
             }
 
-            if (headers.ContainsKey("NServiceBus.RabbitMQ.CallbackQueue"))
+            if (deserializedHeaders.ContainsKey("NServiceBus.RabbitMQ.CallbackQueue"))
             {
-                headers[Headers.ReplyToAddress] = headers["NServiceBus.RabbitMQ.CallbackQueue"];
+                deserializedHeaders[Headers.ReplyToAddress] = deserializedHeaders["NServiceBus.RabbitMQ.CallbackQueue"];
             }
 
-            return headers;
+            return deserializedHeaders;
         }
 
         string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
--- a/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
+++ b/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
@@ -21,7 +21,7 @@
     /// <item><description>we generate an exchange for each queue so that we can do direct sends to the queue. it is bound as a fanout exchange</description></item>
     /// </list>
     /// </summary>
-    class ConventionalRoutingTopology : IRoutingTopology, IDeclareQueues
+    class ConventionalRoutingTopology : IRoutingTopology, IDeclareQueues, ISupportDelayedDelivery
     {
         readonly bool useDurableExchanges;
 
@@ -79,6 +79,11 @@
                 channel.QueueDeclare(address, useDurableExchanges, false, false, null);
                 Initialize(channel, address);
             }
+        }
+
+        public void BindToDelayInfrastructure(IModel channel, string address, string deliveryExchange, string routingKey)
+        {
+            channel.ExchangeBind(address, deliveryExchange, routingKey);
         }
 
         public void Initialize(IModel channel, string mainQueue)
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
@@ -15,7 +15,7 @@ using RabbitMQ.Client;
 
 class ConfigureScenariosForRabbitMQTransport : IConfigureSupportedScenariosForTestExecution
 {
-    public IEnumerable<Type> UnsupportedScenarioDescriptorTypes => new[] { typeof(AllDtcTransports), typeof(AllNativeMultiQueueTransactionTransports), typeof(AllTransportsWithMessageDrivenPubSub) };
+    public IEnumerable<Type> UnsupportedScenarioDescriptorTypes => new[] { typeof(AllDtcTransports), typeof(AllNativeMultiQueueTransactionTransports), typeof(AllTransportsWithMessageDrivenPubSub), typeof(AllTransportsWithoutNativeDeferral) };
 }
 
 class ConfigureEndpointRabbitMQTransport : IConfigureEndpointTestExecution
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/DelayedDelivery/When_deferring_a_message_longer_than_allowed_maximum.cs
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/DelayedDelivery/When_deferring_a_message_longer_than_allowed_maximum.cs
@@ -1,0 +1,50 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.AcceptanceTests.DelayedDelivery
+{
+    using System;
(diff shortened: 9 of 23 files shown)


src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
```diff
@@ -50,5 +53,46 @@ void CreateQueueIfNecessary(IModel channel, string receivingAddress)
 
             routingTopology.Initialize(channel, receivingAddress);
         }
+
+        void BuildDeferralInfrastructure(IModel channel)
+        {
+            const int maxLevel = 32 - 1;
```
WojcikMike: Do we want to expose this value somewhere to be configurable? Or just set it to 32 for now?

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
```diff
@@ -50,5 +53,46 @@ void CreateQueueIfNecessary(IModel channel, string receivingAddress)
...
+            for (int i = maxLevel; i >= 0; i--)
+            {
+                var level = i.ToString("D2");
+                var nextLevel = (i - 1).ToString("D2");
+
+                channel.ExchangeDeclare($"nsb.delay-level-{level}", ExchangeType.Topic, durableMessagesEnabled);
+
+                var arguments = new Dictionary<string, object>();
+                arguments.Add("x-queue-mode", "lazy");
+                arguments.Add("x-message-ttl", ((long)1 << i) * 1000);
```
WojcikMike: I know that IT people should understand it straightaway, but it took me 15 seconds to recall that `<<` is a shift operator and later on to understand that basically what we are doing is calculating 2^i. However replacing it with `Math.Pow(2, i)` would require casting to long. All in all I wonder if we should not write a little inline comment to express what is it. 

This is my braindump on this matter.

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
```diff
@@ -50,5 +53,46 @@ void CreateQueueIfNecessary(IModel channel, string receivingAddress)
...
+            bindingKey = "0.#";
+
+            for (int i = maxLevel; i >= 1; i--)
+            {
+                var level = i.ToString("D2");
+                var nextLevel = (i - 1).ToString("D2");
+
+                channel.ExchangeBind($"nsb.delay-level-{nextLevel}", $"nsb.delay-level-{level}", bindingKey);
+
+                bindingKey = "*." + bindingKey;
```
WojcikMike: Is there any reason why we are not using string interpolation? As we use it above for exchange/binding/queue names

src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
```diff
@@ -50,7 +52,29 @@ public Task SendMessage(string address, OutgoingMessage message, IBasicPropertie
                 task = TaskEx.CompletedTask;
             }
 
-            routingTopology.Send(channel, address, message, properties);
+            object delayValue;
+            if (properties.Headers.TryGetValue("NServiceBus.Transport.RabbitMQ.Delay", out delayValue))
+            {
+                //TODO: Do we need to remove the header?
+                properties.Headers.Remove("NServiceBus.Transport.RabbitMQ.Delay");
```
WojcikMike: I remember that we had a discussion with Andreas about headers, and basically we should try to limit leaving headers on the messages if not needed. So :+1: for removing it. But it is more of a preference than a need I think

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -79,6 +79,11 @@ public void DeclareAndInitialize(IModel channel, IEnumerable<string> receivingAd
                 channel.QueueDeclare(address, useDurableExchanges, false, false, null);
                 Initialize(channel, address);
             }
+
+            foreach (var address in receivingAddresses)
+            {
+                channel.ExchangeBind(address, "nsb.delay-destination", $"#.{address}");
```
WojcikMike: This blows up our Unit Tests. As this method is being created in here:
https://github.com/Particular/NServiceBus.RabbitMQ/blob/native-delays/src/NServiceBus.RabbitMQ.Tests/Support/ConventionalRoutingTopologyExtensions.cs#L23
And then ExchangeBind fails, as at that point this exchange doesn't exist.

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
SeanFeldman: Thought about it. Not convinced exposing it makes sense. If this would be a native feature, it would be "take it or leave it". Since we introduce a feature that is not found in the broker, we're free to dictate the rules. ASB hat on: don't make configurations too configurable. It will bite you back. And it's painful 😃

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
SeanFeldman: 👍 for inline comments and **not just here**.

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
SeanFeldman: Would string interpolation give us any benefit here?
If anything, a `StringBuilder` would be better as we concatenate 32+1 strings

src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
SeanFeldman: Great catch! @bording we forgot to clean it up as intended to do.

src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs
SeanFeldman: Sorry, failed to communicate. Haven't focused on tests yet.
Note to self: need to communicate better where we are and what is expected

src/NServiceBus.RabbitMQ/Administration/QueueCreator.cs
bording: The problem with using a `StringBuilder` here is that we need to use the intermediate results as well, so we'd be calling `ToString` on every loop iteration anyway.

src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
```diff
@@ -50,7 +52,29 @@ public Task SendMessage(string address, OutgoingMessage message, IBasicPropertie
...
+
+                var delay = (int)delayValue;
+                var bitArray = new BitArray(new[] { delay });
+                var sb = new StringBuilder();
+
+                for (int i = 31; i >= 0; i--)
+                {
+                    sb.Append(bitArray[i] ? "1." : "0.");
```
SeanFeldman: ❤️

WojcikMike:
> Spin up NSB v5 and NSB v6 endpoints. Send a delayed message from v6 to v5 and verify it's working 

With regards to this. Today I run 2 [delayed delivery samples](https://github.com/Particular/docs.particular.net/tree/master/samples/delayed-delivery)
! was run on v5 the other on v6. It seems that everything works as expected. The details of the test can be seen below:
 - Message consumer was running on v5. 
 - Message sender was running both on v5 and v6 - messages from both endpoints were consumed by consumer.
 - On sender v5 I scheduled a set of timeouts (`bus.Defer`) which results in timeout that resolves in send local, I killed this endpoint before the timeout kicks in
 - I run sender v6 and all timeouts were picked up by him and handled. 

All 3 endpoints were using Rabbit transport. Sender had NHibernate persistence configured. After that everything is working.

src/NServiceBus.RabbitMQ.Tests/APIApprovals.Approve.approved.txt
```diff
@@ -18,6 +18,7 @@ namespace NServiceBus
         [System.ObsoleteAttribute("Replaced by NServiceBus.Callbacks package. The member currently throws a NotImple" +
             "mentedException. Will be removed in version 5.0.0.", true)]
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> DisableCallbackReceiver(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions) { }
+        public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> DisableNonNativeTimeouts(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, bool allEndpoints) { }
```
WojcikMike: @bording @SeanFeldman I have troubles with this name. DisableNonNativeTimeouts sounds a bit like double negation. What if we would name it something along: `UseOnlyNativeTimeouts`

bording: Yeah, I agree that the naming still needs some work. Your suggestion sounds good.

SeanFeldman: 👍

src/NServiceBus.RabbitMQ/Administration/DelayInfrastructure.cs
```diff
@@ -7,7 +7,7 @@
     static class DelayInfrastructure
     {
         const int maxLevel = 32 - 1;
-        public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.Delay";
+        public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
```
SeanFeldman: 👍 explicit and clear

src/NServiceBus.RabbitMQ/Administration/DelayInfrastructure.cs
```diff
@@ -8,6 +8,7 @@ static class DelayInfrastructure
     {
         const int maxLevel = 32 - 1;
         public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
+        public const string DelayDestination = "nsb.delay-destination";
```
SeanFeldman: `DelayDestinationExchangeName`? otherwise it can be confused with the actual destination

src/NServiceBus.RabbitMQ/Administration/DelayInfrastructure.cs
```diff
@@ -10,14 +10,16 @@ static class DelayInfrastructure
         public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
         public const string Destination = "nsb.delay-destination";
 
+        public static string LevelName(int level) => $"nsb.delay-level-{level:D2}";
```
SeanFeldman: clever!

bording:
FYI I went ahead and rebased the branch after merging #336.

WojcikMike:
After @bording fixed the version, I've run the tests in the following way:
 - Before every test run I removed delayed-destination exchange to make sure that no leftovers binding impact the tests
 - I used sample related to DelayedDelivery (from the doco)
 - I've run the following scenarions:
  - Sender v6 -> Consumer v6 on default settings - everything ok
  - Sender v6 -> Consumer v6 where sender has DisabledTImeoutManager - everything ok
  - Sender v6 -> Consumer v6 where sender and consumer have set DisabledTImeoutManager - everything ok
  - Sender v6 -> Consumer v6 where sender and consumer have used `AllEndpointsSupportNativeDelays`- everything ok
  - Sender v6 -> Consumer v5 on default settings - everything ok
  - Sender v6 -> Consumer v5 where sender has DisabledTImeoutManager - everything ok

I've marked the tests in this PR as done, if you think that I missed some scenario let me know

src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// TODO: Fill this out
+    /// </summary>
+    public class DelayedDeliverySettings : ExposeSettings
+    {
+        internal DelayedDeliverySettings(SettingsHolder settings) : base(settings) { }
+
+        /// <summary>
+        /// TODO: Fill this out
+        /// </summary>
+        public void DisableTimeoutManager()
```
WojcikMike: Do we want to make this a FluentAPI? I assume that both settings can be set at the same time. So in an endpoint can have `DisableTimeoutManager` and `AllEndpointsSupportNativeDelays` at the same time. But our current API forces us to use:
```
var transport = endpointConfiguration.UseTransport<RabbitMQTransport>().ConnectionString("...");
transport.DelayedDelivery().DisableTimeoutManager();
transport.DelayedDelivery().AllEndpointsSupportNativeDelays();
```
It is not as smooth as it could be. We could use fluent API to make it a one liner as:
```
endpointConfiguration.UseTransport<RabbitMQTransport>().ConnectionString("...").DelayedDelivery().DisableTimeoutManager().AllEndpointsSupportNativeDelays();
```

I am not a great fan of FluentAPI but I want to simplify setting both fields for one endpoint. Fluent API is used in other transports like:
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/ccf2f9ed96b433716e84c00972c16fa393968c60/src/Transport/Config/ExtensionPoints/AzureServiceBusMessagingFactoriesSettings.cs
or
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/ccf2f9ed96b433716e84c00972c16fa393968c60/src/Transport/Config/ExtensionPoints/AzureServiceBusMessageReceiverSettings.cs

@bording @SeanFeldman what do you think?

SeanFeldman: We thought about it @WojcikMike and came to conclusion that fluent API while feels great is not what we'd need. `DisableTimeoutManager` and `AllEndpointsSupportNativeDelays` are not ordered. Fluent API makes sense when you need an order of operations. That's why we ended up with grouping under the `.DelayedDelivery()` scope. 

@bording please correct me if I'm wrong or missed something.

WojcikMike: @SeanFeldman I agree with the order, but what I don't like is writing configuration in 3 lines when It can be done in only 1. In the examples that I send there is no ordering enforced. I am not saying that FluentAPI is what we need. But now we make the configuration a bit cumbersome

SeanFeldman: @WojcikMike fluent apis are discouraged afaik. NSB v4 had a fluent api that was dropped. @bording do you remember what was the thought on fluent apis that steered away from using it?

I suggested another option (below), but we chose not to go that path. We could review configuration approach again if needed.

```c#
var transport = endpointConfig.UseTransport<RabbitMQTransport>().ConnectionString("...");
transport.DelayedDelivery(config => 
{
  config.DisableTimeoutManager();
  config.AllEndpointsSupportNativeDelays();
});

The approach above could be somewhat more aligned with other configurations we see in Core nowadays. Thoughts?

Comment (WojcikMike) [inline src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs]: I understand that we discourage fluent api, as it is no longer trendy and sexy. Let's ignore the fact that we still have fluent api in our code for a moment.

Let's make a step back and put a problem statement. Currently to set both options `DisableTimeoutManager` and `AllEndpointsSupportNativeDelays` person need to write 3 lines of configuration code with some repeated parts. I don't find it intuitive nor easy to use. When I used it I had an impression that I can set only 1 of the 2. I am open to other suggestions, but current state I don't particularly like

Comment (bording) [inline src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs]: It wouldn't be so bad to modify those new methods to let you chain them I suppose, seeing how we do already let you do that for the other settings.

Comment (SeanFeldman) [inline src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs]: I'm good with either decision (scoping, fluent, or func).

**Comment by bording** [comment]
After trying out keeping the `x-death` header around and moving it to another header, I came to the conclusion that it wasn't providing enough value to be worth it. Instead, I'm just removing it from incoming messages if it exists.

@WojcikMike I've also pushed up a change to the settings that allows them to be chained. Let me know what you think.

**Comment by WojcikMike** [comment]
Awesome I like it.

File: src/NServiceBus.RabbitMQ/Administration/DelayInfrastructure.cs [hunk 12]
```diff
@@ -8,7 +8,9 @@
 
     static class DelayInfrastructure
     {
-        const int maxLevel = 32 - 1;
+        const int maxNumberOfBitsToUse = 28;
```
WojcikMike: And why did we change it to 28?

src/NServiceBus.RabbitMQ.AcceptanceTests/DelayedDelivery/When_Deferring_a_message_longer_than_allowed_maximum.cs
```diff
@@ -0,0 +1,59 @@
...
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_Deferring_a_message_longer_than_allowed_maximum : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_throw()
+        {
+            var delay = TimeSpan.FromDays(365 * 10);
```
WojcikMike: Can we use `TimeSpan.MaxValue` in this place. As current code assumes that we are using 28 bits as a delay level which translates to more than 8 years. This value is to big as it represents 10 years. But if we would like to increase 28 bit level to something higher we need to remember to modify this test as well. I would suggest to either take max value, or to somehow expose the bit level value so that we can refer to it in tests.

src/NServiceBus.RabbitMQ/Administration/DelayInfrastructure.cs
bording: We realized that we were encoding 32 bits/levels, but representing the delay as in `int` means we only have 31 bits available (because of the sign bit). We could either change the code to use a `uint` for the delay, or decrease the number of bits we're using.

We decided to decrease the number of bits, and while were were doing that, we thought it made sense to reevaluate how many we actually need. 28 bits gets us about 8 years worth of delay, so that seems like a reasonable upper bound.

However, we can always tweak this before merging, up to a max of 31 bits, or 32 if we make additional changes to accommodate a `uint`. We do have the idea of a negative delay, if we're told to delay into the past, so that would also have to be handled if we switched to `uint`.

src/NServiceBus.RabbitMQ.AcceptanceTests/DelayedDelivery/When_Deferring_a_message_longer_than_allowed_maximum.cs
adamralph: Currently we're going with 28 bits which gives us a max 8 years, if we did decide to go with 31 bits (the max unless we switch from `int` to `long`) then anything above 64 years is enough to future proof this test.

SeanFeldman: 👍 test was modified to test with 100 years.

bording:
With the latest commits I've pushed up, the only thing left to do here would be to make changes to how we're running different permutations of tests, since disabling the timeout manager does impact how the endpoint runs.

However, I also think this is something that could be done in a different PR, since we more generally need to improve the testing setup, so I'm going to go ahead and remove the WIP from this.

bording:
Once this is reviewed, we can decide on how to squash things down before merging it.

bording:
Ok, so I guess I lied about being done! I realized that I could actually just remove the core behavior that reroutes things to the timeout manager instead of having to work around the results of that behavior.

For some reason I had initially ruled out doing this, but it makes things much simpler.

While I was making changes, I went ahead and made a few more tweaks to clean things up a bit more.

src/NServiceBus.RabbitMQ/Configuration/DelayedDeliverySettings.cs
```diff
@@ -21,7 +24,11 @@ public DelayedDeliverySettings DisableTimeoutManager()
         }
 
         /// <summary>
-        /// TODO: Fill this out
+        /// Tells the endpoint that all other endpoints also support native delays.
+        /// <para>
+        /// If this is called, the endpoint will no longer bind destination addresses to the delay
```
WojcikMike: I had to read this sentence 2 times to understand what it is about. What if we would write something as follows:
> If this is called, the endpoint will assume that all other endpoints use native delays and manage their own bindings to delay infrastructure. 

I don't know if this should be the place where we tell everyone that sending endpoint may create bdinging to the destination if this options is not set. We should describe it in doco, and keep this description simple. @bording what do you think?

bording: @WojcikMike Yeah, I can see how going for something simpler makes sense. I'll push up a change.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            channel.ExchangeDeclare(DeliveryExchange, ExchangeType.Topic, true);
+            channel.ExchangeBind(DeliveryExchange, LevelName(0), bindingKey);
+        }
+
+        public static void TearDown(IModel channel)
+        {
+            channel.ExchangeDelete(DeliveryExchange);
+
+            for (var i = maxLevel; i >= 0; i--)
```
adamralph: Same comment regarding `i`.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (delayInSeconds < 0)
+            {
+                delayInSeconds = 0;
+            }
+
+            var bitArray = new BitArray(new[] { delayInSeconds });
+            var sb = new StringBuilder();
+            startingDelayLevel = 0;
+
+            for (var i = maxLevel; i >= 0; i--)
```
adamralph: Same comment regarding `i`.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -62,7 +86,7 @@ public override TransportSendInfrastructure ConfigureSendInfrastructure()
         {
             return new TransportSendInfrastructure(
                 () => new MessageDispatcher(channelProvider),
-                () => Task.FromResult(StartupCheckResult.Success));
+                () => Task.FromResult(DelayInfrastructure.CheckForInvalidSetting(routingTopology is ISupportNativeDelays, disableTimeoutManager)));
```
adamralph:
```c#
routingTopology is ISupportNativeDelays
```

This Boolean is calculated in the constructor. Is it worth caching it in a field?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -33,36 +33,44 @@ public string RetrieveMessageId(BasicDeliverEventArgs message, Dictionary<string
         public Dictionary<string, string> RetrieveHeaders(BasicDeliverEventArgs message)
         {
             var properties = message.BasicProperties;
+            var headers = properties.Headers;
```
adamralph: The existence of both `headers` and `messageHeaders` hurts readability a little. I'd probably prefer if this variable didn't exist and the deserialized headers variable remained named `headers`.

src/NServiceBus.RabbitMQ/Routing/ISupportNativeDelays.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace NServiceBus.Transport.RabbitMQ
+{
+    using global::RabbitMQ.Client;
+
+    /// <summary>
+    /// An extension for <see cref="IRoutingTopology"/> implementations which allows
+    /// for receiving messages from the delay infrastructure.
```
adamralph: Since the interface has the word "native" in it, should it also appear somewhere in this summary? I.e. the name of the interface suggests that there is also something called "non-native" delays.

bording: Good point. How about changing the interface to `ISupportDelayedDelivery` instead?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: The primary reason around having the `headers` variable here is to access the `Headers` property once and store it instead of accessing the property four separate times. If `headers` had a different name, would that be enough to address your concern?

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        public static string LevelName(int level) => $"nsb.delay-level-{level:D2}";
+
+        public static string BindingKey(string address) => $"#.{address}";
+
+        public static void Build(IModel channel)
+        {
+            var bindingKey = "1.#";
+
+            for (var i = maxLevel; i >= 0; i--)
```
bording: `for` loops are the one place where I tend to prefer sticking with `i`, `j`, etc for variable names, but that's more out of tradition than anything. Seems like `level` could work here (and the other loops) well enough if you'd prefer to have them changed.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
```diff
@@ -0,0 +1,111 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Text;
+    using global::RabbitMQ.Client;
+
+    static class DelayInfrastructure
+    {
+        const int maxNumberOfBitsToUse = 28;
+        const int maxLevel = maxNumberOfBitsToUse - 1;
+        public const int MaxDelayInSeconds = (1 << maxNumberOfBitsToUse) - 1;
```
bording: I agree. I must have overlooked that, because that's something I typically do automatically.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
bording: I can see it either way. We only need it in those two places at startup, so is adding an extra bool to the class to keep the value around worth it?

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -30,19 +31,42 @@ public RabbitMQTransportInfrastructure(SettingsHolder settings, string connectio
...
+            {
+                var constraints = new List<Type>
+                {
+                    typeof(DiscardIfNotReceivedBefore),
+                    typeof(NonDurableDelivery)
+                };
+
+                if (disableTimeoutManager)
+                {
+                    constraints.Add(typeof(DelayedDelivery.DelayedDeliveryConstraint));
```
bording: This is the constraint that core looks at in the [`TimeoutManager` feature's prerequisites ](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L24) to see if it should be enabled. By indicating that we support delayed delivery, the timeout manager is disabled.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
```diff
@@ -0,0 +1,111 @@
...
+                channel.ExchangeBind(nextLevel, currentLevel, bindingKey);
+
+                bindingKey = "*." + bindingKey;
+            }
+
+            channel.ExchangeDeclare(DeliveryExchange, ExchangeType.Topic, true);
+            channel.ExchangeBind(DeliveryExchange, LevelName(0), bindingKey);
+        }
+
+        public static void TearDown(IModel channel)
```
bording: The idea here was that everything related to how to construct (or tear down) the delay infrastructure would live here. I kind of liked the idea of having one place to update if something changed instead of having to remember that some test code would also need to be changed. Now that we've got this done, I don't know how much changing we'd be doing, but I'm still not sure what I think about moving it.

Since the test project already has access to internals, if we made `maxLevel` public, then it could be referenced there if you think it really should be in the test project instead.

src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
```diff
@@ -50,7 +54,23 @@ public Task SendMessage(string address, OutgoingMessage message, IBasicPropertie
...
+            {
+                int startingDelayLevel;
+                var routingKey = DelayInfrastructure.CalculateRoutingKey((int)delayValue, address, out startingDelayLevel);
+
+                if (!allEndpointsSupportNativeDelays)
+                {
+                    delayTopology.BindToDelayInfrastructure(channel, address, DelayInfrastructure.DeliveryExchange, DelayInfrastructure.BindingKey(address));
+                }
+
+                channel.BasicPublish(DelayInfrastructure.LevelName(startingDelayLevel), routingKey, properties, message.Body);
```
bording: Good catch! We should be.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
SeanFeldman: I don't like the idea of `maxlevel` being defined in a tests project. There should be a single source of truth.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
SeanFeldman: It's only going to happen at runtime check time. Inline execution is fine imo.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
SeanFeldman: Would a different name make it better @adamralph ?
For example:
```c#
var messageProperties = message.BasicProperties;
var headers = properties.Headers;
```

src/NServiceBus.RabbitMQ/Routing/ISupportNativeDelays.cs
WojcikMike: I like `ISupportDelayedDelivery` as it explains the intent better. In theory you can implement it however you like.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
adamralph: OK, thanks. "constraint" seems like a weird word to use for that...

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
adamralph: No biggie. Fine to leave as-is.

src/NServiceBus.RabbitMQ/Delays/DelayInfrastructure.cs
adamralph: OK, let's leave as-is for now. It's internal, so not an immediate concern anyway.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: Does the `Headers` property have side effects, or does is just return the value of a field? If the latter (and what the semantics of a property getter convey) then it should be fine to call it more than once.

@SeanFeldman I don't have a problem with the `properties` variable, only `headers` and `messageHeaders`.

If we were to rename variables, then I'd probably rename `headers` to `objectHeaders` and `messageHeaders` to `stringHeaders`, since `object` vs `string` is the only difference between them. Or possibly `originalHeaders` and `deserializedHeaders`.

However, this somewhat of a bike shed so let's not spend too much time on it. Unless the task force feels that one of my suggestions (drop the `headers` variable or one of the two renames) is an obvious improvement, let's just leave as-is.

src/NServiceBus.RabbitMQ/Routing/ISupportNativeDelays.cs
adamralph: 👍 to `ISupportDelayedDelivery`. Perhaps some tweaks to the XML docs may also result from the rename.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
SeanFeldman: Delivery constraints sounds funky, agree. It's a core concept. From downstream it probably make more sense to call it "abilities" or "supported delivery modes". For consistency, should we keep it as-is?

adamralph: I don't think we should change the `constraints` variable name. That could add confusion. As you say, we can't change it since it's a core concept so we'll just have to roll with it. `¯\_(ツ)_/¯`

bording: The reason this is called `DeliveryConstraints` from the core perspective is that we are specifying which delivery contraints we support on a message. The two delayed delivery related constraints are `DoNotDeliverBefore` and `DelayDeliveryWith`. They both have a base class of `DelayedDeliveryConstraint` so I said we support that one instead listing the two separate items.

Perhaps this would feel better if the two actual constraints were shown here instead?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: > Does the Headers property have side effects, or does is just return the value of a field? If the latter (and what the semantics of a property getter convey) then it should be fine to call it more than once.

No, it was just an attempt to reduce method calls here, since it didn't look like the JIT was able to inline the getters when I took a look at the generated assembly from the disassembly window.

So, yeah sort of a micro-optimization, but since this is called on every incoming message, I thought it might be worth doing. I can drop it and simplify the code here if you prefer.

src/NServiceBus.RabbitMQ/Routing/ISupportNativeDelays.cs
bording: @adamralph Yes, making that change will ripple through the code in a number of ways, including changing the name of the setting: `AllEndpointsSupportNativeDelays` should become `AllEndpointsSupportDelayedDelivery`.

That way we banish the term "native" everywhere.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
adamralph: @bording thanks. In that light, it makes sense.

> Perhaps this would feel better if the two actual constraints were shown here instead?

It probably would. Is that achievable in a nice way?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: Unless you have great appetite to change, I'd leave as-is. But, you never know, I might send a refactoring PR a month from now 😉

src/NServiceBus.RabbitMQ/Routing/ISupportNativeDelays.cs
adamralph: 👍 I like it.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
bording: > it probably would. Is that achievable in a nice way?

Yes, just need to add `DoNotDeliverBefore` and `DelayDeliveryWith` to the `constraints` collection instead of `DelayedDeliveryConstraint`.

adamralph: 👍 I say go for it.

bording:
I've pushed up some commits that should address all of the feedback so far.

bording:
@adamralph At this point there is nothing outstanding, so I think it's ready to go. If we do end up finding something, I think we could open a new PR for it.

How do you want to handle merging this? Squash down to a single commit?

adamralph:
IMHO squashing to a single commit is probably best.

bording:
@adamralph squashed!

SeanFeldman:
🎉

PR #331, opened 2017-01-30
Switch Consumer_Received into synchronous blocking mode when semaphore has no more slots

danielmarbach:
Switch Consumer_Received into synchronous blocking mode when semaphore has no more slots
- More natural to the way the client is intended to be used
- Will reduce the semaphore acquiring up to the prefetch count when concurrency limit is smaller than prefetch count

The consume received has two phases:
1. Message flow control which is the semaphore acquire and wait phase
1. Message processing

In the message processing phase we want concurrency and therefore we should just fire & forget the processing task. 

During the message flow control phase I'd argue it is more natural to the way the client is designed to block the worker service thread. The interface is void so there is no way to yield the thread. Even with async void we are not gaining anything and if we can this change would avoid the async void. 

@adamralph has the context of the conversation.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -152,25 +152,30 @@
             }
         }
 
-        async void Consumer_Received(object sender, BasicDeliverEventArgs eventArgs)
+        void Consumer_Received(object sender, BasicDeliverEventArgs eventArgs)
         {
             try
             {
-                await semaphore.WaitAsync(messageProcessing.Token).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


adamralph:
@bording and I discussed briefly but didn't come to a conclusion on this yet. Backlogging for now since we have other priorities but we'll circle back to this in due course.

danielmarbach:
Closing because of https://github.com/Particular/NServiceBus.RabbitMQ/pull/335

With these changes and the synchronous Wait on the semaphore we'd always ending up returning on the worker thread and would also need to yield. Although hair-splitting, we can argue the "only yield when necessary" approach enabled by `async void `and the `WaitAsync `on the semaphore is slightly more efficient.

PR #332, opened 2017-01-30
Yield before processing a message

### LINKED ISSUE #330: Blocking operations in the pipeline lower RabbitMQ performance dramatically.
There is an edge case in the RabbitMQ transport which causes blocking code in the pipeline to block the transports message pump unnecessarily. This can occur in production e.g. when using NHibernate Persistence which does not offer fully async APIs. This can cause an endpoint to slow down to very low message throughput (e.g. a 50ms blocking call per message results in a throughput of ~20 msg/sec). Increasing the concurrency setting has no impact on throughput.

### How to Reproduce

Add any behavior or message handler which does a Thread.Sleep and no asynchronous operation.
e.g.:

    class BlockingBehavior : Behavior<ITransportReceiveContext>
    {
        public BlockingBehavior()
        {
        }

        public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
        {
            Thread.Sleep(50);
            await next();
        }
    }

### Cause

The rabbitMQ client offers a event handler we subscribe to when a message to handle should be dispatched. The transports event handler is `async void` but there is no asynchronous operation happening before the blocking call. This causes the thread from the rabbitmq client which invokes the callback to block until an actual asynchronous operation is hit. Event when message handler use fully async code, this is too late in the pipeline, as the event invoking thread can be blocked by some persistence operations in the pipeline (e.g. the outbox). See the event handler implementation: https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs#L155
(note that `semaphore.WaitAsync` will execute synchronously as long the semaphore limit isn't reached)

### Workaround

a `Task.Run` or `Task.Yield` as early as possible in the pipeline can be added using a custom behavior in the `ITransportReceiveContext` stage. This will free up the rabbitmq client's thread and allow further messages to be processed concurrently. This approach should also include increasing the configured concurrency limit (which works again with the workaround).

### Possible Fixes

a very simple approach could be to call `Task.Yield` as early as possible in the `Consumer_Received` handler to free up the calling thread. @bording raised a valid question whether this should be actually a core concern instead of a transport responsibility. Note that e.g. MSMQ uses Task.Run to invoke the pipeline for messages.

ping @Particular/rabbitmq-transport-maintainers @Particular/nservicebus-maintainers

danielmarbach:
Closes #330

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -165,6 +165,12 @@
 
             try
             {
+                // If the semaphore was acquired successfully on the synchronous path the thread entering
+                // the process method will be the worker service thread. It is crucial to offload the processing
+                // onto the worker thread pool. Theoretically the yield would not be required when the semaphore could not be acquired before. 
+                // The the continuation of WaitAsync is scheduled but we have no way of detecting this.
+                // This is no longer needed when rabbitmq client is fully async.
+                await Task.Yield();
                 await Process(eventArgs).ConfigureAwait(false);
             }
             catch (Exception ex)


danielmarbach:
See also https://github.com/Particular/NServiceBus.RabbitMQ/issues/330#issuecomment-276045786

danielmarbach:
Raised also https://github.com/Particular/NServiceBus.RabbitMQ/pull/333 as food for thought. Would need to measure properly if one has benefit over the other but I leave that exercise up to the @Particular/rabbitmq-transport-maintainers :trollface:

adamralph:
Superseded by #333

PR #333, opened 2017-01-31
Yield processing other

danielmarbach:
Another proposal for #332 

Here we only yield if we completed synchronously. When the semaphore didn't have slots available we are pretty much guaranteed that the continuation is scheduled.


adamralph:
Replaced by #335.

Thanks @danielmarbach - your (squashed) commits are part of #335. 😉

Issue #103, opened 2015-10-06
Propagate all BasicDeliverEventArgs arguments 


erezbartal:
When sending a message from native RabbitMQ API to a NSB endpoint not all BasicDeliverEventArgs are being propogated into the bus.CurrentMessageContext object.

These object could be Exchange, RoutingKey, UserID, etc....

Why do I need these parameters: In an IoT environment where non .NET devices connects directly to the broker there is a need to know which device sends the message. The device code can add a property to the message identifying who it is but this is not secure because a compromised code can impersonate to another device. We need to find a way that the broker infrastructure will identify the client.

One of the options to do it is to use client side certificate or any other rabbit routing mechanism. All these solutions eventually mapped to one or more parameters in the BasicDeliverEventArgs or in BasicDeliverEventArgs.BasicProperties. Unfortunately these parameters are not copied into the TransportMessage object (and eventually into the bus.CurrentMessageContext object) at MessageConverter.ToTransportMessage method

andreasohlund:
:+1:

SimonCropp:
@andreasohlund any objections to simply adding them as headers?

bording:
@andreasohlund Is this something that should be added to the 4.0 milestone?

andreasohlund:
I'd say no, we can add this in a minor

andreasohlund:
One option to solve this would be to add the `B` to the message context passed to the core pipeline

bording:
> One option to solve this would be to add the `BasicDeliverEventArgs` to the message context passed to the core pipeline

> If we go that route we should also consider exposing it on the error context

How are you thinking we could do that?

andreasohlund:
Just add it to the contextbag?

bording:
That's not really a great experience to get it back out, but maybe we could add an extension method or something to help.

However, there is no `ContextBag` on `ErrorContext`.

andreasohlund:
Yea, thinking an extension would be a good way to expose it.

Then lets scope put the ErrorContext and add a core issue to get a
contextbag there in the future?

andreasohlund:
@Particular/nservicebus-maintainers the [error context take a `TransportTransaction` as a parameter](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/ErrorContext.cs#L14) and is therefor extensible.

[...]

PR #334, opened 2017-02-01
expose BasicDeliverEventArgs in message context

SimonCropp:
proposal for https://github.com/Particular/NServiceBus.RabbitMQ/issues/103

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/APIApprovals.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/APIApprovals.cs
@@ -1,8 +1,9 @@
 ﻿using System;
+using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using ApprovalTests;
-using NServiceBus;
 using NUnit.Framework;
 using PublicApiGenerator;
 
@@ -13,7 +14,9 @@ public class APIApprovals
     [MethodImpl(MethodImplOptions.NoInlining)]
     public void Approve()
     {
-        var publicApi = Filter(ApiGenerator.GeneratePublicApi(typeof(RabbitMQTransport).Assembly));
+        var combine = Path.Combine(TestContext.CurrentContext.TestDirectory, "NServiceBus.Transports.RabbitMQ.dll");
+        var assembly = Assembly.LoadFile(combine);
+        var publicApi = Filter(ApiGenerator.GeneratePublicApi(assembly));
         Approvals.Verify(publicApi);
     }
 
--- a/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
@@ -128,5 +128,13 @@
             transportExtensions.GetSettings().Set(SettingsKeys.ClientCertificates, clientCertificates);
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Gets the <see cref="BasicDeliverEventArgs"/> for the current message being processed.
+        /// </summary>
+        public static BasicDeliverEventArgs GetBasicDeliverEventArgs(this IMessageHandlerContext context)
+        {
+            return context.Extensions.Get<BasicDeliverEventArgs>();
+        }
     }
 }
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -214,6 +214,7 @@
                 var errorHandled = false;
                 var numberOfDeliveryAttempts = 0;
 
+                contextBag.Set(message);
                 while (!processed && !errorHandled)
                 {
                     try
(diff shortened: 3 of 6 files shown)


SimonCropp:
@Particular/rabbitmq-transport-maintainers thoughts?

adamralph:
Thanks @SimonCropp, I'll take a look ASAP.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -214,6 +214,7 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
                 var errorHandled = false;
                 var numberOfDeliveryAttempts = 0;
 
+                contextBag.Set(message);
```
bording: @SimonCropp FYI, for this to work the way you're intending, you'll have to also undo the optimization currently in place around the creation of the `ContextBag`. Currently, because we don't use the bag for anything, there is a single static instance that is reused for every message, saving having to allocate one per message.

SimonCropp: > you'll have to also undo the optimization currently in place around the creation of the ContextBag

how does that work when other things place things in the context? eg https://docs.particular.net/nservicebus/sql-persistence/#sqlstoragesession

bording: The ContextBag created by the transport gets merged into the ContextBag managed by Core.

src/NServiceBus.RabbitMQ.Tests/APIApprovals.cs
```diff
@@ -13,7 +14,9 @@ public class APIApprovals
     [MethodImpl(MethodImplOptions.NoInlining)]
     public void Approve()
     {
-        var publicApi = Filter(ApiGenerator.GeneratePublicApi(typeof(RabbitMQTransport).Assembly));
+        var combine = Path.Combine(TestContext.CurrentContext.TestDirectory, "NServiceBus.Transports.RabbitMQ.dll");
+        var assembly = Assembly.LoadFile(combine);
+        var publicApi = Filter(ApiGenerator.GeneratePublicApi(assembly));
```
adamralph: What is the purpose of this change?

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -128,5 +128,13 @@ public static TransportExtensions<RabbitMQTransport> SetClientCertificates(this
             transportExtensions.GetSettings().Set(SettingsKeys.ClientCertificates, clientCertificates);
             return transportExtensions;
         }
+
+        /// <summary>
+        /// Gets the <see cref="BasicDeliverEventArgs"/> for the current message being processed.
+        /// </summary>
+        public static BasicDeliverEventArgs GetBasicDeliverEventArgs(this IMessageHandlerContext context)
```
adamralph: I'd prefer to put this in a `MessageHandlerContextExtensions` class.

SimonCropp: i figured it want worth leaking a new type for one method when they will never see that class anyway

src/NServiceBus.RabbitMQ.Tests/APIApprovals.cs
SimonCropp: because given your test runner typeof(RabbitMQTransport).Assembly might resolve to somewhere where dependency dlls do not exist. it was failing for me

adamralph: 👍 good to know.

src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
adamralph: OK. It's a minor point anyway. We can always change later.

adamralph:
As [mentioned in the issue](https://github.com/Particular/NServiceBus.RabbitMQ/issues/103#issuecomment-281376218), @Particular/rabbitmq-transport-maintainers feel that the best way to do this is to provide a config extension which enables the propagation of `BasicDeliverEventArgs` to the handler. By default, the object will not be propagated and the current optimisation will remain, using the singleton `ContextBag`. When the the config extension is enabled, the optimisation will bypassed, and a `ContextBag` will be created for the scope of the message, containing the `BasicDeliverEventArgs`. The extension method should log a warning when the `BasicDeliverEventArgs` does not exist in the `ContextBag`, indicating that this config switch should be used to enable the propagation.

SimonCropp:
Ok. Sounds like this is not required then

adamralph:
@SimonCropp we could have built on top of your changes. They were a step towards the solution. Sorry for not making that clear.

adamralph:
I stumbled upon the commits in my local clone 😄 - https://github.com/Particular/NServiceBus.RabbitMQ/pull/353

PR #335, opened 2017-02-01
Hotfix 4.2.1

### LINKED ISSUE #330: Blocking operations in the pipeline lower RabbitMQ performance dramatically.

adamralph:
fixes #330

### DIFF
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -154,10 +154,10 @@
 
         async void Consumer_Received(object sender, BasicDeliverEventArgs eventArgs)
         {
-            bool needsYield;
+            var eventRaisingThreadId = Thread.CurrentThread.ManagedThreadId;
+
             try
             {
-                needsYield = semaphore.CurrentCount > 0;
                 await semaphore.WaitAsync(messageProcessing.Token).ConfigureAwait(false);
             }
             catch (OperationCanceledException)
@@ -167,13 +167,27 @@
 
             try
             {
-                // If the semaphore was acquired successfully on the synchronous path the thread entering
-                // the process method will be the worker service thread. It is crucial to offload the processing
-                // onto the worker thread pool. Theoretically the yield would not be required when the semaphore could not be acquired before. 
-                // The the continuation of WaitAsync is scheduled but we have no way of detecting this.
-                // This is no longer needed when rabbitmq client is fully async.
-                if (needsYield)
+                // The current thread will be the event-raising thread if either:
+                //
+                // a) the semaphore was entered synchronously (did not have to wait).
+                // b) the event was raised on a thread pool thread,
+                //    and the semaphore was entered asynchronously (had to wait),
+                //    and the continuation happened to be scheduled back onto the same thread.
+                if (Thread.CurrentThread.ManagedThreadId == eventRaisingThreadId)
                 {
+                    // In RabbitMQ.Client 4.1.0, the event is raised by reusing a single, explicitly created thread,
+                    // so we are in scenario (a) described above.
+                    // We must yield to allow the thread to raise more events while we handle this one,
+                    // otherwise we will never process messages concurrently.
+                    //
+                    // If a future version of RabbitMQ.Client changes its threading model, then either:
+                    //
+                    // 1) we are in scenario (a), but we *may not* need to yield.
+                    //    E.g. the client may raise the event on a new, explicitly created thread each time.
+                    // 2) we cannot tell whether we are in scenario (a) or scenario (b).
+                    //    E.g. the client may raise the event on a thread pool thread.
+                    // 
+                    // In both cases, we cannot tell whether we need to yield or not, so we must yield.
                     await Task.Yield();
                 }
 
--- a/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
@@ -154,8 +154,10 @@
 
         async void Consumer_Received(object sender, BasicDeliverEventArgs eventArgs)
         {
+            bool needsYield;
             try
             {
+                needsYield = semaphore.CurrentCount > 0;
                 await semaphore.WaitAsync(messageProcessing.Token).ConfigureAwait(false);
             }
             catch (OperationCanceledException)
@@ -165,6 +167,16 @@
 
             try
             {
+                // If the semaphore was acquired successfully on the synchronous path the thread entering
+                // the process method will be the worker service thread. It is crucial to offload the processing
+                // onto the worker thread pool. Theoretically the yield would not be required when the semaphore could not be acquired before. 
+                // The the continuation of WaitAsync is scheduled but we have no way of detecting this.
+                // This is no longer needed when rabbitmq client is fully async.
+                if (needsYield)
+                {
+                    await Task.Yield();
+                }
+
                 await Process(eventArgs).ConfigureAwait(false);
             }
             catch (Exception ex)


src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
 
             try
             {
+                // The current thread will be the event raising thread if either:
```
bording: Minor grammatical nitpick: Shouldn't this line either be "the event-raising thread" or "the thread raising the event " ?

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                //    and the semaphore was entered asynchronously (had to wait),
+                //    and the continuation happened to be scheduled back onto the same thread.
+                if (Thread.CurrentThread.ManagedThreadId == eventRaisingThreadId)
+                {
+                    // In RabbitMQ.Client 4.1.0, the event is raised by reusing a single explicitly created thread
+                    // so we are in scenario (a) described above.
+                    // We must yield to allow the thread to raise more events while we handle this one,
+                    // otherwise we will never process messages concurrently.
+                    //
+                    // If a future version of RabbitMQ.Client changes it's threading model then either:
```
bording: "its" not "it's"

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                    //
+                    // 1) we are in scenario (a) but we *may not* need to yield.
+                    //    E.g. the client may raise the event on a new explicitly created thread each time.
+                    // 2) we cannot tell whether we are in scenario (a) or scenario (b).
+                    //    E.g. the client may raise the event on thread pool thread.
+                    // 
+                    // In both these cases we cannot tell whether we need to yield or not, so we must yield.
```
bording: "In both of these"

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
adamralph: 👍 I'll change it to "event-raising" thread so the reader naturally associates it with `eventRaisingThreadId`.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
adamralph: 👍 will change

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
adamralph: I believe [both forms are correct](http://ell.stackexchange.com/questions/110665/both-of-these-or-both-these) but I don't have a strong preference. I have a weak preference for omitting "of".

adamralph:
@bording I made the first two grammar fixes (amending the commit since they were so minor).

Let me know how strongly you feel about "both these" vs "both of these".

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Omitting "of" here sounds rather awkward to me, and I'd never write or say it that way, but your links do seem to support that it's not incorrect!

How about "In both cases" and drop "these" from the sentence?

bording: Also, now that I'm looking at it, does there need to be comma after "cases" ?

adamralph: 👍 I'll make those changes and drop the comma.

bording: I was actually wondering if a comma needed to be added?

adamralph: Oh, my mistake. Sure, I'll add one.

adamralph:
@bording I amended with those changes to the last sentence.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
             {
+                // The current thread will be the event-raising thread if either:
+                //
+                // a) the semaphore was entered synchronously (did not have to wait).
+                // b) the event was raised on a thread pool thread,
+                //    and the semaphore was entered asynchronously (had to wait),
+                //    and the continuation happened to be scheduled back onto the same thread.
+                if (Thread.CurrentThread.ManagedThreadId == eventRaisingThreadId)
+                {
+                    // In RabbitMQ.Client 4.1.0, the event is raised by reusing a single explicitly created thread
```
bording: Hard to notice because the sentence is split onto two lines, but there should be a comma between "thread" and "so".

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                    // so we are in scenario (a) described above.
+                    // We must yield to allow the thread to raise more events while we handle this one,
+                    // otherwise we will never process messages concurrently.
+                    //
+                    // If a future version of RabbitMQ.Client changes its threading model then either:
```
bording: Needs a comma between "model" and "then"

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                    //
+                    // 1) we are in scenario (a) but we *may not* need to yield.
```
bording: Needs a comma between "scenario (a)" and "but"

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
bording: Oh, and I think there should probably be one between "single" and "explicitly" because those are both modifying "thread" and could be applied separately.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                    //    E.g. the client may raise the event on a new explicitly created thread each time.
```
bording: Comma between "new" and "explicitly" for the same reason as above.

src/NServiceBus.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -165,6 +167,30 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
+                    // 2) we cannot tell whether we are in scenario (a) or scenario (b).
+                    //    E.g. the client may raise the event on thread pool thread.
```
bording: Needs an "a" between "on" and "thread"

PR #337, opened 2017-02-04
AMQP field headers and custom primitive value headers are NULL

bording:
While working on #329, I hit a point where I needed to improve how incoming header values were being handled. It didn't make sense to do that in the same branch, since it's not specific to native delays, so here is that work.

`ValueToString` should now be able to handle any possible value that could be in the `BasicProperties.Headers` collection, according to the AMQP spec and the RabbitMQ client's [implementation](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs#L111).

CC: @SeanFeldman

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Transport.RabbitMQ.Tests
 {
     using System;
-    using System.Collections;
     using System.Collections.Generic;
     using System.Text;
     using global::RabbitMQ.Client.Events;
@@ -212,27 +211,6 @@
 
             Assert.NotNull(headers);
             Assert.AreEqual("ni", headers["Foo"]);
-        }
-
-        [Test]
-        public void TestCanHandleStringArrayListsHeader()
-        {
-            var message = new BasicDeliverEventArgs
-            {
-                BasicProperties = new BasicProperties
-                {
-                    MessageId = "Blah",
-                    Headers = new Dictionary<string, object>
-                {
-                    {"Foo", new ArrayList{"Bing"}}
-                }
-                }
-            };
-
-            var headers = converter.RetrieveHeaders(message);
-
-            Assert.NotNull(headers);
-            Assert.AreEqual("Bing", headers["Foo"]);
         }
 
         [Test]
--- a/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
@@ -138,6 +138,62 @@
         }
 
         [Test]
+        public void TestCanHandleHeadersWithAllAmqpFieldValues()
+        {
+            var message = new BasicDeliverEventArgs
+            {
+                BasicProperties = new BasicProperties
+                {
+                    MessageId = "Blah",
+                    Headers = new Dictionary<string, object>
+                    {
+                        {"short", (short)42},
+                        {"int", 42},
+                        {"long", 42L},
+                        {"decimal", 42m},
+                        {"sbyte", (sbyte)42},
+                        {"double", 42d},
+                        {"single", 42f},
+                        {"bool", true }
+                    }
+                }
+            };
+
+            var headers = converter.RetrieveHeaders(message);
+
+            Assert.NotNull(headers);
+            Assert.AreEqual("42", headers["short"]);
+            Assert.AreEqual("42", headers["int"]);
+            Assert.AreEqual("42", headers["long"]);
+            Assert.AreEqual("42", headers["decimal"]);
+            Assert.AreEqual("42", headers["sbyte"]);
+            Assert.AreEqual("42", headers["double"]);
+            Assert.AreEqual("42", headers["single"]);
+            Assert.AreEqual("True", headers["bool"]);
+        }
+
+        [Test]
+        public void TestCanHandleAmqpTimestampHeader()
+        {
+            var message = new BasicDeliverEventArgs
+            {
+                BasicProperties = new BasicProperties
+                {
+                    MessageId = "Blah",
+                    Headers = new Dictionary<string, object>
+                    {
+                        {"Foo", new global::RabbitMQ.Client.AmqpTimestamp(int.MaxValue) }
+                    }
+                }
+            };
+
+            var headers = converter.RetrieveHeaders(message);
+
+            Assert.NotNull(headers);
+            Assert.AreEqual("2038-01-19 03:14:07:000000 Z", headers["Foo"]);
+        }
+
+        [Test]
         public void TestCanHandleStringHeader()
         {
             var message = new BasicDeliverEventArgs
--- a/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Transport.RabbitMQ
 {
     using System;
-    using System.Collections;
     using System.Collections.Generic;
     using System.Text;
     using global::RabbitMQ.Client.Events;
@@ -89,7 +88,7 @@
 
                 foreach (var header in messageHeaders)
                 {
-                    deserializedHeaders.Add(header.Key, header.Value == null ? null : ValueToString(header.Value));
+                    deserializedHeaders.Add(header.Key, ValueToString(header.Value));
                 }
             }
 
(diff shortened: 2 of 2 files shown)


src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -98,50 +97,65 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
...
-            }
-
             var bytes = value as byte[];
             if (bytes != null)
             {
                 return Encoding.UTF8.GetString(bytes);
             }
 
-            var dictionary = value as IDictionary<string, object>;
+            var dictionary = value as Dictionary<string, object>;
```
SeanFeldman: Why casting to a `Dictionary<>` rather than `IDictionary<>`?

bording: Because in that case the underlying type is actually a `Dictionary`. Casting to that instead means we don't end up with a heap allocation for enumerating it via the foreach.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -98,50 +97,65 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
 
         static string ValueToString(object value)
         {
-            var s = value as string;
-            if (s != null)
-            {
-                return s;
-            }
-
```
adamralph: Why has this been removed? Naively, it seems to make sense, and if I add it back in, the tests still pass.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -98,50 +97,65 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
...
                 }
 
-                return string.Join(",", valuesToJoin);
+                if (sb.Length > 0)
+                {
+                    sb.Remove(sb.Length - 1, 1);
+                }
+
+                return sb.ToString();
             }
```
adamralph: While I prefer this version of handling the `Dictionary<string, object>` case, it doesn't seem to provide any functional change. If I revert to the old method, the tests still pass.

Is this just a performance and/or readability optimisation? If so, I'd have preferred to see it in a separate refactoring commit, but I'm OK to leave it as-is in this case.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -98,50 +97,65 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
...
 
-            var list = value as IList;
+            var list = value as List<object>;
```
adamralph: Is this making an assumption about RabbitMQ.Client internals?

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -98,50 +97,65 @@ string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
...
                 }
 
-                return string.Join(";", valuesToJoin);
+                if (sb.Length > 0)
+                {
+                    sb.Remove(sb.Length - 1, 1);
+                }
+
+                return sb.ToString();
+            }
```
adamralph: Same comment here as for the `Dictionary<string, object>` case.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: @adamralph No, this change specifically, and the `ValueToString` method in general has nothing to do with NSB core.

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
```diff
@@ -147,36 +146,71 @@ public void TestCanHandleStringHeader()
...
-            Assert.AreEqual("ni", headers["Foo"]);
+            Assert.AreEqual("42", headers["short"]);
+            Assert.AreEqual("42", headers["int"]);
+            Assert.AreEqual("42", headers["long"]);
+            Assert.AreEqual("42", headers["decimal"]);
+            Assert.AreEqual("42", headers["sbyte"]);
+            Assert.AreEqual("42", headers["double"]);
+            Assert.AreEqual("42", headers["single"]);
+            Assert.AreEqual("True", headers["bool"]);
         }
```
bording: The reason I wrote the test the way I did is that all of those values exercise the exact same code path in `ValueToString` : the `value?.ToString()` call at the end. I didn't really see a lot of value in having a bunch of different tests.

The test is actually making sure we can handle all of the possible AMQP field types of a field table, so perhaps `TestCanHandleHeadersWithAllAmqpFieldValues.`

>The short and long cases are not properly typed, they should be (short)42 and 42L.

Good call on that, missed those.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: Because based on how the AMQP field tables are defined in the spec, and how the RabbitMQ client has implemented its parse of them, there are a specific set of values we can encounter here, and `string` will never be one of them.

Based on that, I see no point in keeping it around.

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
```diff
@@ -147,36 +146,71 @@ public void TestCanHandleStringHeader()
...
 
         [Test]
-        public void TestCanHandleStringArrayListsHeader()
```
bording: Same reply as to why I removed the `string` path from the method. We will never get an ArrayList passed to us from the Rabbit client.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: When making these changes, I wasn't really thinking in terms like that. I pretty much gutted it and started over. The way it was written before seemed inefficient, so I didn't rewrite it that way.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: To some degree, yes, but given the nature of what `ValueToString` is doing I think that is both OK and somewhat required.

We know we're being passed a C# representation of an AMQP field table, so that means there is exactly defined set of possible values here. The way the client represents a field array is with a `List<object>`.

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
adamralph: > I didn't really see a lot of value in having a bunch of different tests.

While all the cases are passing there's no advantage. When one of the cases fails, the benefit is seen, since the test runner will tell you exactly which case failed. However, given that all of these pass through exactly the same code path, that will never happen, so I'm happy to leave this method as-is. There's probably a small performance benefit too, but perhaps not significant.

> TestCanHandleHeadersWithAllAmqpFieldValues

👍 I like that. With that rename, it makes sense to keep the `decimal` case inside the test.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: Same reply as the other case as well. Yes it is a more efficient way of handling the case of when you have a field array passed in.

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
adamralph: 👍

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: 👍 make sense.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: Sorry, I was confused. I guess it's an assumption about RabbitMQ.Client internals then. Overall, I think that's OK (see comment about that below.)

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: OK, no problem. I'm fine with leaving this as-is.

For reference, these changes alone make the tests green:

image

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: It's not ideal to have make these assumptions, but I think it's OK. We smoke test every RabbitMQ.Client release anyway, so if the internal changes then we'll know about it pretty quick.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
adamralph: 👍

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
```diff
@@ -138,7 +137,7 @@ public void Should_override_replyto_header_if_native_replyto_is_present()
         }
 
         [Test]
-        public void TestCanHandleStringHeader()
+        public void TestCanHandleHeadersWithPrimitiveTypes()
```
SeanFeldman: I kep silent about it as I'm not a maintainer 😃

bording:
> One more question: is the end result of this change observable by the user? I.e. are we adding a new feature here, or fixing a bug?

@adamralph Missed this question. The change would be visible by an end user if they were expecting to see certain RabbitMQ native headers propagated into the NSB headers collection with a string representation of their value. For example, values inside of the `x-death` header. In native integration scenarios where they might be adding a custom header with a non-string value, this would also be a problem. I hit this with the new Delay header, which is being added as an int. I noticed that the value was lost and was null instead.

Previously, the method assumed that everything coming into it was either some bytes to decode into a string, or a nested collection of byte arrays that should be treated as strings. If any other type came in, you'd end up with a null value instead of the expected value.

Now, it can properly handle any of the actual data types that might be in there, including the unix epoch timestamp fields.

So, based on that, I think that classifying it as a bug makes the most sense?

src/NServiceBus.RabbitMQ.Tests/MessageConverterTests.cs
bording: Yeah, those tests names are not good, but I figured it was better to stick with the current convention for now.

src/NServiceBus.RabbitMQ/Receiving/MessageConverter.cs
bording: Cool, yeah that does seem like the minimum change that would make everything work, though from my actual workflow standpoint, I added the timestamp stuff last.

BTW, one of the additional reasons the commits ended up the way they did, and not more individual work commits, is that I originally started this work on the native delays branch and had some stuff specific to in mixed in. I also added tests at different stages as well. There was a decent amount of interactive rebasing involved to get everything over on develop cleanly!

adamralph:
> So, based on that, I think that classifying it as a bug makes the most sense?

Agreed. I'll go ahead and tentatively rename the PR. Feel free to tweak.

PR #338, opened 2017-02-06
dispose of channels in ChannelProvider

adamralph:
Thanks to @SimonCropp for pointing this out.

@Particular/rabbitmq-transport-maintainers I guess we should consider this as tech-debt, unless we can demonstrate a functional change.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
@@ -53,6 +53,11 @@ namespace NServiceBus.Transport.RabbitMQ
             {
                 connection.Value.Dispose();
             }
+
+            foreach (var channel in channels)
+            {
+                channel.Dispose();
+            }
         }
 
         readonly Lazy<IConnection> connection;


bording:
This isn't actually required. Calling `Dispose` on the connection is all that is needed here to close everything down.

adamralph:
@bording fair enough, but let's bake that into the code with a comment to prevent anyone raising this again. On the other hand, that's an assumption on the implementation and requires internal knowledge of the client (which I guess may change). Do you think it's worth optimising out this code, or better to leave it in place?

bording:
>On the other hand, that's an assumption on the implementation and requires internal knowledge of the client (which I guess may change)

True, though it's also somewhat constrained by how AMQP works, so I'm not sure there's a ton of room for change there.

For example, you call Dispose on a channel, it internally calls an `Abort` method, which is a special way of sending a channel close message, in that it catches exceptions and prevents them from bubbling back up, for example if the connection has already been closed.

That's why if you've closed the connection, there's not really a reason to close each channel separately.

I suppose you could try to dispose the channels first, but then from a correctness standpoint, there would be nothing preventing new channels from being created by a `GetPublishChannel` call, so you'd still potentially be in the same state.

Hmm, so thinking about it some more, I don't suppose it really hurts to make the change you've made here, even though it doesn't actually do anything.

adamralph:
@Particular/rabbitmq-transport-maintainers in the last sync we agreed to make this change. Can someone please approve and merge?

PR #339, opened 2017-02-06
refactor: BasicPropertiesExtensions - in-lined some stuff

adamralph:
Not sure if this is to everyone's taste. I know I in-line too much sometimes.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
+++ b/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
@@ -19,7 +19,7 @@
             }
 
             DiscardIfNotReceivedBefore timeToBeReceived;
-            if (TryGet(deliveryConstraints, out timeToBeReceived) && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
+            if (deliveryConstraints.TryGet(out timeToBeReceived) && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
             {
                 properties.Expiration = timeToBeReceived.MaxTime.TotalMilliseconds.ToString(CultureInfo.InvariantCulture);
             }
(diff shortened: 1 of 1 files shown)


adamralph:
@Particular/rabbitmq-transport-maintainers as discussed, I amended to restrict it to method internals only (plus the conversion to extension method).

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -79,24 +79,13 @@ public static void SetConfirmationId(this IBasicProperties properties, ulong con
...
-            {
-                var headerBytes = properties.Headers[confirmationIdHeader] as byte[];
-                var headerString = Encoding.UTF8.GetString(headerBytes ?? new byte[0]);
-
-                return UInt64.TryParse(headerString, out confirmationId);
-            }
-
-            return false;
+            object value;
+            return properties.Headers.TryGetValue(confirmationIdHeader, out value) &&
```
bording: Can you add a line break before the return statement?

bording: Also, I know this has come up before too, but I'd personally prefer to see this on one line instead of wrapped to two, but that's just my preference.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -79,24 +79,13 @@ public static void SetConfirmationId(this IBasicProperties properties, ulong con
...
         }
 
-        static bool TryGet<T>(List<DeliveryConstraint> list, out T constraint) where T : DeliveryConstraint
-        {
-            constraint = list.OfType<T>().FirstOrDefault();
-
-            return constraint != null;
-        }
+        static bool TryGet<T>(this List<DeliveryConstraint> list, out T constraint) where T : DeliveryConstraint
+            => (constraint = list.OfType<T>().FirstOrDefault()) != null;
```
bording: This is also something I'd probably just write as a single line.

bording:
While looking at this, I noticed that another refactoring/optimizing opportunity here would be to go ahead and get rid of the LINQ usage.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
adamralph: Sure, I'll add the line break. I'd prefer to keep it on two lines if that's OK. One line would be 179 chars long.

src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs
adamralph: Again, I'd prefer to keep it on two lines if that's OK. One line would be 174 chars long.

However, in revisiting this line I noticed that the expression operator `=>` should really be on the first line. That seems to be the more common convention.

adamralph:
> While looking at this, I noticed that another refactoring/optimizing opportunity here would be to go ahead and get rid of the LINQ usage.

True. Perhaps in a subsequent PR?

PR #352, opened 2017-03-02
Add warning when subscribing using an interface

### LINKED ISSUE #269: Log a warning when using an interface for routing in DirectRoutingTopology
*From https://github.com/Particular/NServiceBus.RabbitMQ/issues/11#issuecomment-260711710*

Should probably be released only after #268 is closed.

WojcikMike:
fixes #269

### DIFF
--- a/src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
+++ b/src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
@@ -4,6 +4,7 @@
     using System.Collections;
     using System.Collections.Concurrent;
     using System.Linq;
+    using Logging;
 
     class DefaultRoutingKeyConvention
     {
@@ -23,6 +24,11 @@
 
             var implementedInterface = interfaces.FirstOrDefault();
 
+            if (interfaces.Count > 1)
+            {
+                Logger.WarnFormat("The direct routing topology cannot properly publish a message type that implements more than one relevant interface. The type '{0}' implements the following interfaces: {1}. The interface that will be used is '{2}'. The others will be ignored, and any endpoints that subscribe to those interfaces will not receive a copy of the message.", type, string.Join(", ", interfaces), implementedInterface);
+            }
+            
             if (implementedInterface != null)
             {
                 key = GetRoutingKey(implementedInterface, key);
@@ -59,5 +65,6 @@
 
         static readonly byte[] MsPublicKeyToken = typeof(string).Assembly.GetName().GetPublicKeyToken();
         static readonly ConcurrentDictionary<Type, bool> IsSystemTypeCache = new ConcurrentDictionary<Type, bool>();
+        static readonly ILog Logger = LogManager.GetLogger(typeof(DefaultRoutingKeyConvention));
     }
 }
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
@@ -329,6 +329,7 @@
     <Compile Include="TaskEx.cs" />
     <Compile Include="When_receiving_a_reply_that_contains_a_legacy_callback_header.cs" />
     <Compile Include="When_the_message_contains_a_legacy_callback_header.cs" />
+    <Compile Include="When_publishing_message_implementing_interface_in_direct_topology.cs" />
     <Compile Include="When_using_direct_routing.cs" />
     <Compile Include="When_the_broker_connection_is_lost.cs" />
     <Compile Include="When_using_a_custom_message_id_strategy.cs" />
--- a/src/NServiceBus.RabbitMQ.AcceptanceTests/When_publishing_message_implementing_interface_in_direct_topology.cs
+++ b/src/NServiceBus.RabbitMQ.AcceptanceTests/When_publishing_message_implementing_interface_in_direct_topology.cs
@@ -1,0 +1,90 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.AcceptanceTests
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Features;
+    using Logging;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_publishing_message_implementing_interface_in_direct_topology : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_log_a_warning()
+        {
+            var context = await Scenario.Define<MyContext>()
+                .WithEndpoint<Publisher>(b =>
+                        b.When(c => c.Subscriber1Subscribed, session => session.Publish(new MyRequest())))
+                .WithEndpoint<Receiver>(b => b.When(async (session, c) =>
+                {
+                    await session.Subscribe<IMyRequest>();
+
+                    if (c.HasNativePubSubSupport)
+                    {
+                        c.Subscriber1Subscribed = true;
+                    }
+                }))
+                .Done(c => c.GotTheMessage)
+                .Run();
+
+            Assert.True(context.Logs.Any(l => l.Level == LogLevel.Warn && l.Message.Contains("The direct routing topology cannot properly publish a message type that implements")));
+        }
+
+        public class Publisher : EndpointConfigurationBuilder
+        {
+            public Publisher()
+            {
+                EndpointSetup<DefaultServer>(c => c.UseTransport<RabbitMQTransport>()
+                    .UseDirectRoutingTopology());
+            }
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(builder => {
+                    builder.UseTransport<RabbitMQTransport>().UseDirectRoutingTopology();
+                    builder.DisableFeature<AutoSubscribe>();
+                }, metadata => metadata.RegisterPublisherFor<IMyRequest>(typeof(Publisher)));
+            }
+
+            class MyEventHandler : IHandleMessages<IMyRequest>
+            {
+                readonly MyContext myContext;
+
+                public MyEventHandler(MyContext myContext)
+                {
+                    this.myContext = myContext;
+                }
+
+                public Task Handle(IMyRequest message, IMessageHandlerContext context)
+                {
+                    myContext.GotTheMessage = true;
+
+                    return TaskEx.CompletedTask;
+                }
+            }
+        }
+
+        public class MyRequest : IMyRequest, IOtherRequest
+        {
+        }
+
+        public interface IMyRequest : IEvent
+        {
+        }
+
+        public interface IOtherRequest: IEvent
+        {
+        }
+
+        class MyContext : ScenarioContext
+        {
+            public bool GotTheMessage { get; set; }
+            public bool Subscriber1Subscribed { get; set; }
+        }
+    }
+}


src/NServiceBus.RabbitMQ/Routing/DirectRoutingTopology.cs
```diff
@@ -18,6 +19,12 @@ public DirectRoutingTopology(Conventions conventions, bool useDurableExchanges)
 
         public void SetupSubscription(IModel channel, Type type, string subscriberName)
         {
+            if (type.IsInterface)
```
bording: Instead of adding this check here, I think it should go into `DefaultRoutingKeyConvention.GetRoutingKey` instead, where we can use the `interfaces` collection to log a warning only if there are problematic interfaces found.

This also lets us simplify the warning message by not needing to mention non-system messages.

WojcikMike:
@bording I moved that check to `DefaultRoutingKeyConvention`. I like it a bit more. Yet at the moment we are logging warning when we know that situation is invalid (multiple interfaces). So if everything is ok (only 1 interface) we would not log any warning. Few ideas:

 * Maybe we should throw exception when we know that situation is invalid (multiple interfaces). As at the moment the behaviour is wierd and should be avoided. The chances that someone will miss the warning log entry is high. 
 * Maybe adding more details to warning message specyfing which interface is chosen using FirstOrDefault. That would make troubleshooting easier.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
```diff
@@ -21,6 +22,12 @@ static string GetRoutingKey(Type type, string key = "")
             var interfaces = type.GetInterfaces()
                 .Where(i => !IsSystemType(i) && !IsNServiceBusMarkerInterface(i)).ToList();
 
+            if (interfaces.Count > 1)
+            {
+                Logger.WarnFormat(@"The direct routing topology may not deliver message types with interfaces in their inheritance hierarchy. Refrain from using interface
```
adamralph: I wonder if this message should borrow more from the language used in the warning note under https://docs.particular.net/nservicebus/rabbitmq/routing-topology#direct-routing-topology-enabling-direct-routing-topology

src/NServiceBus.RabbitMQ/Routing/DirectRoutingTopology.cs
```diff
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using global::RabbitMQ.Client;
+    using Logging;
```
adamralph: This should have been removed by the second commit (which should be squashed before merging).

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: @adamralph I do agree that there is some room to tweak the language of the message, but I'm not sure what you've got in mind regarding that warning.

bording:
> Maybe we should throw exception when we know that situation is invalid (multiple interfaces). As at the moment the behaviour is wierd and should be avoided. The chances that someone will miss the warning log entry is high.

I kind of like the idea of just throwing an exception instead of logging, but would this be an acceptable behavior change? I suppose if we're silently letting a problem go through right now, an exception would be an OK way to fix this?

> Maybe adding more details to warning message specyfing which interface is chosen using FirstOrDefault. That would make troubleshooting easier.

👍

WojcikMike:
> I kind of like the idea of just throwing an exception instead of logging, but would this be an acceptable behavior change? I suppose if we're silently letting a problem go through right now, an exception would be an OK way to fix this?

In my opinion. We are stating that behaviour is nondeterministic, and not recomended. As such I feel uncomfortable with allowing it to be used. If the behaviour would be odd but well documented and someone decidedes to use it understanding the rules it would be a different story. But in here, any newly added interface in the event hierarchy can change this behaviour. I would vote for throwing an exception. Either someone will change to other routing topology, remove usage of the interfaces, or keep it limited to only 1 interface in the whole chain. 

@adamralph @bording what do you think?

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: After re-reading it I think it's OK as-is.

adamralph: However, we should not be using a multi-line string literal just because the message is too wide to fit in a code window.

@WojcikMike can you please remove the multi-line string literal. It should be a single line string literal, otherwise the exception message will also contain the line break.

adamralph:
Throwing an exception is a breaking change. I'm fine with doing it, but not now. The first step is the logging  of an exception and that is what is described by #269. We can raise a separate issue for throwing an exception and mark it as breaking.

adamralph:
> Maybe adding more details to warning message specyfing which interface is chosen using FirstOrDefault. That would make troubleshooting easier.

👍 I like this proposal.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: Given that we're only logging this when we find more than one interface, it feels like maybe the message should be more specific here, pointing out that we're logging it because there is more than one?

On the other hand, if we want to prevent people from using any interfaces at all, maybe the `if` should be changed to > 0? It feel like the current error message is assuming that it would be logged if any are found.

bording: Also, "Refrain from using interface for routing purposes.", has a typo. It should be "interfaces". 

However, I'm wondering if this should be changed, to something more specific about not using interfaces for message that are published, because sending doesn't care.

bording: Oh, and one more though: I don't care for the word "may" in the first sentence. I think "might" would be a better choice.

src/NServiceBus.RabbitMQ/Routing/DirectRoutingTopology.cs
WojcikMike: Done.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
WojcikMike: @adamralph multi-line string removed.

WojcikMike: @bording most of your wording comments were addressed.
Regarding

> On the other hand, if we want to prevent people from using any interfaces at all, maybe the if should be changed to > 0? It feel like the current error message is assuming that it would be logged if any are found.

I did that change to actualy implement the feature #269. Which says about logging when an interface is used. Which means > 0.

WojcikMike: > However, I'm wondering if this should be changed, to something more specific about not using interfaces for message that are published, because sending doesn't care.

Maybe something like this:
When publishing, the direct routing topology might not deliver message types with interfaces in their inheritance hierarchy.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
```diff
@@ -23,6 +24,11 @@ static string GetRoutingKey(Type type, string key = "")
 
             var implementedInterface = interfaces.FirstOrDefault();
 
+            if (interfaces.Count > 0)
+            {
+                Logger.WarnFormat("The direct routing topology might not deliver message types with interfaces in their inheritance hierarchy. Refrain from using interfaces for routing purposes. Interfaces found are {0}, the one that will be used for building routing key is {1}", string.Join(", ", interfaces), implementedInterface);
+            }
+
```
adamralph: As discussed today, I like the idea of two messages, one to say that you shouldn't use messages, and one to say that you're using more than one and this the one you're using.

bording: Also, a change to say something like "Message types used for publishing should not implement interfaces" instead of the "refrain" sentence.

WojcikMike: Done, please review

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
```diff
@@ -23,6 +24,16 @@ static string GetRoutingKey(Type type, string key = "")
 
             var implementedInterface = interfaces.FirstOrDefault();
 
+            if (interfaces.Count > 0)
+            {
+                Logger.Warn("The direct routing topology might not deliver message types with interfaces in their inheritance hierarchy. Message types used for publishing should not implement interfaces.");
```
adamralph: My only slight gripe with this, is that most of our samples _do_ publish types which implement an interface: `IEvent`.

Can we say something like "non-system interfaces"? Or "custom interfaces"?

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
```diff
@@ -23,6 +24,16 @@ static string GetRoutingKey(Type type, string key = "")
...
+
+                if (interfaces.Count > 1)
+                {
+                    Logger.ErrorFormat("Publishing type has multiple interfaces. Interfaces are {0}, the one that will be used for building routing key is {1}", string.Join(", ", interfaces), implementedInterface);
```
adamralph: grammaros: "*The* ~P~*p*ublishing type... *The* ~I~*i*nterfaces... ...used for *the* building routing key"

adamralph: I would split the second sentence:

`The interfaces are {0}. The one that will be used...`

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: @adamralph This warning won't be logged if someone is using `IEvent`, so is that relevant? The fact that it's difficult to succinctly specify which interfaces are ignored is why I'm not sure it's a good idea to include that language here.

bording:
Looking at both of the log messages, I think they both need to specify the actual type being processed that is triggering the problem.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: @bording OK, I guess it's better that I try and think of a situation where a user might read that and say "WTF?".

How about:

> I have a few event types, and they all implement `IEvent`, as shown in the samples. Everything works nicely and I deploy my system. When building a new feature, I decide to sprinkle a few of my event types with some new interfaces since it seems to the best way to change things. I smoke test my app and I get a log message telling me that my messages shouldn't implement events. WTF? They've always implemented `IEvent`, so why am I seeing this message for the first time? So you're telling me I'm not supposed to implement interfaces, but your samples all implement from `IEvent`. WTF?

adamralph:
> Looking at both of the log messages, I think they both need to specify the actual type being processed that is triggering the problem.

👍

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: @adamralph I suppose that scenario could happen, but can you come up with a way to accurately and succinctly explain which interfaces are the "bad" ones here?  Both "non-system" and "custom" don't really mean anything without an explanation of what we mean by those terms.

bording: If you look at [DefaultRoutingKeyConvention](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs#L22), the message would need to explain that it's any interface defined in an assembly that has the same public key token as the assembly that contains `string` (which is an assumption that all MS public keys will be the same), and also not one of our marker interfaces.

adamralph: 😆 let's just shoehorn that entire thing into the exception message! :trollface: 

Hmm... this is a real beard-scratcher.

OK, off the top of my head, how about a different approach:

> The direct routing topology might not deliver message types with these interfaces in their inheritance hierarchy: {list of interfaces}. Message types used for publishing should not implement these interfaces.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: Oh, one more thing: Why is this messaged logged at error level? Shouldn't it be at warning level?

bording: @adamralph I think the idea here is that if we detect any interfaces at all, we log the warning, but when we find more than one, it should be logged as an error, because that is definitely going to break. If an endpoint subscribes to one of the interfaces that isn't chosen here, it will not receive any messages.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: Maybe we need to take a step back and reevaluate what we're trying to accomplish with this warning instead.

If we've detected that there is just a single interface, then we don't actually have a problem, because it should work when there is just one.

The problem is actually when we have more than one, which is what the Error log entry is pointing out.

Maybe we just need to remove this `Count > 0` warning, and only log in the `Count > 1` error case?

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
bording: Based on my comment above about only having this condition log anything, I think the message should be expanded to specifically point out that because we're only using a single interface, the other interfaces will not participate in the publishing process at all. How about:

>The direct routing topology cannot properly publish a message type that implements more than one relevant interface. The type '{0}' implements the following interfaces: {1}. The interface that will be used is '{2}'. The others will be ignored, and any endpoints that subscribe to those interfaces will not receive a copy of the message.

bording: Of course even that text has some caveats. The endpoint might still get a copy of the message if it also subscribes to the interface that was chosen, or one of the concrete types in the message's inheritance hierarchy.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: I think I prefer that approach. 👍 to removing this case.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
adamralph: ❤️

adamralph: Regarding the log level, I still think it should be a warning. Someone could have this all running right now with no problems. They upgrade, and all of a sudden error level logs are being spewed everywhere, but no error exists. I think warning level is correct, since it indicates a _possible_ error.

bording: @adamralph I suppose I can get on board with that reasoning...

It still feels like we'd want to put as much attention to this scenario as possible though, since you could consider it a message-loss scenario.

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
WojcikMike: If we are going with this approach of count > 1. We should modify the corresponding task. As currently it states: 

> Log a warning when using **an interface** for routing

src/NServiceBus.RabbitMQ/Routing/DefaultRoutingKeyConvention.cs
WojcikMike: Let me change it to warning level, and use Brandon's message format from here:
https://github.com/Particular/NServiceBus.RabbitMQ/pull/352#discussion_r105018249

WojcikMike:
@bording @adamralph I introduced the changes. To give you an example of warning log. The one from tests looks like this:

> The direct routing topology cannot properly publish a message type that implements more than one relevant interface. The type 'NServiceBus.Transport.RabbitMQ.AcceptanceTests.When_publishing_message_implementing_interface_in_direct_topology+MyRequest' implements the following interfaces: NServiceBus.Transport.RabbitMQ.AcceptanceTests.When_publishing_message_implementing_interface_in_direct_topology+IMyRequest, NServiceBus.Transport.RabbitMQ.AcceptanceTests.When_publishing_message_implementing_interface_in_direct_topology+IOtherRequest. The interface that will be used is 'NServiceBus.Transport.RabbitMQ.AcceptanceTests.When_publishing_message_implementing_interface_in_direct_topology+IMyRequest'. The others will be ignored, and any endpoints that subscribe to those interfaces will not receive a copy of the message.

adamralph:
> superseded

That was for the _previous_ review where I requested changes. This UI clearly still needs some work.

bording:
As @adamralph said, this is ready to be squashed and then we can merge it.

PR #354, opened 2017-03-15
[WIP] upgrade to RabbitMQ.Client 5.0.0

adamralph:


WojcikMike:
I checked it and those 3 tests fails on my box as well. This test is a bit worrying: When_subscribed_to_a_event.Should_receive_the_event_if_subscribed_to_the_base_class

As it seems that creating bindings 2 times between the same types throw exception (in our case it's from EventBase to System.Object. As this hierarchy is present for 2 events used in the test.

bording:
@WojcikMike What exception is being thrown?

WojcikMike:
@bording 
https://builds.particular.net/viewLog.html?buildId=210576&tab=buildResultsDiv&buildTypeId=NServiceBusRabbitMQ_1Buil#testNameId-684505635844872541
Key exist in dictionary

bording:
Ah, I thought you were referring to some underlying exception coming from the client now that would cause the problem. That key exception is just a symptom of something else.

bording:
I'm still making sure all three failures are from the same cause, but here's what I've found so far: rabbitmq/rabbitmq-dotnet-client#314

bording:
Yeah, looks like that is the underlying reason for all three test failures. The two acceptance tests are triggering it because of the `ExchangeBind` call [here](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs#L86) getting called more than once with the same arguments.

adamralph:
@Particular/rabbitmq-transport-maintainers I changed the version to `5.0.1-rc*` so that all we need to do is run the build to smoke test a new RC (although perhaps sometimes a rebase will be required first).

bording:
@adamralph I see that a review has been requested on this. I'm assuming that was done automatically because of the code owners file?

adamralph:
@bording yes, I believe so. I assume it was when I force pushed the amended commit.

I guess we'll see how it goes with that file in place. If it becomes too noisy, we can always remove it.

adamralph:
Work will carry on in #404

Issue #160, opened 2016-06-14
Add environment set up for running tests to CONTRIBUTING.md


adamralph:
~This involves incorporating both [Docker.md](https://github.com/Particular/NServiceBus.RabbitMQ/blob/43ca907eec23451a0da0a239ee6579ad0842caae/Docker.md) and at least some of the 
[content](https://github.com/Particular/NServiceBus.RabbitMQ/files/313969/GETTING_STARTED.md.txt) from https://github.com/Particular/NServiceBus.RabbitMQ/pull/144 into this file,~ We should include info about creating an environment for running the tests, e.g. RabbitMQ locally or on another machine (including VM's or cloud) so that we have a one stop shop for how to contribute to the repo.

distantcam:
We should drop using our own `Dockerfile` since it doesn't have anything in it anyway.

ramonsmits:
Too old

PR #361, opened 2017-03-30
Remove outdated docker information

bording:
These files are out of date, so no reason to keep them around.


WojcikMike:
@bording @adamralph does it mean that we can close: https://github.com/Particular/NServiceBus.RabbitMQ/issues/160

adamralph:
@WojcikMike I don't think so, since our CONTRIBUTING.md still contains no info at all about setting up an environment. The description of #160 is indeed outdated though, so I'll see if I can give it some ❤️

PR #366, opened 2017-04-19
Use C# 7 features to clean up code

bording:
Now that the build servers have been updated, I went through the code and found places where we could take advantage of C# 7. 

The primary area of clean up was being able take advantage of out variables (converting to `var` where possible). 

There were also a few places where is expression pattern matching could be used.


bording:
Hmm, looks like the VS 2017 support has to be enabled manually in TC first...

build.csx
```diff
@@ -16,7 +16,7 @@ var resharperCltUrl = new Uri("https://download.jetbrains.com/resharper/JetBrain
 var resharperCltPath = $"{Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)}/.resharper/{resharperCltUrl.Segments.Last()}";
 var inspectCodePath = $"./.resharper/{Path.GetFileNameWithoutExtension(resharperCltUrl.Segments.Last())}/inspectcode.exe";
 var nuget = "src/.nuget/v4.0.0/NuGet.exe";
-var msBuild = $"{Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}/MSBuild/14.0/Bin/msbuild.exe";
```
bording: MSBuild 15 has moved inside the Visual Studio installation, which means we can't guarantee this path is valid any more.

If we don't want to assume Professional, we need to pull in https://github.com/Microsoft/vswhere to actually figure it out at run time.

build.csx
```diff
@@ -16,7 +16,7 @@ var resharperCltUrl = new Uri("https://download.jetbrains.com/resharper/JetBrain
...
+var msBuild = $"{Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}/Microsoft Visual Studio/2017/Professional/MSBuild/15.0/Bin/MSBuild.exe";
```
bording: MSBuild 15 has moved inside the VS folder, so we can't guarantee it's location any more by simply hardcoding a value.

We'll need to pull in https://github.com/Microsoft/vswhere and figure it out at run time.

adamralph: Sounds good to me. Is this PR still a WIP then?

bording: It probably doesn't make sense to change the script to add this in a separate PR, so yeah I'll go ahead and mark as WIP for now.

bording:
@adamralph I pushed up a change so that vswhere is being used. What do you think of the approach to pass in the path to build.csx via the environment variable?

adamralph:
@bording my initial thought was that this should be done inside the build script, but then I realised that we also need it in `build.cmd`, for the path to `csi.exe`, so I'm 👍 to that approach.

I'll do a full review ASAP. I don't have VS 2017 on my main machine yet, but I guess it's about time, so I'll install it.

bording:
Rebased to resolve conflicts after 6.3.1 merge. 

@adamralph Had a chance to install VS 2017 yet?

WojcikMike:
@bording @adamralph for what is worth I run the tests on my vs 2017 and Rider and no issues with this PR

PR #372, opened 2017-05-08
require VS 2017

adamralph:
I think this should have been done as part of https://github.com/Particular/NServiceBus.RabbitMQ/pull/366

When I open the solution from develop, it opens in VS 2015 (I have both VS 2015 and VS 2017 on this machine) and the solution fails to compile.


bording:
I've gotten so into the habit of right-clicking solution files to open with VS 2017 directly, it didn't occur to me to change this when we actually started requiring VS 2017 after the merge!

PR #373, opened 2017-05-08
converted RabbitMqContext.TryWaitForMessageReceipt() to expression bodied member

adamralph:
Building on the changes in #366

### DIFF
--- a/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
+++ b/src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
@@ -61,10 +61,7 @@
             channelProvider?.Dispose();
         }
 
-        protected bool TryWaitForMessageReceipt()
-        {
-            return TryReceiveMessage(out var message, incomingMessageTimeout);
-        }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -61,10 +61,7 @@ public void TearDown()
             channelProvider?.Dispose();
         }
 
-        protected bool TryWaitForMessageReceipt()
-        {
-            return TryReceiveMessage(out var message, incomingMessageTimeout);
-        }
+        protected bool TryWaitForMessageReceipt() => TryReceiveMessage(out var message, incomingMessageTimeout);
```
bording: One additional tweak here would be to change `message` to a discard (`_`) since we have no use for the result.

adamralph: 👍 I like `_` for unused vars. I'll make the change (will amend directly since the change set is so trivial).

PR #376, opened 2017-05-12
Hotfix 4.3.2

### LINKED ISSUE #374: Delayed retries are not native unless the timeout manager is explicitly disabled
When running in the default backwards-compatible mode where the timeout manager is still enabled, delayed retries are incorrectly sent through the timeout manager.

Once `DisableTimeoutManager` is called, then delayed retries are correctly sent through the delay infrastructure. However, the intention is that they are always sent through the delay infrastructure regardless of this setting.

Given that we say that delayed messages are always sent through the native delay infrastructure, an endpoint that doesn't contain sagas or use the Outbox could be configured for in-memory persistence with no expectation of message loss, but this is not currently the case.

## Who's affected
Anyone using NServiceBus.RabbitMQ 4.3.0 or 4.3.1 without the `DisableTimeoutManager` setting enabled.

## Symptoms
- When using in-memory persistence and you shut down the endpoint, all messages waiting for a delayed retry are lost.
- When using other persistences, you see that new entries in the timeout persistence database are continuing to be created, preventing you from decommissioning the timeout manager with the `DisableTimeoutManager` setting.

### LINKED ISSUE #375: Configuration settings related to the timeout manager are not handled correctly
There are several configuration settings related to the timeout manager that are not handled correctly and lead to incorrect behavior when they are specified without also using the `DisableTimeoutManager` setting.

#### Disabling the timeout manager feature

If the configured persistence does not support timeout storage (NServiceBus 6.0 or later), or if a persistence has not been configured at all (NServiceBus 6.2 or later), the following error message is displayed:

```
The selected persistence doesn't have support for timeout storage. Select another persistence or disable the timeout manager feature using endpointConfiguration.DisableFeature<TimeoutManager>()
```

However, using the setting mentioned in the error message to disable the timeout feature results in the following error:
```
Cannot delay delivery of messages when TimeoutManager is disabled or there is no infrastructure support for delayed messages.
```
This prevents the endpoint from sending delayed messages despite having access to the native delay infrastructure.

#### Send-only endpoints

If the endpoint has been configured as [send-only](https://docs.particular.net/nservicebus/hosting/#self-hosting-send-only-hosting), then the following error occurs when attempting to send a delayed message:
```
Cannot delay delivery of messages when TimeoutManager is disabled or there is no infrastructure support for delayed messages.
```
This prevents the endpoint from sending delayed messages despite having access to the native delay infrastructure.

#### Configuring an external timeout manager address
If an endpoint has been configured to use an external timeout manager through the `UseExternalTimeoutManager` API or the `UnicastBusConfig` `TimeoutManagerAddress` property, the actual value of this setting is ignored, and delayed messages are sent through the native delay infrastructure, except for delayed retries, which will be sent to the external timeout manager (see #374 for details).

-------
When using any of these settings with the `DisableTimeoutManager` setting, then delayed messages work correctly and are sent through the native delay infrastructure.

## Who's affected
Anyone using NServiceBus.RabbitMQ 4.3.0 or 4.3.1 without the `DisableTimeoutManager` setting enabled.

## Symptoms
- You cannot send delayed messages and see an error instead.
- You do not see delayed messages in the timeout persistence storage of an externally configured endpoint.

bording:
The first commit adds a check to `CalculateDelay` to look for timeout manager headers, which will be present on the message if the timeout manager is enabled and the message is a delayed retry. The destination in the header is passed back to the `MessageDispatcher` so it can be used instead if it's found.

The second commit improves the `DelayInfrastructure.CheckForInvalidSettings` to verify that the timeout manager is actually active when we expect to be, and points to `DisableTimeoutManager` if it isn't. There is also some cleanup in `RabbitMQTransportInfrastructure` since `CheckForInvalidSettings` now gets passed the `SettingsHolder` directly.

We could try and identify more specifically the reason when the timeout manager isn't active, but I didn't see that providing a lot of additional value.

Fixes #374 
Fixes #375

### DIFF
--- a/src/NServiceBus.RabbitMQ.sln
+++ b/src/NServiceBus.RabbitMQ.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26430.4
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NServiceBus.RabbitMQ", "NServiceBus.RabbitMQ\NServiceBus.RabbitMQ.csproj", "{BA731749-22C7-4025-8A4D-465AE8E02E61}"
 EndProject
--- a/src/NServiceBus.RabbitMQ.sln.DotSettings
+++ b/src/NServiceBus.RabbitMQ.sln.DotSettings
@@ -1,6 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
 
-	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String>
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String>
 	<s:Boolean x:Key="/Default/CodeEditing/Intellisense/CodeCompletion/AutoCompleteBasicCompletion/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeEditing/Intellisense/CodeCompletion/IntelliSenseCompletingCharacters/IntelliSenseCompletingCharactersSettingCSharp/UpgradedFromVSSettings/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeEditing/Intellisense/LookupWindow/ShowSummary/@EntryValue">True</s:Boolean>
--- a/src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
+++ b/src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
@@ -4,7 +4,9 @@
     using System.Collections;
     using System.Collections.Generic;
     using System.Text;
+    using Features;
     using global::RabbitMQ.Client;
+    using Settings;
 
     static class DelayInfrastructure
     {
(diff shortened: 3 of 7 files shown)


src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -73,13 +75,22 @@ public static void TearDown(IModel channel)
             }
         }
 
-        public static StartupCheckResult CheckForInvalidSetting(bool routingTopologySupportsDelayedDelivery, bool disableTimeoutManager)
+        public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings)
```
SeanFeldman: Is this to ensure that the settings are locked?

bording: This because I know this will run after feature setup, so I can check for the final state of the timeout manager feature when this is called. Doing it earlier would end up being before feature setup, and prerequisites are evaluated as part of feature setup.

src/NServiceBus.RabbitMQ.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26430.4
```
adamralph: Ideally I'd prefer not to see this change in a hotfix but, since it's only tooling, I think it's OK to let it in this time.

src/NServiceBus.RabbitMQ.sln.DotSettings
```diff
@@ -1,6 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
 
-	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String>
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String>
```
adamralph: Ideally I'd prefer not to see this change in a hotfix but, since it's only tooling, I think it's OK to let it in this time.

PR #379, opened 2017-05-12
Hotfix 4.3.2 - additional changes

bording:
Here are the revisions. If the routing topology doesn't support delayed delivery, then delayed retried will continue to work.

The invalid settings have been improved by scoping them based on wether the routing topology supports delayed delivery or not, and some settings actually mean we can go ahead and set `DisableTimeoutManager` to avoid an error condition.

I would still like to add some tests around all of this where possible, so I've marked the PR WIP for now.

Connects to #374 
Connects to #375

### DIFF
--- a/src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
+++ b/src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
@@ -12,7 +12,7 @@
     {
         const int maxNumberOfBitsToUse = 28;
         const int maxLevel = maxNumberOfBitsToUse - 1;
-        const string CoreExternalTimeoutManagerAddressKey = "NServiceBus.ExternalTimeoutManagerAddress";
+        const string coreExternalTimeoutManagerAddressKey = "NServiceBus.ExternalTimeoutManagerAddress";
 
         public const int MaxDelayInSeconds = (1 << maxNumberOfBitsToUse) - 1;
         public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
@@ -79,11 +79,11 @@
         public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings)
         {
             var routingTopologySupportsDelayedDelivery = settings.GetOrDefault<bool>(SettingsKeys.RoutingTopologySupportsDelayedDelivery);
-            var TimeoutManagerDisabled = settings.GetOrDefault<bool>(SettingsKeys.DisableTimeoutManager);
-            var externalTimeoutManagerAddress = settings.GetOrDefault<string>(CoreExternalTimeoutManagerAddressKey) != null;
+            var timeoutManagerDisabled = settings.GetOrDefault<bool>(SettingsKeys.DisableTimeoutManager);
+            var externalTimeoutManagerAddress = settings.GetOrDefault<string>(coreExternalTimeoutManagerAddressKey) != null;
             var timeoutManagerFeatureActive = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Active;
 
-            if (!routingTopologySupportsDelayedDelivery && TimeoutManagerDisabled)
+            if (!routingTopologySupportsDelayedDelivery && timeoutManagerDisabled)
             {
                 return StartupCheckResult.Failed($"Cannot disable the timeout manager when the specified routing topology does not implement {nameof(ISupportDelayedDelivery)}.");
             }
@@ -95,7 +95,7 @@
                     return StartupCheckResult.Failed("An external timeout manager address cannot be configured because the timeout manager is not being used for delayed delivery.");
                 }
 
-                if (!TimeoutManagerDisabled && !timeoutManagerFeatureActive)
+                if (!timeoutManagerDisabled && !timeoutManagerFeatureActive)
                 {
                     return StartupCheckResult.Failed("The timeout manager is not active, but the transport has not been properly configured for this. " +
                         "Use 'EndpointConfiguration.UseTransport<RabbitMQTransport>().DelayedDelivery().DisableTimeoutManager()' to ensure delayed messages can be sent properly.");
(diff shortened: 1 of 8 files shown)


src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -32,7 +33,19 @@ public RabbitMQTransportInfrastructure(SettingsHolder settings, string connectio
...
-            settings.Set(SettingsKeys.RoutingTopologySupportsDelayedDelivery, routingTopology is ISupportDelayedDelivery);
+            var routingTopologySupportsDelayedDelivery = routingTopology is ISupportDelayedDelivery;
+            settings.Set(SettingsKeys.RoutingTopologySupportsDelayedDelivery, routingTopologySupportsDelayedDelivery);
+
+            if (routingTopologySupportsDelayedDelivery)
+            {
+                var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
+                var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
+
+                if (timeoutManagerFeatureDisabled || sendOnlyEndpoint)
```
adamralph: I don't know much about send-only endpoints, but why is it that we disable the timeout manager in the case of a send-only endpoint?

/cc @Particular/nservicebus-maintainers

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -79,16 +78,16 @@ public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings
...
+            if (!routingTopologySupportsDelayedDelivery && TimeoutManagerDisabled)
             {
                 return StartupCheckResult.Failed($"Cannot disable the timeout manager when the specified routing topology does not implement {nameof(ISupportDelayedDelivery)}.");
             }
 
-            if (!TimeoutManagerDisabled && !timeoutManagerFeatureActive)
+            if (routingTopologySupportsDelayedDelivery && externalTimeoutManagerAddress)
             {
-                return StartupCheckResult.Failed("The timeout manager is not active, but the transport has not been properly configured for this. Use 'EndpointConfiguration.UseTransport<RabbitMQTransport>().DelayedDelivery().DisableTimeoutManager()' to ensure delayed messages can be sent properly.");
+                return StartupCheckResult.Failed("Configuring an external timeout manager address is not supported.");
```
adamralph: Should this give some more information as to why? E.g.

> An external timeout manager address cannot be configured. The timeout manager is not being used because the routing topology supports delayed delivery.

?

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
bording: Send-only endpoints have to disable the timeout manager, because they don't start any message pumps, and the timeout manager relies on receiving messages from its satellite queues.

This is one of the prerequisites for the TimeoutManger feature to be active: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L22

So, if this is set, we know the timeout manage will be disabled, so it's safe to tell core we support the native delivery constraints.

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
bording: I suppose adding a "why" to the message could be useful.

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -79,16 +78,16 @@ public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings
         {
             var routingTopologySupportsDelayedDelivery = settings.GetOrDefault<bool>(SettingsKeys.RoutingTopologySupportsDelayedDelivery);
             var TimeoutManagerDisabled = settings.GetOrDefault<bool>(SettingsKeys.DisableTimeoutManager);
-            var timeoutManagerFeatureActive = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Active;
+            var externalTimeoutManagerAddress = settings.GetOrDefault<string>("NServiceBus.ExternalTimeoutManagerAddress") != null;
```
SeanFeldman: We used a class with consts for headers to know where those strings are utilized. Should this be included?

bording:
I've added some tests around the invalid settings check, though it's hard to be exhaustive on all combinations of settings.

I also can't yet come up with a good way to test the delayed retries behavior.

 I believe all feedback has been addressed, so removing the WIP on this.

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
WojcikMike: @adamralph https://docs.particular.net/nservicebus/messaging/delayed-delivery#caveats

> Delayed delivery of messages is supported when the one following requirements are met:

>  - The transport supports delayed delivery natively.
>  - The Timeout Manager feature is enabled and the endpoint is not a send-only endpoint.

bording:
@adamralph I have squashed the fixup commits and changed the core settings keys into consts as requested.

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -12,6 +12,7 @@ static class DelayInfrastructure
     {
         const int maxNumberOfBitsToUse = 28;
         const int maxLevel = maxNumberOfBitsToUse - 1;
+        const string CoreExternalTimeoutManagerAddressKey = "NServiceBus.ExternalTimeoutManagerAddress";
```
adamralph: I missed this one in my review (should also be camel cased).

src/NServiceBus.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -15,6 +16,9 @@
     [SkipWeaving]
     sealed class RabbitMQTransportInfrastructure : TransportInfrastructure, IDisposable
     {
+        const string CoreSendOnlyEndpointKey = "Endpoint.SendOnly";
+        const string CoreHostInformationDisplayNameKey = "NServiceBus.HostInformation.DisplayName";
```
adamralph: And these ones.

PR #382, opened 2017-05-17
Migrate to VS2017 SDK-style project format

bording:
This updates the solution to the new VS 2017 SDK-style csproj format.


src/NServiceBus.RabbitMQ/AssemblyInfo.cs
```diff
@@ -1,9 +1,3 @@
-﻿using System.Reflection;
-using System.Runtime.InteropServices;
+﻿using System.Runtime.InteropServices;
 
-[assembly: AssemblyTitle("NServiceBus.RabbitMQ")]
-[assembly: AssemblyCopyright("Copyright NServiceBus. All rights reserved")]
-[assembly: AssemblyProduct("NServiceBus.RabbitMQ")]
```
adamralph: Because the csproj does not have the equivalent elements for `AssemblyTitle` and `AssemblyProduct`, they are defaulting to:

```C#
// from ILSpy
[assembly: AssemblyProduct("NServiceBus.Transports.RabbitMQ")]
[assembly: AssemblyTitle("NServiceBus.Transports.RabbitMQ")]
```

So, a change from what they currently are. Does that matter?

src/NServiceBus.RabbitMQ/AssemblyInfo.cs
```diff
@@ -1,9 +1,3 @@
...
-[assembly: AssemblyCompany("NServiceBus Ltd.")]
-[assembly: AssemblyConfiguration("release")]
```
adamralph: This is going to sound really stupid, but because this is no longer explicitly specified, the casing in the output assembly is different:

```C#
// from ILSpy
[assembly: AssemblyConfiguration("Release")]
```

Does that matter?

src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
```diff
@@ -1,286 +1,28 @@
...
-    <Compile Include="When_publishing_message_implementing_interface_in_direct_topology.cs" />
-    <Compile Include="When_using_direct_routing.cs" />
-    <Compile Include="When_the_broker_connection_is_lost.cs" />
-    <Compile Include="When_using_a_custom_message_id_strategy.cs" />
-    <Compile Include="When_scaling_out_senders_that_uses_callbacks.cs" />
-    <Compile Include="When_scaling_out_subscribers.cs" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
+    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="6.3.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.7.0" />
```
adamralph: 😍

src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
```diff
@@ -1,286 +1,28 @@
...
-    <Compile Include="TaskEx.cs" />
-    <Compile Include="When_receiving_a_reply_that_contains_a_legacy_callback_header.cs" />
-    <Compile Include="When_the_message_contains_a_legacy_callback_header.cs" />
-    <Compile Include="When_publishing_message_implementing_interface_in_direct_topology.cs" />
-    <Compile Include="When_using_direct_routing.cs" />
-    <Compile Include="When_the_broker_connection_is_lost.cs" />
-    <Compile Include="When_using_a_custom_message_id_strategy.cs" />
-    <Compile Include="When_scaling_out_senders_that_uses_callbacks.cs" />
-    <Compile Include="When_scaling_out_subscribers.cs" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
```
adamralph: For my own education, what is this for? (I think you mentioned it before but I've forgotten.)

build.cmd
```diff
@@ -24,10 +24,10 @@ if not exist %NUGET_LOCAL_DIR%\NuGet.exe (
 )
 
 :: restore packages
-%NUGET_LOCAL_DIR%\NuGet.exe restore .\packages.config -PackagesDirectory ./src/packages -MSBuildVersion 15 -Verbosity quiet
+%NUGET_LOCAL_DIR%\NuGet.exe restore .\packages.config -PackagesDirectory ./build-packages -MSBuildVersion 15 -Verbosity quiet
```
adamralph: I think I'd prefer to call `build-packages` just `packages` if that works. Otherwise, we should probably rename `packages.config` to `build-packages-config`. I guess either strategy could work, although I'm leaning towards the shorter names.

src/NServiceBus.RabbitMQ/AssemblyInfo.cs
bording: I don't believe those have an impact from a breaking-change standpoint, but then can be viewed in the Details tab if you right-click on the assembly and view properties.

In this PR I was leaning towards trying to keep things as close to how they were before, but we do still need to clean up everything here to align to the component naming guidelines, which would be `NServiceBus.Transport.RabbitMQ`.

It looks like I can add the `Product` property to the csproj, and that wires up the attribute correctly. There is also a `Title` property (which I believe was a late addition based on an issue I recall seeing) that changes the package's `Title` value, but it doesn't appear to be properly hooked up to generating the assembly attribute.

I'll go ahead and add the Title attribute back for now, and in a subsequent PR we can make changes to match the naming conventions.

src/NServiceBus.RabbitMQ/AssemblyInfo.cs
bording: That one really shouldn't matter AFAIK, and `Release` is more correct anyway.

src/NServiceBus.RabbitMQ.AcceptanceTests/NServiceBus.RabbitMQ.AcceptanceTests.csproj
bording: It adds the required properties and targets to make test projects work properly in the new project system, for both in VS, and `dotnet test`.

build.cmd
bording: The thought behind this was that this folder is now only for the build dependencies, since everything else is now using PackageReferences. That was why I moved it into the repo root instead of being in `/src`. Once it was there, after running the build script, we'd end up with both a `packages` folder and a `packaging` folder. I didn't care for how similar those looked, so I added `build` to the name of the folder.

The next step after merging this is to use the built-in packaging capabilities, as I've done [here](https://github.com/Particular/NServiceBus.Transport.Msmq/pull/3/commits/b402c6da5d4ce23754052119c63359fbd46ab198) for MSMQ, so once that is done, the `packaging` folder would be gone.

While I still think there is some value in indicating the folder it just for build script dependencies, I'm more OK with just calling it `packages` at that point.

adamralph: OK, I'm fine with `build-packages` for now.

src/NServiceBus.RabbitMQ/AssemblyInfo.cs
adamralph: https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json-to-csproj talks about `<AssemblyTitle>My library</AssemblyTitle>` - is that the property which doesn't work properly?

bording: Ah, it looks like `AsssemblyTitle` does work. I guess since they didn't originally intend to expose `Title` as a NuGet property, they created a separate property to handle that directly, sigh. I'll update and add both of those then.

bording:
@adamralph I've pushed up the change to add Product and AssemblyTitle.

PR #383, opened 2017-05-25
Channel improvements

bording:
This improves the event handler methods in `ConfirmsAwareChannel` by limiting the code inside of the new task to be just the `TaskCompletionSource.SetResult` and `TaskCompletionSource.SetException` calls.

It also updates Particular.CodeRules to 0.2.0, which includes a rule to check for "missing" awaits. The `TaskCompletionSource` calls are intentionally not waited on, so `Ignore()` has been added to satisfy the new rule.

### DIFF
--- a/src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
+++ b/src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs
@@ -127,103 +127,83 @@ namespace NServiceBus.Transport.RabbitMQ
             return tcs.Task;
         }
 
-        class EventState
-        {
-            public ConcurrentDictionary<ulong, TaskCompletionSource<bool>> Messages { get; set; }
-
-            public bool Multiple { get; set; }
-
-            public ulong DeliveryTag { get; set; }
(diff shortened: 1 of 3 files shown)


WojcikMike:
@bording plese merge at your convenience.

adamralph:
Is there any functional change here? Or pure refactoring?

bording:
@adamralph There is an actual change here in the case where you would receive both a basic return and a basic ack. This isn't very likely to happen. The most likely way I know where it would be when using the direct routing topology, publisher confirms enabled, and you send to non-existent endpoint.

When this happens, the basic return is sent first, and then the basic ack, so you'd want to process them in that order. However, since we had all of the logic in the task, if the basic ack task got scheduled first, you could potentially have the `TaskCompletionSource` end up with the task successfully completed instead of faulted.

Now that most of the event handler code is executed synchronously outside of the task, this should not be possible, because the basic return handler will execute first, removing the `TaskCompletionSource` from the collection before the basic ack handler will ever have a chance to see it.

bording:
Based on that, I suppose this should have the bug tag.

PR #384, opened 2017-05-25
Naming conventions

bording:
This updates as much as possible to the component naming conventions. Items that haven't been changed yet because the would be breaks/visible changes of some kind:

- The assembly name
- The Product and Title assembly attributes
- The NuGet package id
- The solution name
- The repo name

FYI, this is built on top of #383, so that needs to be merged first and then this can be rebased.


WojcikMike:
The change matches the https://github.com/Particular/NServiceBus/blob/develop/guidelines/design-guidelines.md#nservicebus-rules so LGTM. I will review it in more details today

adamralph:
I've renamed the issue, and I think it may need to go into release notes, since there are some visible changes, even if trivial.

adamralph:
Apologies, I misread the description. This shouldn't appear in release notes. Updated the title again.

PR #385, opened 2017-05-25
Native packaging

bording:
This moves away from NuGetPackager and a separately defined nuspec file to creating the package directly from the project file.

I've also added in SourceLink, so we can get source linking for free.

FYI, this is built on top of #383 and #384, so they need to be merged first and then this can be rebased.


src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
-    <AssemblyTitle>NServiceBus.RabbitMQ</AssemblyTitle>
   </PropertyGroup>
 
   <ItemGroup>
     <Reference Include="System.Configuration" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="GitVersionTask" Version="3.6.5" PrivateAssets="All" />
+    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" PrivateAssets="All" />
```
WojcikMike: Is there any reason to use beta version?

bording: Yeah, the 4.0.0 beta is required to properly hook into the new project system correctly so that GitVersion's version info is used when creating the package.

bording:
I think the build failure is related to the SourceLink package, so I'm going to remove that for now, and look into what's going on there.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
WojcikMike: Ok

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
     <PackageReference Include="Janitor.Fody" Version="1.4.1" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="6.0.0" />
-    <PackageReference Include="NuGetPackager" Version="0.6.1" PrivateAssets="All" />
+    <PackageReference Include="NServiceBus" Version="[6.0.0, 7.0.0)" PrivateAssets="None" />
     <PackageReference Include="Obsolete.Fody" Version="4.2.2" PrivateAssets="All" />
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
-    <PackageReference Include="RabbitMQ.Client" Version="4.1.1" />
+    <PackageReference Include="RabbitMQ.Client" Version="[4.1.0, 4.2.0)" PrivateAssets="None" />
```
bording: One thing I want to point out is that previously were were building against 4.1.1, but the nuspec was still using 4.1.0 as the minimum version. I've aligned the version with the nuspec version, but maybe we should consider changing this? Since 4.1.2 has a bug that breaks us, maybe we should go ahead and bump this to 4.1.3 (which fixes the bug) ?

WojcikMike: I am leaning towards bumping the version to 4.1.3 simply to protect our customers, @adamralph thoughts?

adamralph: I'm not sure we should break the general rule of declaring the _minimum_ dependency. Sure, some versions in the range from that minimum to the latest may have bugs, but that's a concern of that package.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
+  </ItemGroup>
+
+  <PropertyGroup>
+    <PackageId>NServiceBus.RabbitMQ</PackageId>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
```
adamralph: (Massive nitpick) The current package has a full stop after the "Ltd" abbreviation in the assembly attribute, but not in the nupkg. I guess these can't be separated, and it doesn't matter anyway.

Also, British English favours omission of the full stop, so naturally I'm an enormous fan of this change. 😉

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
     <Reference Include="System.Configuration" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="GitVersionTask" Version="3.6.5" PrivateAssets="All" />
+    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.4.1" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="6.0.0" />
-    <PackageReference Include="NuGetPackager" Version="0.6.1" PrivateAssets="All" />
+    <PackageReference Include="NServiceBus" Version="[6.0.0, 7.0.0)" PrivateAssets="None" />
```
adamralph: Just to check: does this mean that, during build, the _minimum_ required version, i.e. 6.0.0 is restored and used for compilation and testing?

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
+  </ItemGroup>
+
+  <PropertyGroup>
+    <PackageId>NServiceBus.RabbitMQ</PackageId>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <Description>RabbitMQ support for NServicebus</Description>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <Copyright>Copyright 2010-2017 NServiceBus. All rights reserved</Copyright>
```
adamralph: Does the introduction of copyright years mean that we have to update this property before the first release of each year? Does that matter?

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
-    <PackageReference Include="RabbitMQ.Client" Version="4.1.1" />
+    <PackageReference Include="RabbitMQ.Client" Version="[4.1.0, 4.2.0)" PrivateAssets="None" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <PackageId>NServiceBus.RabbitMQ</PackageId>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <Description>RabbitMQ support for NServicebus</Description>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
```
adamralph: How does the value of this property (and some of the others) currently get populated in `develop`? NuGetPackager magic?

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
+    <Product>NServiceBus.RabbitMQ</Product>
+    <AssemblyTitle>NServiceBus.RabbitMQ</AssemblyTitle>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
-    <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
-    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
```
adamralph: Were these three properties redundant, and restating the defaults?

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
-    <PackageReference Include="RabbitMQ.Client" Version="4.1.1" />
+    <PackageReference Include="RabbitMQ.Client" Version="[4.1.0, 4.2.0)" PrivateAssets="None" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <PackageId>NServiceBus.RabbitMQ</PackageId>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <Description>RabbitMQ support for NServicebus</Description>
```
adamralph: The value of this property gets put into an `AssemblyDescription` attribute in the assembly, which doesn't currently happen in `master` (`develop` currently has a regression where the value "Package Description" is used) but I guess this change is OK, and desirable.

packaging/nuget/NServiceBus.Transport.RabbitMQ.nuspec
```diff
@@ -1,25 +0,0 @@
...
-    <title>NServiceBus RabbitMQ</title>
-    <version>$version$</version>
-    <authors>$authors$</authors>
-    <owners>$owners$</owners>
-    <licenseUrl>$licenseUrl$</licenseUrl>
-    <projectUrl>$projectUrl$</projectUrl>
-    <iconUrl>$iconUrl$</iconUrl>
-    <requireLicenseAcceptance>$requireLicenseAcceptance$</requireLicenseAcceptance>
-    <description>RabbitMQ support for NServicebus</description>
-    <releaseNotes></releaseNotes>
```
adamralph: The current package has an empty `releaseNotes` element but since there is no equivalent property in the project file here, the element does not appear in the package produced by this PR. I'm guessing this is fine, but I just wanted to call it out in case someone knows of a bizarre reason that this matters.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -3,32 +3,48 @@
...
+    <Copyright>Copyright 2010-2017 NServiceBus. All rights reserved</Copyright>
+    <PackageTags>nservicebus servicebus msmq cqrs publish subscribe</PackageTags>
+    <PackageIconUrl>http://s3.amazonaws.com/nuget.images/NServiceBus_32.png</PackageIconUrl>
+    <PackageProjectUrl>http://particular.net/</PackageProjectUrl>
+    <PackageOutputPath>..\..\nugets</PackageOutputPath>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="$(OutputPath)$(AssemblyName).pdb">
```
adamralph: I'm more than a little surprised that this is required. I would have expected PDB's to be included by default.

bording:
@adamralph Pushed up an amended commit. Instead hardcoding the 2017 in the copyright, it's now `$([System.DateTime]::UtcNow.ToString(yyyy))`.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: As discussed in today's sync meeting, going to to keep this at 4.1.0.

packaging/nuget/NServiceBus.Transport.RabbitMQ.nuspec
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<package xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-  <metadata xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-    <id>NServiceBus.RabbitMQ</id>
-    <title>NServiceBus RabbitMQ</title>
```
bording: Decided to go ahead and remove it in anticipation if it going away. Makes more sense to just use package id anyway.

packaging/nuget/NServiceBus.Transport.RabbitMQ.nuspec
bording: Yeah, this shouldn't matter. Maybe if they ever update the gallery to include an actual link, we can add this.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: GitVersion 3.x did not know how to work with the new project system, so it was conflicting by generating duplicate properties. GitVersion 4.x knows how to fill its information into the right spots, so we can let the project generate these values again.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: Correct

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: The assembly value was using "Ltd.", but the package itself got its values from NuGetPackager, which provided "Ltd". I decided to align everything on the package values.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: Yeah, the project system putting "Package Description" if you don't specify anything is a bug, IMHO. I need to file an issue.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: Yep, NuGetPackager has all the defaults and fills them into the nuspec placeholders.

src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
bording: NuGet has this crazy idea that you don't want PDBs in the main package, and wants to put them into a separate .Symbols package. It's been that way as long as I can remember.

PR #386, opened 2017-05-26
[WIP] Multi-targeting

bording:
Here's one more idea I had. Since the new project format supports mult-targeting, we can add a net462 target, which then lets us use `TaskCreationOptions.RunContinuationsAsynchronously` conditionally.

This is on top of all the other PRs, so this would need to be merged last, if we think it's worth doing.


adamralph:
Renamed to WIP while 385 isn't merged.

adamralph:
As discussed in yesterday's sync, this is a useful spike, but we're not going to ship a separate net462 assembly for this purpose.

PR #390, opened 2017-06-01
reduce line length of some recently touched files

adamralph:
When I was reviewing a recent PR, a continuous paper cut was having to scroll left and right repeatedly to review the diff, both in GitHub and in my local diff viewer. This PR is an attempt at removing some of that pain.

In some other OSS projects, we've adopted a loose rule of limiting line length to ~120 chars for this purpose. The rule isn't strict, but we apply it were it doesn't add friction in itself.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs
@@ -79,7 +79,8 @@
             }
         }
 
-        public override OutboundRoutingPolicy OutboundRoutingPolicy => new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
+        public override OutboundRoutingPolicy OutboundRoutingPolicy =>
+            new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
 
         public override TransportTransactionMode TransactionMode => TransportTransactionMode.ReceiveOnly;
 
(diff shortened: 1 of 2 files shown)


bording:
I can't help but look at this PR see a bunch of unnecessary line breaks. 😢. I personally find this kind formatting much more difficult to parse, making the code harder to understand.

I'm 👎 for merging this. @WojcikMike what do you think?

WojcikMike:
@bording I hoped it will not come to that. I am indifferent. I can read code formatted in both ways. Personally, I don't have strict rules about breaking lines and I do it inconsistently. 

I think it is more personal preference kind of setting. And as we have 1 supporter and 1 opposer I would vote against it. We can discuss it on the weekly sync

adamralph:
OK, I've dialled back on some of the changes (splitting on conditional operators, assignments and single argument) but dialled up slightly on the long argument lists, which is a change I find it difficult to argue against.

I also reversed some of the `&&` conditionals which not only move the shorter expression to the front, but also makes them more performant (e.g. a null check vs a lookup). These changes could be pulled out into a separate PR if this one doesn't make it in.

Happy to discuss this PR in our next sync.

adamralph:
> I also reversed some of the `&&` conditionals which not only moves the shorter expression to the front, but also makes them more performant (e.g. a null check vs a lookup). These changes could be pulled out into a separate PR if this one doesn't make it in.

OK, so that was a slight brain fart. This new-fangled `out var` threw me off. 😊 

Pushed another fixup! to remove those changes.

adamralph:
Pushed one more fixup! to dial back on a few changes.

adamralph:
And one more. That's all for now.

adamralph:
@Particular/rabbitmq-transport-maintainers I've reduced the scope of this PR even further to just those lines which I found particularly painful to parse during recent reviews. How do you feel about the new diff?

PR #391, opened 2017-06-02
upgrade to 6.3.3 test packages

adamralph:
🎉 upgrading is now a joy!


WojcikMike:
it's not only that, but also the dif is so much smaller. Now one can actually review it in github 🎉

PR #392, opened 2017-06-02
[WIP] switch from IDisposable to TransportInfrastructure.Stop()

adamralph:


SimonCropp:
looks good. but adding `sealed` is just noise

adamralph:
It's true that `sealed` is redundant while the type doesn't perform unmanaged disposal and there is no inheriting type that needs to do unmanaged disposal. The types are internal, so the use of `sealed` is purely defensive against a mistake we may make in the future. @Particular/rabbitmq-transport-maintainers any opinion on that? For the cost of one extra modifier, I'd probably vote to keep it. It does convey to the reader that "this type isn't ready to be (generally) inherited from yet".

SimonCropp:
> For the cost of one extra modifier, I'd probably vote to keep it. It does convey to the reader that "this type isn't ready to be (generally) inherited from yet".

if a code review wont catch that we have much bigger problems

adamralph:
@Particular/rabbitmq-transport-maintainers discussed this today.

Ultimately, the disposable pattern only works here because of a side effect of NSB core putting the instance of `TransportInfrastructure` into a container, which just happens to discover the implementation of `IDisposable` on the inheriting type (`RabbitMQTransportInfrastructure`). The correct pattern for the transport seam is to override the `TransportInfrastructure.Stop()` method. This PR will be re-purposed to switch from `IDisposable` to `Stop()`.

danielmarbach:
@adamralph `TransportInfrastructure` should not be added to the container. It is managed in the `SettingsHolder` which implements a [`Clear`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/SettingsHolder.cs#L303) method that disposes of all values in the settings dictionary that implement `IDisposable`. So the switch is not really necessary

adamralph:
@danielmarbach thanks for the clarification. Still, even though it's `SettingsHolder` and not `Autofac` doing the disposal, would you agree that overriding the `TransportInfrastructure.Stop()` method is the correct pattern to use? Presumably core will respect that and call the method?

danielmarbach:
@adamralph yes it well. I would agree it is the best way to move forward.

adamralph:
replaced by #436 (from an appropriately named branch)

PR #393, opened 2017-06-07
Changed the way PreventRoutingMessagesToTimeoutManager feature is enabled

SzymonPobiega:
EnabledByDefault is evil in the downstreams because it makes the feature globally enabled i.e. if there is any other endpoint in the same AppDomain, it will get this feature enabled no matter which transport it uses. 

In my case I had an MSMQ and RabbitMQ endpoint in same console app. The MSMQ endpoint attempted to setup the PreventRoutingMessagesToTimeoutManager feature but crashed because a mandatory setting key was not present.

This change will only enable the feature in the endpoints that actually use RabbitMQ.


adamralph:
So is this effectively a bug? Something like "PreventRoutingMessagesToTimeoutManager feature is enabled for other endpoints in the same AppDomain"?

SzymonPobiega:
@adamralph yes, this is a bug.

adamralph:
👍 @SzymonPobiega thanks. What do you think of my suggested title?

bording:
I pushed up a minor formatting revision.

Since this is a small change, I'm OK with making it, but is more than one endpoint in a single AppDomain, using different transports even a supported scenario though?

I would sort of expect there to be more issues with making that work correctly than just this. How would you handle separating all of the other things that would end up being shared?

@adamralph Looking at this, I wonder if the feature could be simplified further by moving the `EnableFeatureByDefault` call into the `if (routingTopologySupportsDelayedDelivery)` check in the `RabbitMQTransportInfrastructure` instead.

SzymonPobiega:
@bording Yes, it is a supported scenario. We even have samples showing such a setup e.g. https://docs.particular.net/samples/msmqtosqlrelay/.

adamralph:
>  I wonder if the feature could be simplified further by moving the EnableFeatureByDefault call into the if (routingTopologySupportsDelayedDelivery) check in the RabbitMQTransportInfrastructure instead

@bording would that be purely a simplification, or any functional change? It strikes me that that call should always only have been made when the routing topology supports delayed delivery, so do we have another bug there?

bording:
@adamralph It would just be a simplification. The feature already checks that value directly, so there's no bug.

bording:
@SzymonPobiega Maybe I'm missing something, but I don't see how that sample is an example of this scenario. Each project has a single endpoint, and the bridge isn't using NSB at all.

If two endpoints are running the same process, they'd have the same scanned assemblies, the same handlers, and the same log files. This seems really weird to me.

SzymonPobiega:
@bording sorry, I though the original intention in this sample was to have two NSB endpoints. 

I would not say that because we have a flawed assembly scanning API (that most of us agree is painful) we should not keep other things clean and separated. Just because logging is still static (and acceptance tests experience the pain of this) does not mean we should allow statics all over the code. Quite the contrary, in the V5 time frame we made a huge effort to get rid of all the statics to allow multi-hosting of NSB endpoints in the same process. There should be good reasons if we want to move away from this guidance.

bording:
One minor issue about this change is that by calling `EnableFeatureByDefault` here, I believe it's too late for the `DisplayDiagnosticsForFeatures` class in core to correctly display that the feature was enabled by default.

SzymonPobiega:
Closed, based on a workaround proposed by @bording in https://github.com/Particular/NServiceBus/issues/4804

PR #395, opened 2017-06-13
[WIP] Changing ConventionalRoutingTopology to not create unnecessary exchange

WojcikMike:
Changing ConventionalRoutingTopology to not create an exchange for System:Object


WojcikMike:
It turned out that there is a test: When_subscribed_to_a_event
And it looks like this:
`
public async Task Subscribing_to_Object_should_subscribe_to_all_published_messages()
        {
            await Subscribe<object>();

            await Publish<MyOtherEvent>();
            await Publish<MyEvent>();

            AssertReceived<MyOtherEvent>();
            AssertReceived<MyEvent>();
        }
`
Which breaks (at least locally) after my change. So because aparently we are supporting subscription for objects let me close this PR

PR #396, opened 2017-06-27
Add .NET Core 2.0 target

bording:

PR #398, opened 2017-07-09
Use current  alpha

SimonCropp:


danielmarbach:
Should this be closed in favour of https://github.com/Particular/NServiceBus.RabbitMQ/pull/396 ?

SimonCropp:
@adamralph 

>My only concern with this is that a new alpha may be pushed out at any time, with breaking changes with respect to the last alpha, and that could introduce build failures to other current WIP.

and that is the desired outcome. while u r chasing an alpha of the core you should be always building and testing against the latest

adamralph:
@danielmarbach this is a PR into the branch for #396.

@SimonCropp 

> and that is the desired outcome. while u r chasing an alpha of the core you should be always building and testing against the latest

That is true, but I don't think having the `develop` branch always pulling down latest core alpha is the best solution. If we did this, then after the `netcore` branch is merged into `develop`, every alpha push of core risks a CI failure for either `develop` or any current WIP PR. I'd prefer that we update to the latest alpha explicitly via a PR, which can also fix anything which is breaking with respect to the previously used alpha.

adamralph:
Rebased.

@Particular/rabbitmq-transport-maintainers discussed this today and decided that we will keep this PR open to provide us with a manual canary build to give us warnings of alpha breakages in core. When the migration to NSB 7 alpha has been merged into develop, we'll rebase the branch onto develop and open a new canary PR to develop.

adamralph:
replaced by #400

SimonCropp:
@adamralph little confused. The other pr looks exactly the same

adamralph:
@SimonCropp the other PR is to `develop`.

adamralph:
And I've just realised you can now change the target branch 😊

Issue #134, opened 2016-03-24
ConventionalRoutingTopology pub/sub allows the use of nested types without distinguishing them from un-nested types


bording:
The default ConventionalRoutingTopology has a problem where it can treat different classes in the same namespace as the same for pub/sub, meaning that subscribing to one of the classes will result in receiving messages for all of the related classes.

This can happen when there are nested classes involved. For example, given the following:

```
namespace Foo
{
    public class ClassA : IEvent
    {

    }

    public class ClassB
    {
        public class ClassA : IEvent
        {
        }
    }

    public class ClassC
    {
        public class ClassA : IEvent
        {

        }
    }
}
```

If an endpoint subscribes to any of the `ClassA` events, it will receive a message when any of the `ClassA`s are published.

The reason this happens is that the exchange name that represents the type is created by using just the [namespace and type name](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.RabbitMQ/Routing/ConventionalRoutingTopology.cs#L84)

This means that the nested aspect of the class is lost, and all three will use the same exchange, `Foo:ClassA`.

Instead, it should use the `FullName` of the type, which would preserve the nested aspect and create three separate exchanges.

However, making this change would break backwards compat, so it seems like we can't do that at this point.

Given that this sort of class setup is likely to be rare, perhaps we just document this as something to avoid instead?

## Plan of attack

- [ ] Throw an exception when a message type being published is nested, with an opt-out config option
- [ ] Update docs, including upgrade guide

adamralph:
Discussed this in the @Particular/rabbitmq-transport-maintainers sync today. This is something we can fix, but it would break wire compatibility.

One option put forward to is to create a completely new (opt-in) topology. In practice that would require a simultaneous switch to the new topology for all affected endpoints, plus a clear out of the exchanges from the old topology.

adamralph:
@Particular/rabbitmq-transport-maintainers discussed this today. We will throw an exception when a message type is nested. The fix for an endpoint will be to un-nest the message type. This will not break wire compatibility.

adamralph:
Added a PoA.

adamralph:
Closing, since we will not be fixing this bug. Rather we will [remove the conventional routing topology](https://github.com/Particular/NServiceBus.RabbitMQ/issues/419) at some point. It will be [deprecated in 5.0.0](https://github.com/Particular/NServiceBus.RabbitMQ/issues/411).

PR #403, opened 2017-07-21
Test cleanup

bording:
Two changes here. The first is that AcceptanceTests now uses `QueueBindings` to know what queues to purge, like we were already doing in TransportTests. I tweaked the TransportTests implemenation so that both are consistent.

I also removed some unneeded code from the APIApprovals test, which had the side effect of introducing some whitespace in the approval file.

### DIFF
--- a/src/NServiceBus.RabbitMQ.sln
+++ b/src/NServiceBus.RabbitMQ.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26711.1
+VisualStudioVersion = 15.0.26720.2
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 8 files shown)


bording:
I found that TeamCity has a plugin that should get test reporting for .NET Core, but it seems to be exploding all the tests. I'll figure this out next week.

bording:
Even after fixing one of the issues with the TeamCity plugin, I was still seeing `Could not find a test logger with URI or FriendlyName 'teamcity'`, so I decided to remove it from this PR, and I'll try again later in a separate PR.

The reason we've had 5 failing tests initially was because by removing the "wipe all queues between tests" logic, we're now hitting #134 for those tests. 😢 The reset broker script was also not working correctly, guaranteeing we'd have leftover messages that would cause the tests to fail.

I've got the reset script working again on the .NET Core build agent, and that seems to have the sequencing correct to not cause a test failure, but #134 is still lurking in those tests.

I'll have to see If I can get the tests changed in Core to ensure we don't have a problem since I don't know of another workaround.

PR #410, opened 2017-08-07
Target .NET Standard 2.0 instead

bording:


bording:
I updated the inspections version to the latest EAP, and now everything is passing.

PR #412, opened 2017-08-08
combine IDeclareQueues and IRoutingTopology

### LINKED ISSUE #265: Combine IRoutingTopology and IDeclareQueues
`IDeclareQueues` was introduced in #248 to allow topologies to control the declaration of queues, without making a breaking change. In the next major we will remove `IDeclareQueues` and replace the current `IRoutingTopology.Initialize()` with a method that accepts all queue addresses, and implementations of that method will be expected to declare queues.

Note that we should also consider changing the names of the `addresses` parameters when we make this change, if https://github.com/Particular/NServiceBus/issues/4151 has been released. If https://github.com/Particular/NServiceBus/issues/4151 hasn't been released, then we can spin off another issue to make that change in some future major.

## Plan of attack

- [x] Combine interfaces - #412
- [x] Update docs, including the upgrade guide - https://github.com/Particular/docs.particular.net/pull/3256

adamralph:
connects to #265


adamralph:
@WojcikMike I've pushed a fixup! to fix the code inspection failures.

> should we before merging interfaces mark IDeclareQueues as obsolete?

I'm not sure that we need to. There will be a vanishingly small number of customers inheriting from it, and the code change required after updating is trivial. I'm also not sure how much value it will add, because `IRoutingTopology.void Initialize` has to change from `void Initialize(RabbitMQ.Client.IModel channel, string main)` to `void Initialize(RabbitMQ.Client.IModel channel, string main)` and there's no obsoleting that can be done there.

We will, of course, need to provide an upgrade guide. I've added that to the PoA of #265.

WojcikMike:
My thinking was that if we would Obsolete the IDeclareQueues in that highly unlikely case of someone using it we could display a message saying that it was merged into IRoutingTopology. Now they would get a compiler error that type doesn't exist. I think that upgrade guide would be sufficient in this case, given how unlikely it is for someone to use this interface.

adamralph:
@bording before we merge, do you have any opinion on deprecation vs removal of `IDeclareQueues`?

bording:
I'm OK with just removing it and relying on the upgrade guide.

PR #413, opened 2017-08-11
Handle v5 obsoletes

bording:
This moves the single obsolete method into a separate file.

It also removes the check for the obsolete app settings that were removed in v4, allowing us to remove the dependency on System.Configuration.


bording:
I suppose there's an argument for keeping the obsolete app settings check around, but it didn't seem important enough to me. The worst case is that there might still be app settings defined that don't actually do anything.

Issue #407, opened 2017-07-31
Allow client certificate authentication with an external auth mechanism


dagapitov:
I'm trying to configure NServiceBus (v 6.3.4) with RabbitMQ transport (v 4.3.2) to use client certificate authentication instead of user/password based but getting authentication error (Error: RabbitMQ.Client.Exceptions.BrokerUnreachableException: None of the specified endpoints were reachable ---> RabbitMQ.Client.Exceptions.AuthenticationFailureException: ACCESS_REFUSED - Login was refused using authentication mechanism PLAIN. For details see the broker logfile.)

RabbitMQ instance is configured to support TLS and client certificates auth and it's working fine when RabbitMQ.Client library is used directly as in the next [example](https://github.com/dagapitov/nsb-certificate-auth/blob/master/RabbitmqAuthTest/Program.cs).

Similar configuration with NServiceBus library doesn't work as expected (see https://github.com/dagapitov/nsb-certificate-auth/blob/master/NsbRabbitmqAuthTest/Program.cs).

It looks like NSB rabbitmq transport always use 'Plain' authentication mechanism and I cannot find a way to override this behaviour for ConnectionFactory (similar to RabbitMQ.Client - https://github.com/dagapitov/nsb-certificate-auth/blob/master/RabbitmqAuthTest/Program.cs#L24) 

## Plan of attack

- [x] make changes - #430
- [ ] update docs - https://github.com/Particular/docs.particular.net/pull/3218

bording:
@dagapitov Thanks for reporting this! You're correct, looks like the RabbitMQ client [doesn't include the EXTERNAL auth mechanism option by default](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L147...L148), so we need to add it when we create our connection factory.

We should be able to add this and a get a fix out for you soon!

dagapitov:
@bording Thank you. Do you have any timeline when this fix will be available?

bording:
@dagapitov There's no specific ETA yet. We have already have a 6.3.3 hotfix that should be going out tomorrow, so this fix would likely be in a 4.3.4 release.

adamralph:
We came to the conclusion that this is not a bug, but rather the lack of a feature, i.e. allowing the use of external auth mechanisms. Re-labelling and re-titling.

bording:
[...]

Issue #380, opened 2017-05-15
Add a setting to disable server TLS certificate validation checks


bording:
Per https://github.com/Particular/NServiceBus/issues/4705, the RabbitMQ client does support a way to disable server certificate validation checks when connecting via TLS, but we don't offer a way to configure it.

Is this something we should support?

## Plan of attack

- [x] make changes - #430
- [x] update docs - https://github.com/Particular/docs.particular.net/pull/3218

adamralph:
I'm not opposed to it, if the use case mentioned there is valid, and this will unblock it.

bording:
This will be in the 4.4.0 release.

bording:
@AnatoliiAPFM 4.4.0 has been released and here is the documentation on the new setting that disables remote certificate validation: https://docs.particular.net/transports/rabbitmq/connection-settings#transport-layer-security-support-remote-certificate-validation

PR #415, opened 2017-08-11
Add new certificate-related settings

bording:
This adds the settings to implement #407 and #380 as we've been discussing.

We'll still need to decide how to best handle shipping these in a 4.x minor instead of only having them in 5.x.


src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -42,6 +42,11 @@ public ConnectionFactory(ConnectionConfiguration connectionConfiguration, X509Ce
             connectionFactory.Ssl.Version = SslProtocols.Tls12;
             connectionFactory.Ssl.Enabled = connectionConfiguration.UseTls;
 
+            if (useExternalAuthMechanism)
+            {
+                connectionFactory.AuthMechanisms = new[] { new ExternalMechanismFactory() };
```
adamralph: By default, `AuthMechanisms` contains a single `RabbitMQ.Client.PlainMechanismFactory`. Should we be adding to the collection here, rather than replacing it?

bording: It needs to replace it because the client will choose the first one in the collection that matches any of the auth mechanisms presented by the server.

SO, if the server is configured to allow both PLAIN and EXTERNAL, then if we just added to the collection, we'd always choose PLAIN, making it impossible to use the newly added EXTERNAL option.

adamralph: Thanks @bording. That's what I suspected (as did @WojcikMike) but I wanted to get confirmation. If you rebase this, I'll be happy approve and merge, assuming we're happy for this feature to be added during the beta phase (I don't see why not).

bording:
Rebased

PR #417, opened 2017-08-30
[WIP] switch off test parallelisation

adamralph:


adamralph:
Marking as WIP since the build is still failing.

adamralph:
Withdrawing. The test are already not running in parallel. The problem is being caused by a combination of:

- The upgrade to NSB 7.x has resulted in messages being left behind on the broker after test runs.
- The failure of the broker clean up task since the switch to new build agents (now fixed).

PR #418, opened 2017-08-31
update dependency version ranges to match guidelines

adamralph:
- https://github.com/Particular/NServiceBus/blob/153026de5f624d90ce432d368fce8fedf08ff964/guidelines/coding-guidelines.md#runtime-dependencies

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
+++ b/src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
@@ -13,18 +13,18 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Fody" Version="2.1.2" PrivateAssets="All" />
-    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" PrivateAssets="All" />
-    <PackageReference Include="Janitor.Fody" Version="1.5.1" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="7.0.0-alpharelease0039" PrivateAssets="None" />
-    <PackageReference Include="Obsolete.Fody" Version="4.2.4" PrivateAssets="All" />
-    <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -17,18 +17,22 @@
...
 
   <ItemGroup>
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.2" PrivateAssets="All" />
-    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.1.2" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="[2.*,3)" PrivateAssets="All" />
+    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="[2.*,3)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="[4.4.0,5)" />
```
adamralph: I had to add this, since NSB alpha0039 took this dependency, but the latest beta doesn't. However, we still need it, because of this: https://github.com/Particular/NServiceBus.RabbitMQ/blob/7f513552334b8388b914b8a20d4da6bf3cdbe920/src/NServiceBus.Transport.RabbitMQ/Configuration/ObsoleteAppSettings.cs#L12

I'm not sure if we want to take this dependency, or if we should look at a different way of retrieving those obsolete app settings for the purpose of throwing those exceptions. Moreover, do we even need to keep doing that? Perhaps we can even just remove those checks for 5.0?

bording: @adamralph See https://github.com/Particular/NServiceBus.RabbitMQ/pull/413 where I've already removed all of that.

adamralph: Thanks @bording: let's get that one merged in first. Marking this as WIP until then.

adamralph:
This should be good to go now.

PR #420, opened 2017-09-04
fix exchange name convention parameter

### LINKED ISSUE #241: Exchange name convention for direct routing topology is incorrectly typed
The only time the user specified exchange name convention is invoked, it is passed `null` for both the `address` and `eventType` parameters - https://github.com/Particular/NServiceBus.RabbitMQ/blob/fc98f523d4cd1379acb0d5dc6b3a8da0dd9bc077/src/NServiceBus.RabbitMQ/Routing/DirectRoutingTopology.cs#L48

The proposed solution is to either remove the exchange name convention parameter, or make it a string - https://github.com/Particular/NServiceBus.RabbitMQ/blob/fc98f523d4cd1379acb0d5dc6b3a8da0dd9bc077/src/NServiceBus.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs#L19

## Plan of attack

- [x] Make change - #420
- [x] Update docs, including the upgrade guide - https://github.com/Particular/docs.particular.net/pull/3256

adamralph:
fixes #241 

Initially, I was wondering why didn't propose obsoleting the current method overload, until I realised it's not possible, since the parameter is optional. I.e. most calls to the method are simply `UseDirectRoutingTopology()`, so a deprecated overload would result in a compiler error due to an ambiguous method call. Unless we change the name of the method, our option is a direct break, without deprecation. Given that we are deprecating the direct routing topology entirely in 5.0.0, a direct break is a better option than a method rename.


WojcikMike:
I like the direct break approach

Issue #389, opened 2017-05-31
Change remaining items to match naming conventions


bording:
#384 adjusted as much as possible to follow the naming conventions, but there are some things that couldn't be changed without introducing breaks of some kind. We should consider adjusting these as well when we can.

## Plan of attack

- [x] Assembly name - #424
- [x] Product and Title assembly attributes - #424
- [x] Solution name - #424
- ~Repo name~ - #433
- ~CI project name~ _- has knock on effects due to the build templates, Octopus,etc., so not worth it_
- ~NuGet package ID~ _- cost outweighs benefit, use convention for new packages only_

adamralph:
FYI I changed the CI project name, which meant I had to override the inherited GitUrl parameters to point to the "wrongly" named repo, and also, for the regular build (the old non-.NET Core build which is paused), I had to override the SolutionPath parameter to point the "wrongly" named NServiceBus.RabbitMQ.sln.

adamralph:
@Particular/rabbitmq-transport-maintainers I've crossed out the "NuGet package ID" item since it seems it's best to keep it as-is, and only use the convention for new packages.

bording:
Is there anything else here that we actually want to change? I see Repo name is still unchecked, though I'm wondering if changing that will also have impacts on CI that aren't worth it?

adamralph:
As discussed, we'll leave the repo name as-is for now. Raised #433 for that.

PR #422, opened 2017-09-05
Renaming

adamralph:
Connects to #389


adamralph:
I'm killing this PR in favour of a new one, since this one seems to be stuck with a build status with an old build which couldn't locate the solution, and that build is since been removed.

PR #423, opened 2017-09-05
Renaming

adamralph:
Connects to #389


adamralph:
I don't believe it. This PR is suffering from the same fate as #422. Going to push to a new branch entirely.

PR #430, opened 2017-09-12
Release 4.4.0

bording:


src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -63,7 +63,9 @@ public void TearDown()
 
         protected bool TryWaitForMessageReceipt()
         {
+            // ReSharper disable NotAccessedVariable
             IncomingMessage message;
+            // ReSharper restore NotAccessedVariable
```
adamralph: Huh? Why is R# reporting this? We need it for the `out` parameter. Is it because R# thinks we should be using an `out var`, and we don't have our language version set in the `csproj` file? If so, should we set the language version instead?

bording: I'm not sure, but I think it's because there is a newer inspections version on TC now. I had originally thought that we were still targeting C# 6 on master, but it looks like it's already C# 7. I'll see if the error goes away by using a discard instead.

src/NServiceBus.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -61,11 +61,7 @@ public void TearDown()
             channelProvider?.Dispose();
         }
 
-        protected bool TryWaitForMessageReceipt()
-        {
-            IncomingMessage message;
-            return TryReceiveMessage(out message, incomingMessageTimeout);
-        }
+        protected bool TryWaitForMessageReceipt() => TryReceiveMessage(out _, incomingMessageTimeout);
```
adamralph: So much nicer!

PR #443, opened 2017-10-04
Decouple message and exchange/queue durability

### LINKED ISSUE #167: Decouple queue/exchange durabilty and message durability
Currently, we use `NServiceBus.DurableMessagesConfig.DurableMessagesEnabled` to control the durability of:

* messages
* exchange
* queues

This is somewhat [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

bording:
Connects to #167


src/NServiceBus.Transport.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -154,5 +154,17 @@ public static TransportExtensions<RabbitMQTransport> UseExternalAuthMechanism(th
...
+
+        /// <summary>
+        /// Specifies whether exchanges and queues should be declared as durable or not.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        /// <param name="useDurableExchangesAndQueues">Specifies whether exchanges and queues should be declared as durable or not.</param>
+        /// <returns></returns>
+        public static TransportExtensions<RabbitMQTransport> UseDurableExchangesAndQueues(this TransportExtensions<RabbitMQTransport> transportExtensions, bool useDurableExchangesAndQueues)
+        {
+            transportExtensions.GetSettings().Set(SettingsKeys.UseExternalAuthMechanism, useDurableExchangesAndQueues);
```
adamralph: `UseExternalAuthMechanism` should be `UseDurableExchangesAndQueues`.

bording: Ah, copypasta error. Will fix!

bording: I've pushed a fixup commit for this, which I'll squash before merging.

bording:
> We have the new setting, and the throwing of the exception, but don't we actually need to do something to use that new setting to control the durability of the exchanges and queues separately from the durability of the messages?

@adamralph I'm not sure I understand this comment. The change I've made [here](https://github.com/Particular/NServiceBus.RabbitMQ/pull/443/files#diff-179646131f3010c16f5abb4c584dfb24R130) means that we're using the value from the new setting to control durability.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -122,13 +122,23 @@ public override Task Stop()
...
         IRoutingTopology CreateRoutingTopology()
         {
-            var durable = settings.DurableMessagesEnabled();
-
             if (!settings.TryGet(out Func<bool, IRoutingTopology> topologyFactory))
             {
                 throw new InvalidOperationException("A routing topology factory has not been configured. Configure a routing topology with a TransportExtensions<RabbitMQTransport>.UseXXXRoutingTopology() method.");
             }
 
+            if (!settings.TryGet(SettingsKeys.UseDurableExchangesAndQueues, out bool durable))
```
adamralph: > > We have the new setting, and the throwing of the exception, but don't we actually need to do something to use that new setting to control the durability of the exchanges and queues separately from the durability of the messages?
> 
> @adamralph I'm not sure I understand this comment. The change I've made [here](https://github.com/Particular/NServiceBus.RabbitMQ/pull/443/files#diff-179646131f3010c16f5abb4c584dfb24R130) means that we're using the value from the new setting to control durability.

Apologies, I misread things. I assumed that the `durable` variable referred to the entirety of "durability" i.e. messages, exchanges and queues, and forgot that core is taking care of message durability, and the transport only controls exchange and queue durability.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -122,13 +122,23 @@ public override Task Stop()
...
+            {
+                if (settings.DurableMessagesEnabled())
+                {
+                    durable = true;
+                }
+                else
+                {
+                    throw new Exception("When durable messages are disabled, 'EndpointConfiguration.UseTransport<RabbitMQTransport>().UseDurableExchangesAndQueues()' needs to be called to specify exchange and queue durability settings.");
```
adamralph: > needs to be called

I think I prefer:

> must also be called

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs
```diff
@@ -122,13 +122,23 @@ public override Task Stop()
...
+                }
+            }
+
```
adamralph: For some reason (it could be just me) I'm still finding this a little difficult to read. I think it's the name `durable` and polarity of the inner conditional. How would this look instead?

```C#
if (!settings.TryGet(SettingsKeys.UseDurableExchangesAndQueues, out bool useDurableExchangesAndQueues))
{
    if (!settings.DurableMessagesEnabled())
    {
        throw new Exception(...);        
    }
    
    useDurableExchangesAndQueues = true;
}
```

bording: I wrote it originally to avoid the inner `!`, but I can see how "if not enabled, throw" might be clearer. I'll make the change.

bording:
@adamralph More fixup commits to address your feedback.

bording:
@adamralph Done!

Issue #427, opened 2017-09-05
Require topology choice


adamralph:
This will allow future deprecation of the conventional routing topology, which is the default topology.

## Plan of attack

- [x] make changes - #428
- [x] update docs
  - [x] https://github.com/Particular/docs.particular.net/pull/3183
  - [x] https://github.com/Particular/docs.particular.net/pull/3184

PR #447, opened 2017-10-11
Revise some exception messages

bording:
Connects to #427.


adamralph:
FYI I connected this with #427 and removed it from the milestone, since it is fixing the exception message which was introduced in the implementation of that issue.

That means that the addition of the full stops to the other two exception messages will not appear in our release notes, but I think we can live with that. :-)

PR #449, opened 2017-10-11
Misc cleanup

bording:
This includes a few minor changes that clean things up.

The approval file change is related to ensuring the approval file can correctly be found when/if TC schedules a test run on a separate build agent from the one that compiled the code.


src/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
```
adamralph: This isn't a clean up, but rather it allows the latest C# minor and instead of the latest major, but I think it's OK to be part of this PR.

src/NServiceBus.Transport.RabbitMQ.Tests/TestApprover.cs
```diff
@@ -0,0 +1,33 @@
+﻿#if NET452
+namespace NServiceBus.Transport.RabbitMQ.Tests
+{
+    using System.IO;
+    using ApprovalTests;
+    using ApprovalTests.Namers;
+
+    static class TestApprover
```
adamralph: Why has this class been introduced? Is it purely to move the approved API to a subfolder?

adamralph: Discussed offline: this is so that we can copy the build artifacts to a dependent build on another agent during CI. The default behaviour is walk up the call stack looking for source file info. This swaps that out for using a known file location relative to the assembly.

PR #460, opened 2017-12-22
[WIP] refactor a few things

adamralph:
Marking as WIP since this primary purpose of this PR is to discuss a couple of ideas, rather than have it merged (perhaps splitting into more commits would be better before merging).

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelPool.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelPool.cs
@@ -4,11 +4,11 @@ namespace NServiceBus.Transport.RabbitMQ
     using System.Collections.Concurrent;
     using global::RabbitMQ.Client;
 
-    sealed class ChannelProvider : IDisposable
+    sealed class ChannelPool : IDisposable
     {
-        public ChannelProvider(ConnectionFactory connectionFactory, IRoutingTopology routingTopology, bool usePublisherConfirms)
+        public ChannelPool(ConnectionFactory connectionFactory, IRoutingTopology routingTopology, bool usePublisherConfirms)
         {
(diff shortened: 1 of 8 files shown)


src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -63,7 +63,7 @@ public ConnectionFactory(ConnectionConfiguration connectionConfiguration, X509Ce
             }
         }
 
-        public IConnection CreatePublishConnection() => CreateConnection("Publish");
+        public IConnection CreateMessagesConnection() => CreateConnection("Messages");
```
bording: I don't really care for this change, because the connection name is visible in the RabbitMQ management UI.

This means we go from seeing that the connection is used for publishing, to "messages" which I think is too generic. All connections to a message broker are for messages!

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelPool.cs
```diff
@@ -4,19 +4,19 @@ namespace NServiceBus.Transport.RabbitMQ
     using System.Collections.Concurrent;
     using global::RabbitMQ.Client;
 
-    sealed class ChannelProvider : IDisposable
+    sealed class ChannelPool : IDisposable
```
bording: The channels retrieved from this class are only for sending messages to the broker via the `BasicPublish` API call. These changes lose that distinction.

I think we need to retain that in some way.

src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
adamralph: What about the "Administration" connection?

adamralph:
Discussed with @bording. Closing this and may use some of the ideas in later PR's.

PR #461, opened 2018-01-03
Hotfix 4.4.1

### LINKED ISSUE #456: Lazily creating publish connection causes some weird behavior
As brought up in [Discourse](https://discuss.particular.net/t/rabbitmq-transport-brokerunreachableexception/272), the behavior around opening the publish connection means that we don't try to open the connection until the first send attempt, which overall seems like a decent optimization. If an endpoint doesn't need to send messages, then we don't open up an unused connection.

However, this does have a couple drawbacks:

- For a send-only endpoint, that means we start the endpoint without actually opening a connection to the broker at all. Any problem with the connection string or connectivity to the broker won't be discovered until the first attempt to send a message.
- [We initialize the connection lazily using `Lazy<T>`](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs#L11), so if the single attempt to open the connection throws an exception, that exception is cached and we can never open a publish connection for the lifetime of the endpoint.

### LINKED ISSUE #458: New x-first-death headers are not removed before processing an incoming message
RabbitMQ 3.7.0 has added some additional headers that get set when a message is dead lettered:

- x-first-death-exchange
- x-first-death-queue
- x-first-death-reason

We should consider adding these to the list of headers we remove in `MessageConverter.RetrieveHeaders` to prevent them being on a message if it's forwarded.

While we have some header removing in 5.x and 4.4.x, I noticed that we didn't go back to 3.5.x and add anything for this there. We should probably do that as well since a message could be sent to a 3.5.x endpoint from a newer endpoint and it could have these headers on it.

### LINKED ISSUE #463: Minimum RabbitMQ.Client version throws NullReferenceException in HeartbeatReadTimerCallback
https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/257

bording:
Fixes #458 
Fixes #456 
Fixes #463

### DIFF
--- a/packaging/nuget/NServiceBus.RabbitMQ.nuspec
+++ b/packaging/nuget/NServiceBus.RabbitMQ.nuspec
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[6.0.0, 7.0.0)" />
-      <dependency id="RabbitMQ.Client" version="[4.1.0, 5.1.0)" />
+      <dependency id="RabbitMQ.Client" version="[5.0.1, 5.1.0)" />
     </dependencies>
   </metadata>
   <files>
--- a/src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
@@ -8,12 +8,16 @@ namespace NServiceBus.Transport.RabbitMQ
     {
         public ChannelProvider(ConnectionFactory connectionFactory, IRoutingTopology routingTopology, bool usePublisherConfirms)
         {
-            connection = new Lazy<IConnection>(() => connectionFactory.CreatePublishConnection());
-
+            this.connectionFactory = connectionFactory;
             this.routingTopology = routingTopology;
             this.usePublisherConfirms = usePublisherConfirms;
 
             channels = new ConcurrentQueue<ConfirmsAwareChannel>();
+        }
+
+        public void CreateConnection()
+        {
+            connection = connectionFactory.CreatePublishConnection();
         }
 
         public ConfirmsAwareChannel GetPublishChannel()
@@ -24,7 +28,7 @@ namespace NServiceBus.Transport.RabbitMQ
             {
                 channel?.Dispose();
 
-                channel = new ConfirmsAwareChannel(connection.Value, routingTopology, usePublisherConfirms);
+                channel = new ConfirmsAwareChannel(connection, routingTopology, usePublisherConfirms);
             }
 
             return channel;
@@ -49,9 +53,9 @@ namespace NServiceBus.Transport.RabbitMQ
 
         void DisposeManaged()
         {
-            if (connection.IsValueCreated)
+            if (connection != null)
             {
-                connection.Value.Dispose();
+                connection.Dispose();
             }
 
             foreach (var channel in channels)
@@ -60,9 +64,10 @@ namespace NServiceBus.Transport.RabbitMQ
             }
         }
 
-        readonly Lazy<IConnection> connection;
+        readonly ConnectionFactory connectionFactory;
         readonly IRoutingTopology routingTopology;
         readonly bool usePublisherConfirms;
         readonly ConcurrentQueue<ConfirmsAwareChannel> channels;
+        IConnection connection;
     }
 }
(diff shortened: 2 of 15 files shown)


packaging/nuget/NServiceBus.RabbitMQ.nuspec
```diff
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[6.0.0, 7.0.0)" />
-      <dependency id="RabbitMQ.Client" version="[4.1.0, 5.1.0)" />
+      <dependency id="RabbitMQ.Client" version="[5.0.1, 5.1.0)" />
```
bording: I went ahead and bumped the client up to 5.0.1. I figured it made sense to go ahead and do it as part of the hotfix since we know there's the bug in 4.1.0. What do you think? I can always remove the commit if we don't want this.

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -16,7 +16,10 @@ static class DelayInfrastructure
 
         public const int MaxDelayInSeconds = (1 << maxNumberOfBitsToUse) - 1;
         public const string DelayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
-        public const string DeadLetteredMessageHeader = "x-death";
+        public const string DeadLetteredMessageHeader1 = "x-death";
+        public const string DeadLetteredMessageHeader2 = "x-first-death-exchange";
+        public const string DeadLetteredMessageHeader3 = "x-first-death-queue";
+        public const string DeadLetteredMessageHeader4 = "x-first-death-reason";
```
bording: I'm not really happy with these names, but I wasn't sure what to use instead. Maybe go with `DeathHeader`, `FirstDeathExchangeHeader`, etc?

packaging/nuget/NServiceBus.RabbitMQ.nuspec
adamralph: @bording I think it makes sense to have this in the patch, but should we have a "bug" to link to in the release notes, since it's unrelated to both the current bugs that we have raised?

src/NServiceBus.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
adamralph: How about we use the same names, and  just convert the `kekab-case` to `PascalCase`, with the `Header` suffix? I.e. `XDeathHeader`, etc.

src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -8,14 +8,18 @@ class ChannelProvider : IChannelProvider, IDisposable
...
-            connection = new Lazy<IConnection>(() => connectionFactory.CreatePublishConnection());
-
+            this.connectionFactory = connectionFactory;
             this.routingTopology = routingTopology;
             this.usePublisherConfirms = usePublisherConfirms;
 
             channels = new ConcurrentQueue<ConfirmsAwareChannel>();
         }
 
+        public void CreateConnection()
```
adamralph: Something doesn't smell quite right about this, but I'm not sure what it is yet. However, I think it's good enough for the patch. I may get some crazy refactoring idea for `develop` later. 😉

bording:
@adamralph I pushed up a fixup commit for the header names. I'll squash it once you've looked it over.

src/NServiceBus.RabbitMQ/Connection/ChannelProvider.cs
bording: I'll be interested to see what you might come up with. We need some way to start the connection at endpoint start, but not sooner. The other places that get a `ConnectionFactory` all get called at endpoint start by core, so that's when those connections are opened.

adamralph:
@bording the test failures on the build server are genuine. A null connection is being passed to the `ConfirmsAwareChannel` constructor because `ChannelProvider.CreateConnection()` is not being called in the context of the tests.

src/NServiceBus.Transport.RabbitMQ.TransportTests/App_Packages/NSB.TransportTests.6.1.1/NServiceBusTransportTest.cs
```diff
@@ -86,6 +86,8 @@ public void TearDown()
             ReceiveInfrastructure = TransportInfrastructure.ConfigureReceiveInfrastructure();
             SendInfrastructure = TransportInfrastructure.ConfigureSendInfrastructure();
 
+            await TransportInfrastructure.Start();
```
adamralph: Are we going to have the same problem on `develop`, but there we can't change the source files?

bording: Yes, but I've opened https://github.com/Particular/NServiceBus/issues/5099 to fix this in core. We'll have to work around it differently in the mean time.

bording:
@adamralph This should be ready to merge now.

PR #462, opened 2018-01-08
Hotfix 3.5.2

### LINKED ISSUE #464: Minimum RabbitMQ.Client version throws NullReferenceException in HeartbeatReadTimerCallback
https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/257

### LINKED ISSUE #465: Headers are not removed before processing an incoming message

bording:
This is bringing the changes from 4.4.1 over to the support-3.5 branch.

The publish connection isn't created via `Lazy`, so that change isn't needed here.

The unused clustering tests were causing inspection errors, so I removed them.  
 
Fixes #464
Fixes #465


src/NServiceBus.RabbitMQ/MessageConverter.cs
```diff
@@ -118,5 +126,12 @@ static string ValueToString(object value)
 
 
         static ILog Logger = LogManager.GetLogger(typeof(MessageConverter));
+
+        const string delayHeader = "NServiceBus.Transport.RabbitMQ.DelayInSeconds";
+        const string xDeathHeader = "x-death";
+        const string xFirstDeathExchangeHeader = "x-first-death-exchange";
+        const string xFirstDeathQueueHeader = "x-first-death-queue";
+        const string xFirstDeathReasonHeader = "x-first-death-reason";
+        const string confirmationIdHeader = "NServiceBus.Transport.RabbitMQ.ConfirmationId";
```
adamralph: In https://github.com/Particular/NServiceBus.RabbitMQ/pull/461 the variables start with an upper case character. Is there a reason that a different convention is being used here?

adamralph: As discussed, this is because they are private fields here, but public fields in the other case.

PR #473, opened 2018-02-01
Run API Approval test on both TFMs

bording:
This lets the API approvals test run against the `netstandard2.0` assembly in addition to the `net452` assembly.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalTestConfig.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalTestConfig.cs
@@ -1,8 +1,6 @@
-﻿#if NET452
 using ApprovalTests.Reporters;
-#if DEBUG
+#if NET452
 [assembly: UseReporter(typeof(DiffReporter), typeof(AllFailingTestsClipboardReporter))]
 #else
-[assembly: UseReporter(typeof(DiffReporter))]
-#endif
+[assembly: UseReporter(typeof(QuietReporter))]
 #endif
--- a/src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
+++ b/src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
@@ -12,16 +12,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="ApprovalTests" Version="3.0.13" NoWarn="NU1701" />
+    <PackageReference Include="ApprovalUtilities" Version="3.0.13" NoWarn="NU1701" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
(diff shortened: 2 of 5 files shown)


src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalTestConfig.cs
```diff
@@ -1,8 +1,6 @@
-﻿#if NET452
-using ApprovalTests.Reporters;
-#if DEBUG
+﻿using ApprovalTests.Reporters;
+#if NET452
 [assembly: UseReporter(typeof(DiffReporter), typeof(AllFailingTestsClipboardReporter))]
```
adamralph: This config is currently only used under Debug for NET452. With this change we will now use it for both Debug and Release. Does that matter? (I've no idea what it does.)

src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
```diff
@@ -12,16 +12,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="ApprovalTests" Version="3.0.13" NoWarn="NU1701" />
+    <PackageReference Include="ApprovalUtilities" Version="3.0.13" NoWarn="NU1701" />
```
adamralph: Why has this been added?

src/NServiceBus.Transport.RabbitMQ.Tests/TestApprover.cs
```diff
@@ -16,18 +16,7 @@ public static void Verify(string text)
...
-                var assemblyPath = GetType().Assembly.Location;
-                var assemblyDir = Path.GetDirectoryName(assemblyPath);
-
-                sourcePath = Path.Combine(assemblyDir, "..", "..", "..", "ApprovalFiles");
-            }
-
-            public override string SourcePath => sourcePath;
-
-            readonly string sourcePath;
+            public override string SourcePath { get; } = Path.Combine(TestContext.CurrentContext.TestDirectory, "..", "..", "..", "ApprovalFiles");
```
adamralph: I was going to drop the RHS down to the next line in my fixup, but I was sure you'd disagree. 😉

src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
adamralph: As discussed, `NU1701` is not transitive.

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalTestConfig.cs
adamralph: As discussed, this is fine, but after further review we had to swap the .NET Core lib to `QuietReporter`.

PR #482, opened 2018-08-09
[WIP] Fix leaking headers

timbussmann:
Adding basic transport tests reproducing the issue of leaking headers.

Need to see whether this also fails on the build server due to weird RabbitMQ client exceptions once the fix is applied.

Todo:
* [ ] rebase against a release/hotfix branch

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
@@ -1,0 +1,69 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.TransportTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NServiceBus.TransportTests;
+    using NUnit.Framework;
+
+    public class When_modifying_incoming_headers : NServiceBusTransportTest
+    {
+        [Test]
+        public async Task Should_roll_back_header_modifications_for_immediate_retries()
+        {
+            var messageRetries = new TaskCompletionSource<MessageContext>();
+            var firstInvocation = true;
+
+            await StartPump(context =>
+                {
+                    if (firstInvocation)
+                    {
+                        context.Headers["test-header"] = "modified";
+                        firstInvocation = false;
+                        throw new Exception();
+                    }
+
+                    messageRetries.SetResult(context);
+                    return Task.FromResult(0);
+                },
+                context => Task.FromResult(ErrorHandleResult.RetryRequired),
+                TransportTransactionMode.None);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"test-header", "original"}
+            });
+
+            var retriedMessage = await messageRetries.Task;
+
+            Assert.AreEqual("original", retriedMessage.Headers["test-header"]);
+        }
+
+        [Test]
+        public async Task Should_roll_back_header_modifications_when_handling_error()
+        {
+            var errorHandled = new TaskCompletionSource<ErrorContext>();
+
+            await StartPump(context =>
+                {
+                    context.Headers["test-header"] = "modified";
+                    throw new Exception();
+                },
+                context =>
+                {
+                    errorHandled.SetResult(context);
+                    return Task.FromResult(ErrorHandleResult.Handled);
+                },
+                TransportTransactionMode.None);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"test-header", "original"}
+            });
+
+            var errorContext = await errorHandled.Task;
+
+            Assert.AreEqual("original", errorContext.Message.Headers["test-header"]);
+        }
+    }
+}
--- a/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
@@ -40,7 +40,7 @@
         }
 
         [Test]
-        public async Task Should_roll_back_header_modifications_when_handling_error()
+        public async Task Should_roll_back_header_modifications_before_handling_error()
         {
             var errorHandled = new TaskCompletionSource<ErrorContext>();
 
@@ -64,6 +64,40 @@
             var errorContext = await errorHandled.Task;
 
             Assert.AreEqual("original", errorContext.Message.Headers["test-header"]);
+        }
+
+        [Test]
+        public async Task Should_roll_back_header_modifications_from_handling_error()
+        {
+            var messageRetries = new TaskCompletionSource<MessageContext>();
+            var firstInvocation = true;
+
+            await StartPump(context =>
+                {
+                    if (firstInvocation)
+                    {
+                        firstInvocation = false;
+                        throw new Exception();
+                    }
+
+                    messageRetries.SetResult(context);
+                    return Task.FromResult(0);
+                },
+                context =>
+                {
+                    context.Message.Headers["test-header"] = "modified";
+                    return Task.FromResult(ErrorHandleResult.RetryRequired);
+                },
+                TransportTransactionMode.None);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"test-header", "original"}
+            });
+
+            var retriedMessage = await messageRetries.Task;
+
+            Assert.AreEqual("original", retriedMessage.Headers["test-header"]);
         }
     }
 }
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -244,7 +244,8 @@
                 {
                     try
                     {
-                        var messageContext = new MessageContext(messageId, headers, message.Body ?? new byte[0], transportTransaction, tokenSource, contextBag);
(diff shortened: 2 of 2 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -244,7 +244,8 @@ async Task Process(BasicDeliverEventArgs message)
                 {
                     try
                     {
-                        var messageContext = new MessageContext(messageId, headers, message.Body ?? new byte[0], transportTransaction, tokenSource, contextBag);
+                        var msgHeaders = new Dictionary<string, string>(headers);
```
bording: This introduces an unneeded dictionary copy/allocations on the main message processing path. Instead consider creating a new header dictionary by calling `messageConverter.RetrieveHeaders(message)` only when an error has actually occurred and potential modifications to the original collection could be observed.

timbussmann: This was just a basic test to see whether it passes the tests, I'll look into more efficient approaches. I wanted to commit these changes because this exact change caused the test to fail locally due to weird exceptions: `RabbitMQ.Client.Exceptions.BrokerUnreachableException : None of the specified endpoints were reachable` and I wanted to see whether it happens on the build server too.

timbussmann:
I've changed the code to recreate the message headers via `messageConverter.RetrieveHeaders(message)` when dealing with exceptions. This could probably be slightly further optimized by doing some actual content verification to detect whether the headers have been changed. That approach could avoid allocations in case of no header modifications but would take about the same amount of time to run. Thoughts @Particular/rabbitmq-transport-maintainers ?

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
```diff
@@ -0,0 +1,69 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.TransportTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NServiceBus.TransportTests;
+    using NUnit.Framework;
+
+    public class When_modifying_incoming_headers : NServiceBusTransportTest
```
timbussmann: @Particular/nservicebus-maintainers should we also add those tests to the core transport tests?

andreasohlund: Makes sense. That said it would mean that some transports that doesn't honor this will break when we update them to core 7.1 but I guess it's good to know and we can always disable the test for that specific transport?

timbussmann: > but I guess it's good to know and we can always disable the test for that specific transport?

that was my idea as well. Once this is fixed in core, we can get rid of those tests again easily.

andreasohlund: :shipit:

adamralph:
> This could maybe be further optimized by trying to check for modified headers before recreating, not sure whether it makes sense to invest more time finding a solution to this?

AFAIK the only way to do that would be to keep a copy of the original dictionary, in order to do a comparison, but then we may as well use that original collection, and we're back to the problem of the extra allocation.

I can't think of a better way than the current implementation.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -251,6 +251,7 @@ async Task Process(BasicDeliverEventArgs message)
                     catch (Exception ex)
                     {
                         ++numberOfDeliveryAttempts;
+                        headers = messageConverter.RetrieveHeaders(message);
                         var errorContext = new ErrorContext(ex, headers, messageId, message.Body ?? new byte[0], transportTransaction, numberOfDeliveryAttempts);
                         errorHandled = await onError(errorContext).ConfigureAwait(false) == ErrorHandleResult.Handled;
```
bording: Do we also need to protect against potential modification from the `onError` call?

timbussmann: if somebody uses a custom recoverability policy, this could potentially be an issue. I'm not sure our own recoverability logic could leak anything but I'll double check.

timbussmann: by default, there are no header modifications going on in the error handling logic for immediate retries. However, if you write your custom recoverability policy. Also, code subscribing to the `MessageToBeRetried` event in the event aggregator has full access to the context, including headers. This event is only available internally in core at the moment (there were ideas to make them public though). Thoughts?

adamralph: Would I be correct in thinking that the only way to guard against this would be to read the headers immediately before the call to `onMessage(messageContext)`, as well as before every call to `onError(errorContext)` as we are doing here?

timbussmann: @adamralph but core is only accepting `Dictionary<string, string>` so there is no elegant way to work around that I think?

adamralph: @timbussmann I guess we could derive from `Dictionary` and instiate that type in the header retrieval. We'd pass that object to `onMessage` and `onError`. It could track whether any of the values for existing keys are reassigned, or whether any keys are added or removed. Back in the transport code we could ask the object to give us the original dictionary, which would be just itself if unmodified or a new object otherwise. Although I suspect it may be difficult to write the internals of this new type without resulting in more, often unecessary, allocations on the hotpath. 🤷‍♂️

bording: It would make things messier, but we could move the initial `RetrieveHeaders` and `RetrieveMessageId` code inside the while loop, right before the `onMessage` call.

That way we always get a new header collection, but we aren't allocating an additional collection in the case of a message that is processed on the first try.

adamralph: @bording yeah, that's what I meant up here: https://github.com/Particular/NServiceBus.RabbitMQ/pull/482#discussion_r209936999

It means paying that cost for every call to `onMessage` and `onError`, but perhaps it's a necessary evil.

bording: What cost? How would that be different from how it currently is?

bording:
@adamralph and I had a conversation, and we realized we were talking about the same thing. We decided to go ahead and push up a change where the initial retrieve calls were inside the while loop. We were thinking we could extract the common logic into a method and then just call that right before `onMessage` and `onError`.

However, when I started looking at doing that, it was more complicated than we were thinking it would be because of needing to return a header collection, message ID, and a bool to determine if we need to exit early... but the method would also need to be async.

Since we can't use tuples yet, I decided to push up a different change.

The initial calls with all of the protection for poison messages are left untouched. Instead, if it looks like we'll need another loop iteration because `onError` returned `ErrorHandleResult.RetryRequired`, I'm now going ahead and creating the header collection again.

This should ensure that both `onMessage` and `onError` can't observe changes to the collections without adding additional allocations in the case where the message is processed on the first attempt.

bording:
While I went ahead and approved this, I'm wondering if we should have another test that would fail if `onError` isn't guarded?

adamralph:
> I'm wondering if we should have another test that would fail if `onError` isn't guarded

I find the tests difficult to read, but isn't the second one, `Should_roll_back_header_modifications_when_handling_error`, doing that?

bording:
> I find the tests difficult to read, but isn't the second one, Should_roll_back_header_modifications_when_handling_error, doing that?

If it was, it should have been failing until I pushed my change, right?

adamralph:
@bording yeah, I guess it should have. I'm not sure what that test is doing then.

timbussmann:
> I'm not sure what that test is doing then.

the tests I've added verify that:
* when a message is retried immediately, the previous changes to headers from the pipeline are not visible to the second processing attempt
* when a message is failing, header modifications from the pipeline are not visible to the `onError` handler, as onError can potentially create delayed retries or error messages using this header.

Based on my comment here: https://github.com/Particular/NServiceBus.RabbitMQ/pull/482#discussion_r209899753 I did not add the logic which @bording added afterwards (as the default implementations do not modify headers on immediate retries). If you want to protect the headers from being modified due to custom recoverability policies, Brandon's changes make sense, they aren't required otherwise though. I'll leave that for the rabbit maintainers to decide.

bording:
@adamralph @WojcikMike What do you think?

adamralph:
@bording @timbussmann I believe we should protect against header modification in `onError`, so the change is good as-is, and that we should add the extra test that @bording suggests.

timbussmann:
I've added a test case to cover the scenario covered by @bording changes. It still feels like a very extreme edge case to handle (modifying headers in the recoverability policy in case the policy returns an immediate retry) but I'm fine with handling this. Should be good now.

timbussmann:
also changed the target branch to hotfix-5.0.1

WojcikMike:
:+1: for 1 commit

bording:
FYI, I talked with @timbussmann about the fact that needing to make these changes aren't just scoped to RabbitMQ, but some of these tests will fail on all/most transports.

Let's hold off on doing anything with this PR until we decide how to handle this.

Issue #484, opened 2018-08-31
Endpoint getting stuck with "System.Exception: Failed to add " in the logs


dasjestyr:
## Problem
I'm having an issue that is causing all of my services to halt and no longer process messages. I'm not seeing any application errors, but in the logs of a couple different services (consumers), I'm getting logs like this:

> System.Exception: Failed to add 3 at NServiceBus.Transport.RabbitMQ.ConfirmsAwareChannel.GetConfirmationTask() at NServiceBus.Transport.RabbitMQ.ConfirmsAwareChannel.SendMessage(String address, OutgoingMessage message, IBasicProperties properties) at NServiceBus.Transport.RabbitMQ.MessageDispatcher.SendMessage(UnicastTransportOperation transportOperation, ConfirmsAwareChannel channel) at NServiceBus.Transport.RabbitMQ.MessageDispatcher.Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, ContextBag context) at NServiceBus.DelayedRetryExecutor.Retry(IncomingMessage message, TimeSpan delay, TransportTransaction transportTransaction) at NServiceBus.RecoverabilityExecutor.DeferMessage(DelayedRetry action, ErrorContext errorContext) at NServiceBus.Transport.RabbitMQ.MessagePump.Process(BasicDeliverEventArgs message) at NServiceBus.Transport.RabbitMQ.MessagePump.Consumer_Received(Object sender, BasicDeliverEventArgs eventArgs)

[...]

PR #487, opened 2018-09-06
[WIP] Disable auto-recovery

tmasternak:
## Overview
This PR is a proposal of solution to solve #484. I disables auto-recovery.

## Rationale
With my current understanding of the code base we use the auto-recoverability only during sends in the `MessagePump` when dealing with poison messages or in `MessageDispatcher`. The only scenario that comes to my mind when we take advantage of this feature is when channels stored in the `ChannelFactory` pool get closed and than recovered both happening when those are not used. Otherwise those channels would be removed from the pool either when picked from or returned to the pool. 

As a result I propose to turn auto-recoverability off. We might get some penalty in terms of performance (which I doubt is the case) however the code will become easier to understand and reason about.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
@@ -31,7 +31,6 @@
                 UserName = connectionConfiguration.UserName,
                 Password = connectionConfiguration.Password,
                 RequestedHeartbeat = connectionConfiguration.RequestedHeartbeat,
-                AutomaticRecoveryEnabled = true,
                 NetworkRecoveryInterval = connectionConfiguration.RetryDelay,
                 UseBackgroundThreadsForIO = true
             };


tmasternak:
BTW do we know why the sequence no [check](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Connection/ConfirmsAwareChannel.cs#L118) was put in place?

adamralph:
> BTW do we know why the sequence no [check](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Connection/ConfirmsAwareChannel.cs#L118) was put in place?

@tmasternak I suspect it's because we _want_ an exception to be thrown if that call fails even though, by design, it never should. I.e. what we really want is `ConcurrentDictionary.Add()` rather than `TryAdd()`.

bording:
@tmasternak and I had a chat about this today, and we do rely on automatic recovery in the design of the transport, so turning it off really isn't an option.

> @tmasternak I suspect it's because we want an exception to be thrown if that call fails even though, by design, it never should. I.e. what we really want is ConcurrentDictionary.Add() rather than TryAdd().

During the initial 4.x development, there was a period of time where I was hitting this scenario occasionally even though by design it shouldn't be possible. I added the exception while attempting to debug it. I was never able to come up with a reliable repro, and then it stopped happening at all. 

I thought about removing the check, but it seemed better to leave it in case the problem was still lurking in there somewhere.

PR #489, opened 2018-09-07
Use RecoverySucceeded event to clear any orphaned messages

bording:
Alternative to #488.

@tmasternak I think this will also clean up anything left in the `messages` collection that was missed by `Channel_ModelShutdown`.


tmasternak:
The initial proposal still has a race condition and violates following safetly property:

> If a channel is in the pool and is opened it can't contain an element in the messages collection that is grater than `NextSequeneceNo` of the underlying rabbit channel.

The changes that I have made fixes that and makes sure the above condition holds. The arguments that I have come from the model-based testing using TLA+.

Here are specifications of the original PR and the one after my commit. Plus the sample execution that shows problems with the first specification:

https://github.com/tmasternak/NServiceBus.ModelChecking/tree/master/RabbitMQ

bording:
Closing in favor of #494

PR #494, opened 2018-09-12
Remove ChannelProvider automatic connection recovery

bording:
Instead of trying to fight connection automatic recovery race conditions, this instead lets automatic recovery be enabled/disabled on a per-connection basis.

`ChannelProvider` has been updated to not use automatic recovery and instead creates a new connection when the existing connection has been disconnected. This means that once a connection is closed, all the channels on that connection will be closed as well, and will never be usable again.

Once a new connection is established, new channels will be created.

CC: @tmasternak

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -2,13 +2,15 @@ namespace NServiceBus.Transport.RabbitMQ
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Threading.Tasks;
     using global::RabbitMQ.Client;
 
     sealed class ChannelProvider : IDisposable
     {
-        public ChannelProvider(ConnectionFactory connectionFactory, IRoutingTopology routingTopology, bool usePublisherConfirms)
+        public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay, IRoutingTopology routingTopology, bool usePublisherConfirms)
         {
             this.connectionFactory = connectionFactory;
+            this.retryDelay = retryDelay;
 
             this.routingTopology = routingTopology;
             this.usePublisherConfirms = usePublisherConfirms;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -19,6 +21,35 @@ public ChannelProvider(ConnectionFactory connectionFactory, IRoutingTopology rou
         public void CreateConnection()
         {
             connection = connectionFactory.CreatePublishConnection();
+            connection.ConnectionShutdown += Connection_ConnectionShutdown;
```
tmasternak: We probably want to unsubscribe from this event somewhere.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -19,6 +21,35 @@ public ChannelProvider(ConnectionFactory connectionFactory, IRoutingTopology rou
...
+        }
+
+        void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
+        {
+            if (e.Initiator != ShutdownInitiator.Application)
+            {
+                Task.Run(Reconnect).Ignore();
```
tmasternak: This is risky as we allow potentially multiple `Reconnects` happening at the same time. If the newly created connection shutsdown immediately we have two threads in the `Reconnect`.

I suggest creating a never ending `Reconnect` task and make `Connection_ConnectionShutdown` set bool flag to indicate if reconnection is needed + unsubscribe from the handler (all in lock statement).

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
bording: Why do you think that is needed?

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
bording: > This is risky as we allow potentially multiple Reconnects happening at the same time. If the newly created connection shutsdown immediately we have two threads in the Reconnect.

I think we'll need to chat more about this, because I'm not seeing the problem.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
tmasternak: @bording my bad. I've confused the dependency direction between the Connection and the ChannelProvider and thought we might have a memory leak here. All good.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
bording: Discussed this with @tmasternak and the concern was around the `reconnected` value, but since that is a local variable, there should be no problem.

bording:
Sounds like this approach is going to work, but we actually need to release this is a patch, so I'm going to recreate this on a hotifx branch against master instead.

PR #495, opened 2018-09-13
Transport getting stuck with "System.Exception: Failed to add " in the logs

bording:
Solves #495 

@tmasternak We'll need to update the name of the PR, link to the relevant public issue, and get the milestones sorted out before this will be ready to release.


bording:
One thing I'm wondering, should we add some logging around the connection being lost and the reconnection attempts?

src/NServiceBus.Transport.RabbitMQ/Connection/ConfirmsAwareChannel.cs
```diff
@@ -200,13 +207,15 @@ void SetException(ulong key, string exceptionMessage)
             }
         }
 
-
         public void Dispose()
         {
+            ClearMessages();
```
bording: This change is unnecessary. Calling `Dispose` on an open channel will call `Channel_ModelShutdown` on it. I'm going to revert this.

PR #502, opened 2018-11-01
Update test packages

bording:
This includes the update to the test adapter package that finally fixes the conflict with Mono.Cecil.


adamralph:
Is the same change required on `master` and any `support-x` branches?

bording:
@adamralph Yes, we need the updated test adapter on any branches where we want to be able to run tests in VS. However, it's possible that some branches are still using an older version of Microsoft.NET.Test.Sdk that doesn't have a conflict with the old alpha test adapter. If so, they wouldn't have to be updated.

PR #505, opened 2018-12-13
Miscellaneous repo cleanup updates

bording:
Lots of minor changes that are overdue.


.gitattributes
```diff
@@ -64,11 +64,6 @@
 *.bat text eol=crlf
 
 # Custom for Visual Studio
-*.csproj merge=union
-*.vbproj merge=union
-*.fsproj merge=union
-*.dbproj merge=union 
-*.sln text eol=crlf merge=union
```
adamralph: Where did these come from in the first place?

src/NServiceBus.Transport.RabbitMQ/FodyWeavers.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Weavers>
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Weavers GenerateXsd="false">
```
adamralph: What does this do?

.gitattributes
bording: The previous repostandards file add them. Some early template from GitHub I think included them.

src/NServiceBus.Transport.RabbitMQ/FodyWeavers.xml
bording: New versions of Fody now want to generate an xml schema file on build, so that tells it not to.

PR #511, opened 2018-12-28
convert reset virtual host from script to exe

adamralph:


reset-virtual-host.cmd
```diff
@@ -1,35 +1,4 @@
...
-
-:: find VS install path
-for /f "usebackq tokens=*" %%i in (`script-packages\vswhere.1.0.62\tools\vswhere -latest -products * -requires Microsoft.Component.MSBuild -property installationPath`) do (
-  set VS_INSTALL_PATH=%%i
-)
 
-:: run script
-"%VS_INSTALL_PATH%\MSBuild\15.0\Bin\Roslyn\csi.exe" reset-virtual-host.csx %*
\ No newline at end of file
+dotnet run --project src/targets --no-launch-profile -- %*
```
bording: What's the need for `--no-launch-profile`?

src/targets/Targets.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
bording: Minor nitpick, but I prefer to have `TargetFramework` be the first property.

reset-virtual-host.cmd
adamralph: @bording I usually have that in there so that if I'm ever debugging the targets project in VS, and supplying arguments to it, it will not affect running the command in the console. However, you can argue YAGNI, so I'm happy to remove it.

src/targets/Targets.csproj
adamralph: I prefer alphabetical order, since the list of properties can get quite big sometimes and I'd like a deterministic way to eyeball a specific property. I'm not hung up on this though. I can change it if you feel strongly enough about it.

adamralph:
@bording 

> I like this change, but I'm not sure I care for "Targets" being the name of the project. It doesn't feel descriptive enough.

I'm open to suggestions.

> There is also a casing mismatch between the project name and the folder.

Yes, in this case, I can see the logic behind that. In this repo, all the code is under `src`, including the tests, which is why I also put this project under that top level folder. In some other projects, I have `targets` as a peer of `src` and `tests`, and I omit the `Targets` folder which would normally be below that for the project, since the top level `targets` will only ever contain one project. I'll hold off with the casing change until we decide a better name for the project.

src/targets/Targets.csproj
bording: If you look at the other Particular repos, having `TargetFramework` first is something I've standardized on.

bording:
> I'm open to suggestions.

Nothing is immediately coming to mind, but I know every time I see "Targets" I think MSBuild targets. The fact these are for "build script" things makes me want to see something that make it more clear that is what they are for.

bording:
@adamralph I never did figure out a name I like better for this, so let's just get it in as-is. Can you squash the fixup commits and then I'll merge it.

bording:
I went ahead and squashed the commits, so I'll merge this when it's green.

bording:
So this PR is old enough that the TeamCity changes mean it won't pick up the status any more. We'll need to close this PR an open a new one. :cry:

adamralph:
Replaced with https://github.com/Particular/NServiceBus.RabbitMQ/pull/535

PR #517, opened 2019-01-10
Use Core's TimeoutManager migration mode

bording:
This updates the transport to use Core's [TimeoutManager migration mode](https://github.com/Particular/NServiceBus/pull/5227) introduced in the 7.1 release.

This lets us get rid of the hack of needing to sometimes tell Core that we don't actually support native delays just to keep the TimeoutManager poller running.


src/NServiceBus.Transport.RabbitMQ/NServiceBus.Transport.RabbitMQ.csproj
```diff
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Fody" Version="3.3.3" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="[7.0.0, 8.0.0)" />
+    <PackageReference Include="NServiceBus" Version="[7.1.6, 8.0.0)" />
```
boblangley: Should this be `[7.1.0,8.0.0)`?

bording: Why? 7.1.0 was never released, so it would already need to be a patch version. Since that's the latest version out, I don't see a reason not to set the minimum version to it.

bording:
Replaced by #520.

PR #528, opened 2019-03-20
Improve Connection event logging

### LINKED ISSUE #468: Consider using connect blocked/unblocked events to log when connections are blocked
As mentioned in [Discourse](https://discuss.particular.net/t/rabbitmq-transport-blocked-connections-handling/298/7), it could be useful for debugging purposes to see when a connection has been blocked by the broker because of a memory or disk alarm.

We could add handlers for the relevant events that log when they occur.

bording:
This added logging for the connection blocked/unblocked events and also ensures that the log entry for when a connection is shut down unexpectedly is consistent.

Fixes #468

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
@@ -5,7 +5,7 @@
     using System.Security.Authentication;
     using System.Security.Cryptography.X509Certificates;
     using global::RabbitMQ.Client;
-    using NServiceBus.Logging;
+    using Logging;
 
     class ConnectionFactory
     {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -73,7 +76,20 @@ public IConnection CreateConnection(string connectionName, bool automaticRecover
...
 
-                return connectionFactory.CreateConnection(connectionName);
+                var connection = connectionFactory.CreateConnection(connectionName);
+
+                connection.ConnectionBlocked += (sender, e) => Logger.WarnFormat("'{0}' connection blocked: {1}", connectionName, e.Reason);
+                connection.ConnectionUnblocked += (sender, e) => Logger.WarnFormat("'{0}' connection unblocked}", connectionName);
+
+                connection.ConnectionShutdown += (sender, e) =>
+                {
+                    if (e.Initiator != ShutdownInitiator.Application || (e.Initiator == ShutdownInitiator.Application && e.ReplyCode != 200))
```
andreasohlund: Nit:

Would this read better as 

```
if (e.Initiator == ShutdownInitiator.Application && e.ReplyCode == 200)
{
   return;
}

Logger....                 
```

bording: Yeah, I suppose so. I'll change it.

PR #529, opened 2019-03-20
Update RabbitMQ.Client to 5.1.0

### LINKED ISSUE #501: Our RabbitMQ transport is not compatible with the latest RabbitMQ client (5.1.0)
RabbitMQ transport has a dependency range of (>= 5.0.1 && < 5.1.0) so it is incompatible with the latest RabbitMQ.Client.

bording:
Fixes #501 

I kept the maximum range to the next minor (5.2), but give the client's focus on actually following SemVer, perhaps we should consider opening it up to 6.0.0 instead?


andreasohlund:
>I kept the maximum range to the next minor (5.2), but give the client's focus on actually following SemVer, perhaps we should consider opening it up to 6.0.0 instead?

Given the apiapprovals you added to their repo I'd say that we should?

PR #530, opened 2019-03-21
Delete MAINTAINERS.md

andreasohlund:
Since we're moving away from maintainers and repo and PR guidelines is being documented centrally.

@bording  Should I pull the parts about updating the client out into a rabbitmq-client.md?


bording:
@andreasohlund Looking at that, I don't really see anything worth keeping.

PR #532, opened 2019-03-21
Fix leaking headers

### LINKED ISSUE #485: Header manipulations persist between processing attempts
When a header is added/deleted/modified in the first attempt to process a message and that attempt fails, these header manipulations are visible when the message is retried in a subsequent attempt.

### LINKED ISSUE #491: Header manipulations done while processing a message are visible in recoverability policy
When a header is added/deleted/modified in an attempt to process a message and that attempt fails, these header manipulations are visible when the recoverability policy is executed to determine if/how the message should be retried.

### LINKED ISSUE #492: Header manipulations done in recoverability policy are visible during next attempt to process a message
When a header is added/deleted/modified in a recoverability policy and the policy decides to retry the processing immediately, these header manipulations are visible during the subsequent attempts at processing the message.

bording:
Brought back the changes originally proposed in #482.

Fixes #485
Fixes #491 
Fixes #492

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_modifying_incoming_headers.cs
@@ -1,0 +1,110 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.TransportTests
+{
+    using System;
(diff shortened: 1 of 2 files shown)


bording:
The tests added here don't really belong here. We should get them in Core so we can remove them from here eventually.

Issue #537, opened 2019-04-11
Align RabbitMQ behavior when an exception happens in the recoverability policy


boblangley:
By not properly handling exceptions when the recoverability policy throws, and not raising a critical error, we prevent users from taking an action when the underlying messaging infrastructure is partially failing. Unnecessary logging clutters the log files with exception information already logged by recoverability in Core. Log level warn downplays the significance of the accident, masquerading transport errors.

PR #541, opened 2019-04-12
Exceptions thrown in recoverability policy are not propagated as critical error

boblangley:
Backport to 5.0 for #537

## Who's affected
Anyone using the transport.

## Symptoms
Exceptions thrown in recoverability policy are not propagated as critical error and hide the underlying infrastructure problems that user code could be responding to.

## Description
By not properly handling exceptions when the recoverability policy throws, and not raising a critical error, the transport does not allow taking an action when the underlying messaging infrastructure is partially failing. Unnecessary logging clutters the log files with exception information already logged by recoverability in Core. Log level warn downplays the significance of the accident, masquerading transport errors.

### DIFF
--- a/src/NServiceBus.RabbitMQ/packages.config
+++ b/src/NServiceBus.RabbitMQ/packages.config
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Fody" version="3.3.3" targetFramework="net452" developmentDependency="true" />
+  <package id="Fody" version="4.2.1" targetFramework="net452" developmentDependency="true" />
   <package id="GitVersionTask" version="3.6.5" targetFramework="net452" developmentDependency="true" />
   <package id="Janitor.Fody" version="1.6.5" targetFramework="net452" developmentDependency="true" />
   <package id="Microsoft.Diagnostics.Tracing.EventSource.Redist" version="1.1.28" targetFramework="net452" />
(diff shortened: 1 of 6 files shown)


src/NServiceBus.RabbitMQ/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Fody" version="3.3.3" targetFramework="net452" developmentDependency="true" />
+  <package id="Fody" version="4.2.1" targetFramework="net452" developmentDependency="true" />
```
bording: Why the Fody upgrade? If you were getting an error about `IntermediateDir`, just build again, and it works.

boblangley: I tried that, I even did git clean->rebuild->build->build. It didn't matter how many times, for whatever reason that did not work for me. The 4x release included a commit to address this and after upgrading, it fixed it.

Is there some reason _not_ to upgrade Fody?

bording:
I don't think this was ready to be merged yet. The Fody changes are a concern. I'm looking at the branch now.

src/NServiceBus.RabbitMQ/packages.config
bording: Fody alters the assembly, so it adds to the scope of changes being introduced in the patch. There have also been bugs in the past where it corrupts the PDBs in some way that makes them unusable for debugging. That's why I prefer to not upgrade Fody in hotfixes.

bording:
I was able to confirm that the Fody upgrade wasn't required, but the PR is already merged, so I guess we'll keep it.

PR #553, opened 2019-06-20
Add docker instructions

andreasohlund:


README.md
```diff
@@ -3,3 +3,12 @@
 The [NServiceBus.RabbitMQ NuGet package](https://www.nuget.org/packages/NServiceBus.RabbitMQ) provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/).
 
 For more information, see the [documentation](https://docs.particular.net/nservicebus/rabbitmq/).
+
+
+## Running Tests Locally
+
+By default the tests will connect to a RabbitMQ broker on localhost listening on port `5672`
```
bording: Where did this information come from? This isn't correct. You have to have a `RabbitMQTransport_ConnectionString` environment variable set.

andreasohlund: Just copy pasted it from mongodb:) I’ll add the correct instructions. 

Why not default to localhost:5672 to avoid messing with the env variable?

bording: > Why not default to localhost:5672 to avoid messing with the env variable?

I don't have a specific reason against it, but I don't think any of the people who worked on that code run a broker on localhost, so it didn't occur to us to do that.

It's also much clearer with the current approach about why the tests can't connect to a broker vs. having implicit behavior.

andreasohlund:
Fair enough, will update the instructions

andreasohlund:
This should be correct now

PR #557, opened 2019-10-17
Update NServiceBus.Core packages

timbussmann:


danielmarbach:
Brandon is not around today and tomorrow. I'll take the responsibility ;)

Issue #564, opened 2019-12-05
Is it Possible Extract AMQP connection string from NServiceBus.RabbitMQ


TraGicCode:
Hello,

We currently have an asp.net core 3.0 application using NServiceBus and are wanting to utilize the follow healthcheck nuget package.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs

Unfortunately, NServiceBus.RabbitMQ looks like it accepts a connection string ( which is in the format of a sql connection string ) instead of an AMQP connection string that rabbitmq expects ( https://www.rabbitmq.com/uri-spec.html ) and therefore we essentially either need to "massage" the NServiceBus RabbitMQ Connection string or have 2 connection strings in our appsettings.json

Is there any classes within NServiceBus.RabbitMQ that are part of the libraries public interface that we can use to convert the NserviceBus specific rabbitmq connection string to the mentioned AMQP connection string?

bording:
> Is there any classes within NServiceBus.RabbitMQ that are part of the libraries public interface that we can use to convert the NserviceBus specific rabbitmq connection string to the mentioned AMQP connection string?

No, there is no way to do that.

SeanFeldman:
After discussing this issue we'd like to slightly modify the suggestion in the following way:

1. Allow configuring the transport with a valid [RabbitMQ (client) connection string](https://www.rabbitmq.com/uri-spec.html).
1. Additional configurations that are NServiceBus transport specific to be exposed as configuration APIs.

@TraGicCode, this should work for your use-case as well. Would you agree with the suggestion?

TraGicCode:
@SeanFeldman ,

Sounds great to me!

SeanFeldman:
@TraGicCode we're moving forward with this and will release as part of the upcoming enhancement release. It will be possible to use an AMQP connection string and set the rest of the parameters that are not standard to the RMQ client (but used by NServiceBus) via code, using configuration API. Documentation will be updated accordingly.

Thank you for your patience.

PR #610, opened 2020-04-02
AMQP connection string support

SeanFeldman:
Related to https://github.com/Particular/NServiceBus.RabbitMQ/issues/564

Get the transport to work with AMQP connection string, and configure anything transport-specific via configuration API. Providing support to the currently supported transport-specific connection (key1=value1;key2=value2).


SeanFeldman:
Superceeded by #613

PR #636, opened 2020-05-21
Update Particular.Packaging to 0.8.0

bording:


bording:
Wanted to verify that new packaging doesn't cause any weird Compile problems, and it looks fine.

PR #639, opened 2020-05-27
Update to RabbitMQ.Client 6.0

bording:
This converts the transport to using the new 6.x client in the minimally invasive way possible.

Adding the message body copy early in the `Consumer_Received` handler seems to have done the trick of getting the tests to pass, but it's possible more changes might be needed.

However, this change doesn't fully shield users from the side effects of the original message body memory being gone by the time we process a message. We have two features in the transport that by-design leak this out to the user;

#### Custom message id strategy

We expose the message body memory through a `Func<BasicDeliverEventArgs, string>` the can be provided to the transport, and that means the `Body` property could potentially be inspected inside it.

####  `ContextBag` access

We also put the `BasicDeliverEventArgs` into the message and error pipeline's `ContextBag` so that you could retrieve it from a handler to access values from it. I don't believe this is documented, but we added it a while back as an alternative to having to copy every property and expose them through some custom API.

---
Both of these have the unfortunate side effect of exposing the internals of the client, and that means potentially accessing the body memory that is no longer valid. Since we are working on a new major, we could alter/remove these APIs, but as soon as the 6.2.0 client is out with Daniel's change, both of these would become safe again since they'd once again be accessible while the body memory is available.

So, does this give us a reason to wait for the 6.2.0 client, or are we okay with there being this potential problem until the 6.2.0 client is released?

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
+++ b/src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
@@ -46,7 +46,7 @@ namespace NServiceBus.Transport.RabbitMQ
         void BindToDelayInfrastructure(RabbitMQ.Client.IModel channel, string address, string deliveryExchange, string routingKey);
         void Initialize(RabbitMQ.Client.IModel channel, System.Collections.Generic.IEnumerable<string> receivingAddresses, System.Collections.Generic.IEnumerable<string> sendingAddresses);
         void Publish(RabbitMQ.Client.IModel channel, System.Type type, NServiceBus.Transport.OutgoingMessage message, RabbitMQ.Client.IBasicProperties properties);
-        void RawSendInCaseOfFailure(RabbitMQ.Client.IModel channel, string address, byte[] body, RabbitMQ.Client.IBasicProperties properties);
+        void RawSendInCaseOfFailure(RabbitMQ.Client.IModel channel, string address, System.ReadOnlyMemory<byte> body, RabbitMQ.Client.IBasicProperties properties);
         void Send(RabbitMQ.Client.IModel channel, string address, NServiceBus.Transport.OutgoingMessage message, RabbitMQ.Client.IBasicProperties properties);
         void SetupSubscription(RabbitMQ.Client.IModel channel, System.Type type, string subscriberName);
         void TeardownSubscription(RabbitMQ.Client.IModel channel, System.Type type, string subscriberName);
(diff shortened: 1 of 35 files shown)


bording:
@ramonsmits I made the changes we talked about. I think this should be ready to merge now.

I do have one other change I want to consider, but it's not directly related to the client upgrade, so I can raise that as a separate PR after this one is merged.

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -17,10 +17,7 @@ namespace NServiceBus
...
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> SetClientCertificate(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, System.Security.Cryptography.X509Certificates.X509Certificate2 clientCertificate) { }
         public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> SetClientCertificate(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, string path, string password) { }
-        [System.ObsoleteAttribute("Use `RabbitMQTransportSettingsExtensions.SetClientCertificate(TransportExtensions" +
-            "<RabbitMQTransport> transportExtensions, X509Certificate2 clientCertificate)` in" +
-            "stead. Will be treated as an error from version 6.0.0. Will be removed in versio" +
-            "n 7.0.0.", false)]
+        [System.ObsoleteAttribute(@"Use `RabbitMQTransportSettingsExtensions.SetClientCertificate(TransportExtensions<RabbitMQTransport> transportExtensions, X509Certificate2 clientCertificate)` instead. The member currently throws a NotImplementedException. Will be removed in version 7.0.0.", true)]
```
tmasternak, line 20: I assume this will be removed in a separate PR. Right?

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -31,34 +28,25 @@ namespace NServiceBus
...
         [System.ObsoleteAttribute("Use `RabbitMQTransportSettingsExtensions.UseCustomRoutingTopology(TransportExtens" +
             "ions<RabbitMQTransport> transportExtensions, Func<bool, IRoutingTopology>)` inst" +
             "ead. The member currently throws a NotImplementedException. Will be removed in v" +
-            "ersion 6.0.0.", true)]
-        public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> UseRoutingTopology<T>(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions)
-            where T : NServiceBus.Transport.RabbitMQ.IRoutingTopology, new () { }
+            "ersion 7.0.0.", true)]
+        public static NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> UseRoutingTopology(this NServiceBus.TransportExtensions<NServiceBus.RabbitMQTransport> transportExtensions, System.Func<bool, NServiceBus.Transport.RabbitMQ.IRoutingTopology> topologyFactory) { }
```
tmasternak, line 35: I assume this will be removed in a separate PR. Right?

PR #649, opened 2020-07-07
Exchange and Queue name length validation

### LINKED ISSUE #453: When binding name is too long the exception raised by the RabbitMQ client is meaningless
A customer came to me with the following error at endpoint startup:

```
RabbitMQ.Client.Exceptions.WireFormattingException: Short string too long; UTF-8 encoded length=260, max=255
   at RabbitMQ.Client.Impl.WireFormatting.WriteShortstr(NetworkBinaryWriter writer, String val)
   at RabbitMQ.Client.Framing.Impl.QueueBind.WriteArgumentsTo(MethodArgumentWriter writer)
   at RabbitMQ.Client.Impl.Command.TransmitAsSingleFrame(Int32 channelNumber, Connection connection)
```

Customer was using `DirectRoutingTopology` and one binding was too long, the thing is that all their events inherit from a base class, thus binding name was `namespace-base-class.namespace-derived-class`. They added a new event where the entire string was 5 chars longer than allowed. Being the above error useless it took us, the customer and myself, nearly to hours to figure out what was going on and to identify the source of the error.

Given that we are creating those bindings we should wrap the broker exception in a custom one adding some details such as the entity name that is causing the failure.

tmasternak:
fixes #453


bording:
I'm not sure this the approach we should take for trying to solve #453. 

The catching of exceptions around the declarations I believe are still needed because there are other scenarios where those can through.

I'm also not sure the length check you added will actually catch all scenarios for length problems. The underlying AMQP protocol allows for 255 bytes, not characters, so for UTF-8, that could result in fewer characters if they are multi-byte encoded.

For the direct routing topology, there is still a missing scenario where the binding key can be too long.

Based on this, I think we might just need to catch the exception and rethrow a friendlier exception.

tmasternak:
@bording the 6.x Rabbit Client [removed the check](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/e593e3642910847f94f9406338beb1629148e8d4#diff-f706a6b6696e2a6f77fe83ffc7a031c1L441) altogether. 

The current behavior is that when a name gets encoded to more than 255 bytes the call will timeout on the client-side. Looking at the broker logs it seems that this is the client library issue.

In summary, the client is no longer throwing an exception that we can "wrap". Secondly, in my opinion current behavior is even more confusing than previously. As a result, I'd rather do the check on our end.

bording:
>@bording the 6.x Rabbit Client removed the check (in WireFormatting.cs line 442) altogether.

>The current behavior is that when a name gets encoded to more than 255 bytes the call will timeout on the client-side. Looking >at the broker logs it seems that this is the client library issue.

This is definitely a bug that should be fixed in the client: https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/907

At this point, I'd like to get this fixed there vs. trying to work around it in the transport.

Let's discuss what that means for this PR.

tmasternak:
@bording I agree that fixing that in the client is the best option. 

I think we should open a PR in the client.

tmasternak:
PR opened in the Rabbit Client https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/908

PR #650, opened 2020-07-08
Using delegates instead of generic `Func`.

ramonsmits:
I don't like the `Func<T>` style public API's. This is an alternative based on delegates.


ramonsmits:
Reviewed with @bording which didn't see much value in it and he prefers the Func style generic API's. It is a matter of personal preference. I've also looked into the performance aspect of [Func vs delegates](https://stackoverflow.com/questions/43769065/func-vs-custom-delegate-performance) and did my own test. There is no speed difference between either on my system:

|           Method |     Mean |    Error |   StdDev |
|----------------- |---------:|---------:|---------:|
| CallWithDelegate | 25.60 ns | 0.458 ns | 0.470 ns |
|     CallWithFunc | 27.28 ns | 0.197 ns | 0.185 ns |

The deciding factor is how much value is this breaking API providing to our customer and that is no value except syntactic sugar which is why this is closed.

PR #665, opened 2020-08-24
Adjust to client changes embracing the new concurrency setting

danielmarbach:


src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -82,12 +83,14 @@ public ConnectionFactory(string endpointName, ConnectionConfiguration connection
...
 
-        public IConnection CreateConnection(string connectionName, bool automaticRecoveryEnabled = true)
+        public IConnection CreateConnection(string connectionName, bool automaticRecoveryEnabled = true, int consumerDispatchConcurrency = 1)
         {
             lock (lockObject)
             {
                 connectionFactory.AutomaticRecoveryEnabled = automaticRecoveryEnabled;
                 connectionFactory.ClientProperties["connected"] = DateTime.Now.ToString("G");
+                // it is OK to modify as long as we are under a lock
+                connectionFactory.ConsumerDispatchConcurrency = consumerDispatchConcurrency;
```
danielmarbach, line 93: I felt given there are assumptions that it is OK to modify the connection factory settings per connection creation and we already use a lock here I decided to just piggy back on this for now

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -76,10 +73,9 @@ public Task Init(Func<MessageContext, Task> onMessage, Func<ErrorContext, Task<E
         public void Start(PushRuntimeSettings limitations)
         {
             maxConcurrency = limitations.MaxConcurrency;
-            semaphore = new SemaphoreSlim(limitations.MaxConcurrency, limitations.MaxConcurrency);
```
danielmarbach, line 79: Technically we could still use a semaphore but since the client now takes care of the concurrency I replaced it with a simple tracker for the shutdown aspects of the code

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -63,8 +62,6 @@ public Task Init(Func<MessageContext, Task> onMessage, Func<ErrorContext, Task<E
 
             circuitBreaker = new MessagePumpConnectionFailedCircuitBreaker($"'{settings.InputQueue} MessagePump'", timeToWaitBeforeTriggeringCircuitBreaker, criticalError);
 
-            exclusiveScheduler = new ConcurrentExclusiveSchedulerPair().ExclusiveScheduler;
```
danielmarbach, line 66: See 

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/878#issuecomment-652609884

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/878/commits/77052546862e54df70819710f19431a9be0db71f

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -153,10 +150,16 @@ void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
...
         {
-            var eventRaisingThreadId = Thread.CurrentThread.ManagedThreadId;
+            if (messageProcessing.Token.IsCancellationRequested)
+            {
+                return;
+            }
+
+            Interlocked.Increment(ref numberOfExecutingReceives);
 
+            // technically we don't need this anymore
```
danielmarbach, line 162: Should I remove this one? Technically with the new concurrency code and the optimized buffer handling we don't require this. It is only here because core doesn't support readonly memory

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -117,7 +113,7 @@ public async Task Stop()
             consumer.Received -= Consumer_Received;
             messageProcessing.Cancel();
 
-            while (semaphore.CurrentCount != maxConcurrency)
+            while (Volatile.Read(ref numberOfExecutingReceives) != maxConcurrency)
```
bording, line 116: Is `Volatile.Read` actually required here? I always have to go read up on volatile stuff to recall when it might be needed.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
bording: Yeah I think it makes sense to move this back into the `Process` method and remove `eventArgsCopy` since it isn't needed now.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
bording: Also, I don't think the logic for the condition is correct now. With the semaphore, `CurrentCount == maxConcurrency` means there are no messages being processed.

With the counter approach, we need to wait until it is 0, right?

danielmarbach: Yeah sorry I wasn't careful enough here and rushed things in between other stuff. Will fix. Sorry for that lousy change

danielmarbach: Have a look, what condition do you prefer? I currently have `> 0` because I felt it is a more safe value than `!= 0` but I might be overthinking it

src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
bording: Yeah this seems fine to do here. Not sure it needs the comment since we already have other things being modified inside the lock.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -30,14 +30,13 @@ sealed class MessagePump : IPushMessages, IDisposable
         PushSettings settings;
         CriticalError criticalError;
         MessagePumpConnectionFailedCircuitBreaker circuitBreaker;
-        TaskScheduler exclusiveScheduler;
 
         // Start
         int maxConcurrency;
-        SemaphoreSlim semaphore;
+        long numberOfExecutingReceives;
```
bording, line 36: Hmm, I think I like `numberOfMessagesBeingProcessed` better

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -117,7 +113,7 @@ public async Task Stop()
             consumer.Received -= Consumer_Received;
             messageProcessing.Cancel();
 
-            while (semaphore.CurrentCount != maxConcurrency)
+            while (Interlocked.Read(ref numberOfExecutingReceives) > 0)
```
bording, line 116: This condition seems fine. It shouldn't be possible for it to be less than zero, but if it was somehow we'd still want to skip the loop.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -260,7 +217,7 @@ async Task Process(BasicDeliverEventArgs message, byte[] messageBody)
                         var contextBag = new ContextBag();
                         contextBag.Set(message);
 
-                        var messageContext = new MessageContext(messageId, headers, messageBody ?? Array.Empty<byte>(), transportTransaction, tokenSource, contextBag);
+                        var messageContext = new MessageContext(messageId, headers, message.Body.ToArray() ?? Array.Empty<byte>(), transportTransaction, tokenSource, contextBag);
```
bording, line 220: Do we want to recreate the array every time? Seems like we could do this once earlier in the method and then reuse the result.

danielmarbach: What if a mutator changes the byte array during the execution and then if the execution fails those changes would leak into the error pipeline or potentially worse the FLRs right?

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
danielmarbach: Happy to do the change. Will also remove the comment

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
bording: This would mean that the mutator has changed individual elements inside the array, which seems unlikely. That most likely result in message body that couldn't be deserialized.

If you look back to the message pump before we reacted to the 6.0 changes, [it just directly passed `message.Body` into the Context ctors](https://github.com/Particular/NServiceBus.RabbitMQ/blob/5.2.0/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs#L251), so I don't think there would be any harm in going back to something similar by doing the array conversion once.

danielmarbach: I'm not a fan of having code that can have side effects across retries but given this was always the case with this transport I'm not opposed to restoring back to this behavior given that you have more context in this transport than I do.

bording: > I'm not a fan of having code that can have side effects across retries

In general I'm not either, but I also think that this is core's responsibility not the transport.

bording (approved):

PR #697, opened 2020-10-16
Upgrade to Alpha 631 with DateTimeOffset changes

ramonsmits:


ramonsmits:
@bording In the current code base there is a "connected" property set. I've googled for info on it but previously it was set using the local time which likely was incorrect. Pinging you on this specific change as maybe you know more about it.

andreasohlund (approved):

bording:
> @bording In the current code base there is a "connected" property set. I've googled for info on it but previously it was set using the local time which likely was incorrect. Pinging you on this specific change as maybe you know more about it.

All client properties are exposed in the RabbitMQ management UI, so it is a way to know when that connection was established:

image


When I first added it, that was the only way to see that information. However, since that was first added, the RabbitMQ management UI added that information based on the server's clock, so it's redundant now. We could just remove it now and rely on the server-provided timestamp:

image


Notice that it also shows in local time, not UTC.

ramonsmits:
@bording based on your feedback I've removed setting the "connected" property.

@andreasohlund I think this is now ready to merge when build is 🟢

PR #685, opened 2020-10-12
Allow messages to be sent using the non-persistent delivery mode

andreasohlund:

PR #703, opened 2020-10-20
Avoid leaking the new header for non-persistent delivery onto the wire

andreasohlund:
Also including @bording 's comments from https://github.com/Particular/NServiceBus.RabbitMQ/pull/685

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
@@ -23,7 +23,7 @@
 
             var delayed = CalculateDelay(deliveryConstraints, out var delay);
 
-            properties.Persistent = !messageHeaders.ContainsKey(UseNonPersistentDeliveryHeader);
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -23,7 +23,15 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
 
             var delayed = CalculateDelay(deliveryConstraints, out var delay);
 
-            properties.Persistent = !messageHeaders.ContainsKey(UseNonPersistentDeliveryHeader);
+            if (messageHeaders.ContainsKey(UseNonPersistentDeliveryHeader))
```
bording, line 26: This could be replaced with `properties.Persistent = !messageHeaders.Remove(UseNonPersistentDeliveryHeader);`

andreasohlund: 👍

ramonsmits (approved):

PR #720, opened 2021-01-11
Upgrade to core V8

SzymonPobiega:


src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -1,34 +1,31 @@
...
 
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
         var connectionString = Environment.GetEnvironmentVariable("RabbitMQTransport_ConnectionString");
 
         if (string.IsNullOrEmpty(connectionString))
         {
-            throw new Exception("The 'RabbitMQTransport_ConnectionString' environment variable is not set.");
+            connectionString = "host=localhost";
+            //throw new Exception("The 'RabbitMQTransport_ConnectionString' environment variable is not set.");
```
tmasternak, line 20: Should we bring this back?

src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -77,22 +74,23 @@ void PurgeQueues()
...
-                    channel.QueuePurge(queue);
-                }
-                catch (Exception ex)
-                {
-                    Console.WriteLine("Unable to clear queue {0}: {1}", queue, ex);
-                }
-            }
-        }
+        //TODO: Cleanup
+        //var queues = queueBindings.ReceivingAddresses.Concat(queueBindings.SendingAddresses);
```
tmasternak, line 78: Maybe this can be handled as [in SQL](https://github.com/Particular/NServiceBus.SqlServer/pull/743/files#diff-f3f090d3380d288229eb729f362e4a68497ef1703b3feeef8968e686af0bbacfR61) via internal for-tests only properties?

src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/When_using_a_custom_message_id_strategy.cs
```diff
@@ -36,8 +36,7 @@ public Receiver()
                 EndpointSetup<DefaultServer>(c =>
                 {
                     c.EnableFeature<StarterFeature>();
-                    c.UseTransport<RabbitMQTransport>()
-                        .CustomMessageIdStrategy(m => customMessageId);
+                    ((RabbitMQTransport) c.ConfigureTransport()).MessageIdStrategy = m => customMessageId;
```
tmasternak, line 39: Should this use `ConfigureRabbitMQTransport` to avoid casting?

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -5,44 +5,58 @@ namespace NServiceBus
     public class RabbitMQTransport : NServiceBus.Transport.TransportDefinition
     {
         public RabbitMQTransport() { }
-        public override string ExampleConnectionStringForErrorMessage { get; }
-        public override NServiceBus.Transport.TransportInfrastructure Initialize(NServiceBus.Settings.SettingsHolder settings, string connectionString) { }
+        public RabbitMQTransport(string connectionString) { }
+        public System.Security.Cryptography.X509Certificates.X509Certificate2 ClientCertificate { get; set; }
+        public System.TimeSpan HeartbeatInterval { get; set; }
+        public string Host { get; set; }
```
tmasternak, line 11: Should we allow setting `Host` once the `ConnetionString` is set? 

Same question for other parts of the connection string.

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -5,44 +5,58 @@ namespace NServiceBus
...
-    }
 }
 namespace NServiceBus.Transport.RabbitMQ
 {
-    [System.Obsolete("The timeout manager has been removed, so it is no longer possible to consume lega" +
-        "cy delayed messages from timeout storage. Will be removed in version 8.0.0.", true)]
-    public class DelayedDeliverySettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
+    public class ConventionalRoutingTopology : NServiceBus.Transport.RabbitMQ.IRoutingTopology
+    {
+        public ConventionalRoutingTopology(bool useDurableExchanges, System.Func<System.Type, string> exchangeNameConvention = null) { }
```
tmasternak, line 41: Should the topology type be public? Are the members meant to be used by the users?

src/NServiceBus.Transport.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationWithAmqpTests.cs
```diff
@@ -1,124 +1,124 @@
...
-        }
-
-        [Test]
-        public void Should_set_default_retry_cert_passphrase()
-        {
-            Assert.AreEqual(defaults.CertPassphrase, null);
-        }
-    }
-}
+﻿//namespace NServiceBus.Transport.RabbitMQ.Tests.ConnectionString
```
tmasternak, line 1: Should we bring this back?

src/NServiceBus.Transport.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -13,53 +17,61 @@ class RabbitMqContext
...
-            routingTopology = new ConventionalRoutingTopology(true);
             receivedMessages = new BlockingCollection<IncomingMessage>();
 
             var connectionString = Environment.GetEnvironmentVariable("RabbitMQTransport_ConnectionString");
 
             if (string.IsNullOrEmpty(connectionString))
             {
-                throw new Exception("The 'RabbitMQTransport_ConnectionString' environment variable is not set.");
+                connectionString = "host=localhost";
+                //throw new Exception("The 'RabbitMQTransport_ConnectionString' environment variable is not set.");
```
tmasternak, line 29: Should we bring this back?

src/NServiceBus.Transport.RabbitMQ.Tests/RabbitMqContext.cs
```diff
@@ -13,53 +17,61 @@ class RabbitMqContext
...
-            messageDispatcher = new MessageDispatcher(channelProvider);
-
-            var purger = new QueuePurger(connectionFactory);
-
-            messagePump = new MessagePump(connectionFactory, new MessageConverter(), "Unit test", channelProvider, purger, TimeSpan.FromMinutes(2), 3, 0);
+            var connectionStringBuilder = new DbConnectionStringBuilder { ConnectionString = connectionString };
 
-            routingTopology.Reset(connectionFactory, new[] { ReceiverQueue }.Concat(AdditionalReceiverQueues), new[] { ErrorQueue });
+            //TODO: Parse any settings in the connection string 
+            var transport = new RabbitMQTransport { Host = (string)connectionStringBuilder["Host"] };
```
tmasternak, line 35: Should the connetion string be obligatory constructor paramter?

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -9,17 +16,255 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates new instance of the RabbitMQ transport.
+        /// </summary>
+        public RabbitMQTransport(string connectionString) 
+            : base(TransportTransactionMode.ReceiveOnly, true, true, true)
+        {
+            var connectionStringParams = AmqpConnectionString.Parse(connectionString);
```
tmasternak, line 171: Do we need to support `NServiceBusConnectionString` format as well?

tmasternak (changes requested):

src/NServiceBus.Transport.RabbitMQ/obsoletes-v8.cs
```diff
@@ -0,0 +1,171 @@
+﻿using System;
+
+#pragma warning disable 1591
+
+namespace NServiceBus
+{
+#pragma warning disable 618
+    static partial class RabbitMQTransportSettingsExtensions
```
timbussmann, line 8: should be public I guess?

src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
SzymonPobiega: Yep, I brought it back in the latest commit.

src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
SzymonPobiega: 👍 Added in the latest commit.

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
SzymonPobiega: I treat connection string as a way to pass initial bunch of parameters that you can tweak later. I think this is in line with the API shape because the property-based design allows you to set and re-set things as many times as you want.

src/NServiceBus.Transport.RabbitMQ.Tests/ConnectionString/ConnectionConfigurationWithAmqpTests.cs
SzymonPobiega: Yep. I added it back in the latest commit.

src/NServiceBus.Transport.RabbitMQ.Tests/RabbitMqContext.cs
SzymonPobiega: Yep. Added back.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
SzymonPobiega: Yep. Added.

src/NServiceBus.Transport.RabbitMQ.Tests/RabbitMqContext.cs
SzymonPobiega: I don't think so. The only obligatory thing that needs to be set is `Host`. Hard to decide if `RabbitMQTransport(string)` ctor should take a connection string or the host.

src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigurationHelpers.cs
```diff
@@ -0,0 +1,11 @@
+﻿using NServiceBus;
+using NServiceBus.Configuration.AdvancedExtensibility;
+using NServiceBus.Transport;
+
+static class ConfigurationHelpers
+{
+    public static RabbitMQTransport ConfigureRabbitMQTransport(this EndpointConfiguration configuration)
+    {
+        return (RabbitMQTransport) configuration.GetSettings().Get<TransportDefinition>();
```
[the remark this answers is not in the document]
SzymonPobiega: 👍 For now I switched to using `ConfigureTransport` and casting here. Not a big deal ;-)

src/NServiceBus.Transport.RabbitMQ/Configuration/AmqpConnectionString.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace NServiceBus.Transport.RabbitMQ
+{
+    class AmqpConnectionString
+    {
+        public static Action<RabbitMQTransport> Parse(string connectionString)
```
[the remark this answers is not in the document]
SzymonPobiega: Well, it could take the instance of the transport and set properties on it. I somehow don't like a mutating static method but I guess this is not much better since it returns a thing that does the mutation.

src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -15,7 +20,7 @@ class ConnectionFactory
         readonly global::RabbitMQ.Client.ConnectionFactory connectionFactory;
         readonly object lockObject = new object();
 
-        public ConnectionFactory(string endpointName, ConnectionConfiguration connectionConfiguration, X509Certificate2Collection clientCertificateCollection, bool disableRemoteCertificateValidation, bool useExternalAuthMechanism, TimeSpan? heartbeatInterval, TimeSpan? networkRecoveryInterval)
+        public ConnectionFactory(string endpointName, string host, int port, string vhost, string userName, string password, bool useTls, X509Certificate2Collection clientCertificateCollection, bool validateRemoteCertificate, bool useExternalAuthMechanism, TimeSpan heartbeatInterval, TimeSpan networkRecoveryInterval)
```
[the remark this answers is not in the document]
SzymonPobiega: There is _some_ stuff happening there besides accessing properties e.g. the `Port` property can be null and when it is passed here, we substiture null for a default value based on the `useTls` setting. So I'd rather keep it that way than have internal methods/properties on the transport definition to access these "postprocessed" property values.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -9,17 +16,256 @@
...
+
+        /// <summary>
+        /// Should the client validate the broker certificate when connecting via TLS.
+        /// </summary>
+        public bool ValidateRemoteCertificate { get; set; } = true;
+
+        /// <summary>
+        /// Specifies if an external authentication mechanism should be used for client authentication.
+        /// </summary>
+        public bool UseExternalAuthMechanism { get; set; } = false;
```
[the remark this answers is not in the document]
SzymonPobiega: Not sure I understand. You mean I should always specifiy default value for `bool` properties? I think this is just by accident.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -9,17 +16,256 @@
...
+            {
+                var receivingAddresses = receivers.Select(x => x.ReceiveAddress).ToArray();
+                await CreateQueueIfNecessary(receivingAddresses, sendingAddresses, connectionFactory).ConfigureAwait(false);
+            }
+
+            return new RabbitMQTransportInfrastructure(hostSettings, receivers, connectionFactory,
+                    routingTopology, channelProvider, converter, TimeToWaitBeforeTriggeringCircuitBreaker, PrefetchCountCalculation);
+        }
+
+        Task CreateQueueIfNecessary(string[] receivingQueues, string[] sendingQueues, ConnectionFactory connectionFactory)
```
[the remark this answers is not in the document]
SzymonPobiega: 👍

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -41,59 +38,57 @@ sealed class MessagePump : IPushMessages, IDisposable
...
             this.channelProvider = channelProvider;
-            this.queuePurger = queuePurger;
-            this.timeToWaitBeforeTriggeringCircuitBreaker = timeToWaitBeforeTriggeringCircuitBreaker;
-            this.prefetchMultiplier = prefetchMultiplier;
-            this.overriddenPrefetchCount = overriddenPrefetchCount;
+            this.prefetchCountCalculation = prefetchCountCalculation;
+            this.settings = settings;
+            this.criticalErrorAction = criticalErrorAction;
+
+            Subscriptions = new SubscriptionManager(connectionFactory, routingTopology, settings.ReceiveAddress);
```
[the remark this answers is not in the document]
SzymonPobiega: 👍

src/NServiceBus.Transport.RabbitMQ/obsoletes-v8.cs
SzymonPobiega: 👍

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
tmasternak: After double-checking I realized that the routing topology types do not expose any additional members than `IRoutingTopology`. I'm good with the current visbility.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -41,59 +38,60 @@ sealed class MessagePump : IPushMessages, IDisposable
...
-            this.timeToWaitBeforeTriggeringCircuitBreaker = timeToWaitBeforeTriggeringCircuitBreaker;
-            this.prefetchMultiplier = prefetchMultiplier;
-            this.overriddenPrefetchCount = overriddenPrefetchCount;
+            this.prefetchCountCalculation = prefetchCountCalculation;
+            this.settings = settings;
+            this.criticalErrorAction = criticalErrorAction;
+
+            if (settings.UsePublishSubscribe)
+            {
+                Subscriptions = new SubscriptionManager(connectionFactory, routingTopology, settings.ReceiveAddress);
```
bording, line 56: If the thing that creates a message pump has access to all the things required to create a `SubscriptionManager` it seems better to me for it to create one if needed instead. That means you can stop passing in `routingTopology` just to be able to create this.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -9,17 +16,256 @@
...
+
+        /// <summary>
+        /// The password to pass to the broker for authentication.
+        /// </summary>
+        public string Password { get; set; } = "guest";
+
+        /// <summary>
+        /// The routing topology to use. If not set the conventional routing topology will be used <seealso cref="ConventionalRoutingTopology"/>.
+        /// </summary>
+        public IRoutingTopology RoutingTopology
```
bording, line 71: Ok, yeah this is what I afraid of when I saw that the topology classes had been made public.. This is a strictly worse configuration experience compared to what we had before. This puts the onus on the caller to have to create a topology class in all instances of using this API.

In the past that would only have been required when they had created their own custom topology implementation.

src/NServiceBus.Transport.RabbitMQ/Configuration/AmqpConnectionString.cs
bording, line 7: This is parsing a connection string. The idea that this is returning an entire `RabbitMQTransport` instance to do that seems like a major smell to me.

We used to have a `ConnectionConfiguration` class for a reason.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -9,17 +16,256 @@
...
+
+        /// <summary>
+        /// Creates new instance of the RabbitMQ transport.
+        /// </summary>
+        public RabbitMQTransport(string connectionString) 
+            : base(TransportTransactionMode.ReceiveOnly, true, true, true)
+        {
+            if (connectionString.StartsWith("amqp", StringComparison.OrdinalIgnoreCase))
+            {
+                AmqpConnectionString.Parse(connectionString)(this);
```
bording, line 178: Having to call Parse and then cast like that? This is not something that should be required to do. Why not keep the `ConnectionConfiguration` class and not have all this going on?

bording (dismissed):
There is a lot of formatting cleanup that also need to be done, but I didn't call that out specifically anywhere.

Overall, I can follow the reasoning behind some/most of the changes, but the configuration stuff is a big change that I'm not sure I understand the point of just yet.

My initial reaction here is that we've made the configuration code much worse, and the experience is also going to be worse for people using the transport.

src/NServiceBus.Transport.RabbitMQ/Configuration/AmqpConnectionString.cs
SzymonPobiega: It does not return the instance but rather an action that configures that instance. Plus the `RabbitMQTransport` is now a configuration object only.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
SzymonPobiega: They only need to do it if they chose to select a non-default topology. In this version the conventional topology is made the default. Since very few users use the direct one, I don't this is going to be a major issue for the majority.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
SzymonPobiega: case? This is passing the current instance to the configuration action. Why having two classes with exactly same properties, one public and one private?

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
SzymonPobiega: Good point. I'll discuss that with other TF members to make this consistent across transports.

src/NServiceBus.Transport.RabbitMQ/Routing/ConventionalRoutingTopology.cs
```diff
@@ -21,55 +23,101 @@
...
     {
-        readonly bool useDurableExchanges;
-
-        public ConventionalRoutingTopology(bool useDurableExchanges)
+        /// <summary>
+        /// Creates a new instance of conventional routing topology.
+        /// </summary>
+        /// <param name="useDurableExchanges">Indicates whether exchanges and queues declared by the routing topology should be durable.</param>
+        /// <param name="exchangeNameConvention">The convention to use when naming exchanges.</param>
+        public ConventionalRoutingTopology(bool useDurableExchanges, Func<Type, string> exchangeNameConvention = null)
```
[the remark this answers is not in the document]
SzymonPobiega: Oh I didn't notice that the custom convention was previously only allowed in a direct routing topology (as an optional argument to the `UseDirectRoutingTopology` method. I don't think it is end of the world if we made if an optional argument also here. Is I mentioned above, the conventional topology is now default so only users who know what they are doing are going to instantiate this class.

src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -77,30 +74,30 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
...
+            if (dispatchProperties.DoNotDeliverBefore != null)
             {
                 delayed = true;
-                delay = Convert.ToInt64(Math.Ceiling((doNotDeliverBefore.At - DateTimeOffset.UtcNow).TotalSeconds));
+                delay = Convert.ToInt64(Math.Ceiling((dispatchProperties.DoNotDeliverBefore.At - DateTimeOffset.UtcNow).TotalSeconds));
 
                 if (delay > DelayInfrastructure.MaxDelayInSeconds)
                 {
-                    throw new Exception($"Message cannot be sent with {nameof(DoNotDeliverBefore)} value '{doNotDeliverBefore.At}' because it exceeds the maximum delay value '{TimeSpan.FromSeconds(DelayInfrastructure.MaxDelayInSeconds)}'.");
+                    throw new Exception($"Message cannot set to be delivered at '{dispatchProperties.DoNotDeliverBefore.At}' because the delay exceeds the maximum delay value '{TimeSpan.FromSeconds(DelayInfrastructure.MaxDelayInSeconds)}'.");
```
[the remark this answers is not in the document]
SzymonPobiega: 👍

src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -77,30 +74,30 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
...
+            else if (dispatchProperties.DelayDeliveryWith != null)
             {
                 delayed = true;
-                delay = Convert.ToInt64(Math.Ceiling(delayDeliveryWith.Delay.TotalSeconds));
+                delay = Convert.ToInt64(Math.Ceiling(dispatchProperties.DelayDeliveryWith.Delay.TotalSeconds));
 
                 if (delay > DelayInfrastructure.MaxDelayInSeconds)
                 {
-                    throw new Exception($"Message cannot be sent with {nameof(DelayDeliveryWith)} value '{delayDeliveryWith.Delay}' because it exceeds the maximum delay value '{TimeSpan.FromSeconds(DelayInfrastructure.MaxDelayInSeconds)}'.");
+                    throw new Exception($"Message cannot be delayed by '{dispatchProperties.DelayDeliveryWith.Delay}' because it exceeds the maximum delay value '{TimeSpan.FromSeconds(DelayInfrastructure.MaxDelayInSeconds)}'.");
```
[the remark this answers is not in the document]
SzymonPobiega: 👍

src/NServiceBus.Transport.RabbitMQ/Routing/DirectRoutingTopology.cs
```diff
@@ -8,40 +10,93 @@
     /// <summary>
     /// Route using a static routing convention for routing messages from publishers to subscribers using routing keys.
     /// </summary>
-    class DirectRoutingTopology : IRoutingTopology
+    public class DirectRoutingTopology : IRoutingTopology
```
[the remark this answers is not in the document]
SzymonPobiega: The only alternative would be to use enums but this makes is awkward when combined with a custom topology because that would have to be an instance.

In alternative could be some static field e.g. have `Topologies` static class with fields for direct and conventional topology. Then the user could do

```
config.RoutingTopology = Topologies.Direct;
config.RoutingTopology = Topologies.Conventional;
config.RoutingTopology = new MyCustomTopology();
config.RoutingTopology = new ConventionalRoutingTopology(false, oneConvention, anotherConvention)
```

but since the most common case of conventional topology is covered by default settings, I think that would be excessive complexity. If a user knows what direct topology is, they can surely be bothered by instantiating a public class.

tmasternak (approved):

src/NServiceBus.Transport.RabbitMQ/Routing/DirectRoutingTopology.cs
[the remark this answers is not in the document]
bording: Hard disagree here. If we are "improving" the transport seam, why is leading to APIs that are worse than what we had before?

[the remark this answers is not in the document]
SzymonPobiega: I disagree that this is worse if you take into account all the usage scenarios. The advantage of the new API is that is allows users to inspect the state of the configuration object.

SzymonPobiega:
@bording we discussed the transport config APIs during today's sync and we decided to:
 - revert making conventional the default topology. In order to enforce the choice of topology we added it as a constructor parameter. Topologies are still public in case someone wants to created them with non-default arguments
 - the `exchangeNameConvention` parameter to `ConventionalRoutingTopology` is going to continue to be there as it is useful in case of events being defined as nested classes. It is now not going to be exposed for the non-advanced users so I don't see any harm in exposing it
 - the prefetch count is now a custom delegate so that the user knows that the argument is and what they should return
 - the connection string has been made mandatory
 - the NServiceBus connection string is not removed because we don't feel like changes required to SC to make it work can be done in scope of this TF
 - the certificate stays are property. We feel like having the user new up an instance of `X509Certificante` is not a deal breaker for our users.

This is definitely not the last TF that is going to be run in scope of the V8 so there is no problem with DevExp making the APIs even more polished. We wanted to focus on getting them consistent while not blocking any existing use cases.

bording:
@SzymonPobiega Reading your response, I'll admit to be being rather disheartened, because I don't really see how much of that addresses the concerns I raised.

>the exchangeNameConvention parameter to ConventionalRoutingTopology is going to continue to be there as it is useful in case of events being defined as nested classes. It is now not going to be exposed for the non-advanced users so I don't see any harm in exposing it

This one specifically I **really** disagree with. There's basically almost nothing you can do with that parameter that leaves you with a correct working system. The only thing you could do would be to use [`FullName` instead](https://github.com/Particular/NServiceBus.RabbitMQ/issues/134). If you want to enable that, there should just be `bool` parameter you can set to opt in to set that behavior.

SzymonPobiega:
@bording I've removed the additional parameter. Should be now good to go.

SzymonPobiega:
@bording I had to merge it to be able to smoke test the transports. Any further adjustments to the API are going to be done by subsequent task forces, either in Architecture or DevExperience.

PR #745, opened 2021-02-26
Upgrade to core V8 852

SzymonPobiega:


src/NServiceBus.Transport.RabbitMQ/InternalsVisibleTo.cs
```diff
@@ -1,3 +1,4 @@
...
\ No newline at end of file
+[assembly: InternalsVisibleTo("NServiceBus.Transport.RabbitMQ.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100dde965e6172e019ac82c2639ffe494dd2e7dd16347c34762a05732b492e110f2e4e2e1b5ef2d85c848ccfb671ee20a47c8d1376276708dc30a90ff1121b647ba3b7259a6bc383b2034938ef0e275b58b920375ac605076178123693c6c4f1331661a62eba28c249386855637780e3ff5f23a6d854700eaa6803ef48907513b92")]
+[assembly: InternalsVisibleTo("NServiceBus.Transport.RabbitMQ.AcceptanceTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001007f16e21368ff041183fab592d9e8ed37e7be355e93323147a1d29983d6e591b04282e4da0c9e18bd901e112c0033925eb7d7872c2f1706655891c5c9d57297994f707d16ee9a8f40d978f064ee1ffc73c0db3f4712691b23bf596f75130f4ec978cf78757ec034625a5f27e6bb50c618931ea49f6f628fd74271c32959efb1c5")]
```
timbussmann, line 4: why is that necessary now?

SzymonPobiega: Because I had to add back an internal ctor for `ConventionalRoutingTopology` to make sure the tests are isolated. Otherwise they clash because the default version of the topology does not take the parent type of a nested class into account.

SzymonPobiega:
@timbussmann @tmasternak @mauroservienti ping

src/NServiceBus.Transport.RabbitMQ/InternalsVisibleTo.cs
timbussmann: is that related to the core 852 upgrade?

SzymonPobiega: Yes and no. I discovered while working on 852 that tests on my machine are not reliable (they were reliable on the build server).

timbussmann: > Yes and no. I discovered while working on 852 that tests on my machine are not reliable (they were reliable on the build server).

would have been better to split that into a dedicated PR I think.

Not a huge fan of this but since it's only for tests I think we can get this merged and look into alternatives if this becomes an issue.

timbussmann (approved):

mauroservienti (approved):

PR #763, opened 2021-03-11
Support cancellation

andreasohlund:


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -204,69 +228,73 @@ async Task Process(BasicDeliverEventArgs message)
...
-                        }
-                    }
+                    await onMessage(messageContext, messageProcessingCancellationTokenSource.Token).ConfigureAwait(false);
+                    processed = true;
                 }
-
-                if (processed && tokenSource.IsCancellationRequested)
+                catch (OperationCanceledException ex) when (messageProcessingCancellationTokenSource.IsCancellationRequested)
                 {
+                    Logger.Info("Message processing cancelled.", ex);
```
DavidBoike, line 252: Is it weird to do an Info that carries an exception? When the endpoint is shutting down do you really need a dozen exception traces stuffed into the log file? I don't think I've logged like this in any of the work I've done.

DavidBoike (approved):
Just one question more than anything.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
andreasohlund: I agree, should we remove the ex or skip logging here completely? (cc @adamralph )

adamralph: I'm clutching at straws here, but I guess it *might* be useful to look at the stack trace to see exactly where the operation was cancelled, but that is a `Debug()` thing at best.

However, it probably makes sense to keep the `Info()` (we've done that in other transports), and I'm not sure we'd ever go so far as doing this:

```suggestion
                    Logger.Info("Message processing cancelled.");
                    Logger.Debug("Message processing cancelled.", ex);
```

adamralph: Discussed with @andreasohlund and we agreed to remove the logging for now, to align with changes already merged in other transports.

Issue #525, opened 2019-03-01
Support multiple RabbitMQ node hostnames


AlanMacdonald:
The rabbitmq dotnet client project appears to support multiple hostnames in the ConnectionFactory class when calling CreateConnection.  NServiceBus doesn't utilise this functionality as far as I can tell, which means if using a RabbitMQ cluster we need a load balancer in front of it.  If NServiceBus supported a comma separated host list for connecting to with random host selection then it would probably suffice instead of needing a load balancer which would be easier.

bording:
We actually used to expose this from the transport, but it was removed in 3.0.0. Take a look at https://groups.google.com/forum/#!topic/particularsoftware/8EkMRN97hVU/discussion to see more about why we decided to remove it.

> If NServiceBus supported a comma separated host list for connecting to with random host selection then it would probably suffice instead of needing a load balancer which would be easier.

At the time we removed the feature, the client didn't do any sort of random selection for the initial connection attempt. It only randomized and tried the other entries during the automatic connection recovery process. We weren't using the client's built-in connection recovery at the time (we do now), so I don't think we got any benefit from exposing the feature.

Looking at the client code now, it seems to indicate that it would now randomly select an entry even for the initial connection attempt, but I'm not actually sure it's implemented correctly. I'd need to study the code a bit more to be sure, but I believe it might still only use the randomization during the automatic connection recovery process.

bennettfactor:
Hi, this would be useful for my use case, in which we would prefer to use client-defined config that is provided in the native C# RabbitMQ client.

tmasternak:
Hi @bennettfactor that you for sharing your feedback. This will be considered when planning the next enhancement release.

WilliamBZA:
Hi @bennettfactor, this will be included in the next major version of the transport. We're introducing support for Quorum queues via a `RabbitMQClusterTransport` which allows you to specify multiple nodes to connect to.

PR #795, opened 2021-04-15
Add support for quorum queues

WilliamBZA:
Adds a flag to enable setting the endpoint's incoming queues as Quorum queues during creation.

We'll have to manually discard incoming messages with an expired TTBR though


src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -199,6 +199,11 @@ public TimeSpan NetworkRecoveryInterval
             }
         }
 
+        /// <summary>
+        /// Should this endpoint's incoming queues be created as quorum queues.
+        /// </summary>
+        public bool IsEndpointQuorum { get; set; } = true;
```
andreasohlund, line 205: I think we should use a name that better match the rabbitmq concept to make it more clear that this isn't a NServiceBus feature. How about `UseRabbitMqQuorumQueues` or `UseQuorumQueues` ?

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -32,33 +32,29 @@ public class RabbitMQTransport : TransportDefinition
...
         /// <summary>
-        ///     Creates new instance of the RabbitMQ transport.
+        /// Creates new instance of the RabbitMQ transport.
         /// </summary>
         /// <param name="topology">The built-in topology to use.</param>
         /// <param name="connectionString">Connection string.</param>
-        public RabbitMQTransport(Topology topology, string connectionString)
-            : this(GetBuiltInTopology(topology), connectionString)
+        /// <param name="useQuorumQueue">Whether this endpoint consumes messages from quorum queues or classic queues.</param>
+        public RabbitMQTransport(Topology topology, string connectionString, bool useQuorumQueue = false)
```
timbussmann, line 40: I was wondering whether it would make more sense to introduce a cluster-specific transport type, e.g. `RabbitMQClusterTransport` or something like that. Also connects to https://github.com/Particular/NServiceBus.RabbitMQ/issues/525 which would use APIs only related for clusters.

andreasohlund: I like it!

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -199,6 +205,15 @@ public TimeSpan NetworkRecoveryInterval
             }
         }
 
+        /// <summary>
+        /// Specifies whether the endpoint should ignore failures to declare the incoming queues because the queue already exists with a different configuration (e.g. arguments).
+        /// This is set to <code>false</code> by default which will make the endpoint fail to start when the queue configuration does not match an existing queue.
+        /// This option has no effect if installers have been disabled.
+        /// </summary>
+        public bool AllowInputQueueConfigurationMismatch { get; set; } = false;
```
andreasohlund, line 213: Since the only thing we allow to mismatch is what we call `QueueMode` should we rename it to ``AllowInputQueueModeMismatch` to make it more clear?

timbussmann: technically this can mismatch on any type of argument mismatch between existing queue and queue created by the installers. It should be about the queue time 99% of the cases I guess.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -223,12 +226,29 @@ async Task Process(BasicDeliverEventArgs message, CancellationToken messageProce
...
+                headers.ContainsKey(NServiceBus.Headers.TimeToBeReceived))
+            {
+                var maxTimeToBeReceived = DateTime.ParseExact(headers[NServiceBus.Headers.TimeSent],
+                        "yyyy-MM-dd HH:mm:ss:ffffff Z",
+                        CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal)
+                    .Add(TimeSpan.Parse(headers[NServiceBus.Headers.TimeToBeReceived]));
+
+                if (maxTimeToBeReceived < DateTime.UtcNow)
+                {
+                    AckMessage();
```
timbussmann, line 247: should we remove that again @WilliamBZA ?

src/NServiceBus.Transport.RabbitMQ/Routing/DirectRoutingTopology.cs
```diff
@@ -129,11 +154,11 @@ void CreateExchange(IModel channel, string exchangeName)
 
             try
             {
-                channel.ExchangeDeclare(exchangeName, ExchangeType.Topic, useDurableExchanges);
+                channel.ExchangeDeclare(exchangeName, ExchangeType.Topic, useDurableEntities);
             }
             catch (Exception)
             {
-
+                // TODO: Any better way to make this idempotent?
```
timbussmann, line 161: FYI this has just been copied over from the conventional topology which contains this comment

timbussmann (approved):

PR #799, opened 2021-04-19
Parse AMQP connection strings at cleanup

### LINKED ISSUE #750: Tests purge operations cannot be run when using a secure connection (UseTls=true)
When testing against a broker that requires the connection to use TLS, e.g. Amazon MQ service broker, tests purge operations cannot be performed and if not commented out all tests will fail. The purge/clean-up steps in the tests create a custom connection factory to connect directly to the broker and perform clean-up tasks. The custom factory creation doesn't implement/support secure connections.

Once this issue and https://github.com/Particular/NServiceBus.RabbitMQ/issues/751 have been completed, the entire test suite should be re-run against an ActiveMQ instance to validate support.

### LINKED ISSUE #751: Some tests targeting .NET Core 3.x fail when run against Amazon MQ service broker
Running the tests suite against the Amazon MQ service broker causes some (11 all my attempts) tests to fail. Tests targeting the .NET Framework are all green. Only some of the tests targeting .NET Core 3.x are failing. The two more frequent reasons for failure are:

- "No handlers could be found for message type: <insert base class here>" which seems to point to something that is off whit inheritance support
- "An error occurred while attempting to extract logical messages from incoming physical message"

Once this issue and https://github.com/Particular/NServiceBus.RabbitMQ/issues/750 have been completed, the entire test suite should be re-run against an ActiveMQ instance to validate support.

WilliamBZA:
Fixes #750 
Fixes #751 

Allows the test suite to parse amqp connection strings as well as ones that are in the format of db connection strings.


WilliamBZA:
Ping @timbussmann @andreasohlund

src/NServiceBus.Transport.RabbitMQ.TransportTests/RabbitMqConnectionStringParser.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
```
andreasohlund, line 1: Should we link to `src/targets/RabbitMqConnectionStringParser.cs` rather than copy pasting it?

src/targets/RabbitMqConnectionStringParser.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
+using System.Data.Common;
+
+class RabbitMqConnectionStringParser
```
andreasohlund, line 4: Same as above

src/NServiceBus.Transport.RabbitMQ.TransportTests/RabbitMqConnectionStringParser.cs
WilliamBZA: Done :+1:

src/targets/RabbitMqConnectionStringParser.cs
WilliamBZA: Done :+1:

andreasohlund (approved):

timbussmann:
that will cause conflicts with the changes introduced in #795 :(

timbussmann:
would it be an option to incorporate the changes made here: https://github.com/Particular/NServiceBus.RabbitMQ/pull/795/files#diff-eb8304f617529b8783372b726dd26709f9259c279725d56dc7c09f7a47da95dd into this PR maybe?

WilliamBZA:
Rebased and changed the target branch to merge into #795 instead

PR #808, opened 2021-04-28
Add support for multiple RabbitMQ node hostnames

WilliamBZA:
Fixes https://github.com/Particular/NServiceBus.RabbitMQ/issues/525

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
+    using System.Linq;
     using System.Net.Security;
     using System.Security.Authentication;
     using System.Security.Cryptography.X509Certificates;
@@ -18,9 +19,9 @@
         readonly string endpointName;
         readonly global::RabbitMQ.Client.ConnectionFactory connectionFactory;
         readonly object lockObject = new object();
-        List<string> hostnames;
+        List<AmqpTcpEndpoint> hostnames;
 
-        public ConnectionFactory(string endpointName, string host, int port, string vhost, string userName, string password, bool useTls, X509Certificate2Collection clientCertificateCollection, bool validateRemoteCertificate, bool useExternalAuthMechanism, TimeSpan heartbeatInterval, TimeSpan networkRecoveryInterval, List<string> additionalHostnames = null)
+        public ConnectionFactory(string endpointName, string host, int port, string vhost, string userName, string password, bool useTls, X509Certificate2Collection clientCertificateCollection, bool validateRemoteCertificate, bool useExternalAuthMechanism, TimeSpan heartbeatInterval, TimeSpan networkRecoveryInterval, List<string> additionalHostnames)
         {
             if (endpointName is null)
             {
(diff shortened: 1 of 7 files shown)


andreasohlund (approved):

src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
```diff
@@ -103,7 +107,15 @@ public IConnection CreateConnection(string connectionName, bool automaticRecover
                 connectionFactory.AutomaticRecoveryEnabled = automaticRecoveryEnabled;
                 connectionFactory.ConsumerDispatchConcurrency = consumerDispatchConcurrency;
 
-                var connection = connectionFactory.CreateConnection(connectionName);
+                IConnection connection = null;
+                if (additionalHostnames != null && additionalHostnames.Count > 0)
+                {
+                    connection = connectionFactory.CreateConnection(additionalHostnames, connectionName);
```
timbussmann, line 113: shouldn't this also take the "primary" host name into account?

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
```diff
@@ -15,5 +15,14 @@ public RabbitMQClusterTransport(Topology topology, string connectionString, Queu
             : base(GetBuiltInTopology(topology), connectionString, queueMode)
         {
         }
+
+        /// <summary>
+        /// Adds a new cluster node
+        /// </summary>
+        /// <param name="hostName">An additional hostname of a node in the cluster.</param>
+        public void AddNode(string hostName)
```
timbussmann, line 23: so there is no option to change the port of the additional hosts?

WilliamBZA: Correct. The `CreateConnection` method accepts either a list of hostnames, or a list of `AmqpTcpEndpoint`s.

Theoretically, using `AmqpTcpEndpoint` might allow different ports etc., but I'm not sure if that's even true. We set the port etc. on the `ConnectionFactory` class, so I'm not sure how we would set multiple of them.

We could add an overload to accept instances of `AmqpTcpEndpoint`s? It would mean leaking a rabbit class to our API though.

src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs
WilliamBZA: :+1: Good catch! I've pushed a fix

andreasohlund (approved):

andreasohlund (approved):

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
```diff
@@ -9,12 +9,22 @@ public class RabbitMQClusterTransport : RabbitMQTransport
...
             : base(GetBuiltInTopology(topology), connectionString, queueMode, delayedDeliverySupportConfiguration == DelayedDeliverySupport.UnsafeEnabled)
         {
         }
+
+        /// <summary>
+        /// Adds a new node for use within a cluster.
+        /// </summary>
+        /// <param name="host">The hostname of the node.</param>
+        /// <param name="port">The port of the node.</param>
+        public void AddClusterNode(string host, int port = 5672)
```
timbussmann, line 25: I'm ok with this, but I'm still a bit unhappy about having to use a connection string for the first node and then this API for all further nodes. Feels weird somehow.

timbussmann (approved):

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
WilliamBZA: Given the API we have to work with, I can only assume there's some reason why the rabbit client only allows for these options to change. Fun note: this would never work with AMQP, so if your broker is an AMQP broker, you _HAVE_ to use a load balancer.

PR #816, opened 2021-05-06
Use connectionFactory's ssl options

WilliamBZA:
Fixes the connection to ssl and AMQPS brokers regression

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
@@ -1,7 +1,5 @@
 ﻿namespace NServiceBus
 {
-    using System;
-
     /// <summary>
     /// Transport definition for RabbitMQ in cluster configuration.
     /// </summary>
@@ -24,9 +22,9 @@
         /// </summary>
         /// <param name="host">The hostname of the node.</param>
         /// <param name="port">The port of the node.</param>
-        public void AddClusterNode(string host, int port = 5672)
+        public void AddClusterNode(string host, int port = -1)
         {
-            additionalHosts.Add(new Tuple<string, int>(host, port));
+            additionalHosts.Add((host, port));
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
```diff
@@ -24,7 +26,7 @@ public RabbitMQClusterTransport(Topology topology, string connectionString, Queu
         /// <param name="port">The port of the node.</param>
         public void AddClusterNode(string host, int port = 5672)
```
timbussmann, line 27: could we use `-1` as the default port instead? According to https://rabbitmq.github.io/rabbitmq-dotnet-client/api/RabbitMQ.Client.AmqpTcpEndpoint.html -1 represents the default port, which might be switch correctly depending on the SSL options?

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
```diff
@@ -24,7 +26,7 @@ public RabbitMQClusterTransport(Topology topology, string connectionString, Queu
...
         {
-            additionalHostnames.Add($"{host}:{port}");
+            additionalHosts.Add(new Tuple<string, int>(host, port));
```
timbussmann, line 29: can't we use the simplified `(x,y)` tuple syntax in this project?

WilliamBZA: [Computer says no](https://www.youtube.com/watch?v=TcRgyAkLeP4)

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
WilliamBZA: Updated.

andreasohlund (approved):

src/NServiceBus.Transport.RabbitMQ/RabbitMQClusterTransport.cs
timbussmann: http://gph.is/1UObEqG

timbussmann (approved):

PR #825, opened 2021-05-22
fix OCE catching, Task.Run, etc.

adamralph:


src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/QuorumQueues/When_using_delayed_delivery_with_disabled_timeouts.cs
```diff
@@ -94,9 +94,13 @@ public async Task Handle(StartSagaMessage message, IMessageHandlerContext contex
                     {
                         await RequestTimeout<SagaTimeout>(context, TimeSpan.FromMinutes(1));
                     }
-                    catch (Exception e)
+#if NET5_0_OR_GREATER
+                    catch (Exception ex) when (ex is not OperationCanceledException)
+#else
+                    catch (Exception ex) when (!(ex is OperationCanceledException))
+#endif
```
DavidBoike, line 101: Was this just for fun?

adamralph: I wish. 🙄 

With the C# 8 version, an IDE diagnostic (I forget the number) is reported as an error due to the `net5.0` TFM. If you apply the code fix, then it won't compile due to the other TFMs. 🎉 See https://github.com/Particular/RepoStandards/pull/46#issuecomment-842276538.

Another option is to suppress the diagnostic, with a note to unsuppress it when the project only targets `net5.0` 🤷.

Another option is to restrict the language version, again with a note to unrestrict it when the project only targets `net5.0`, but I think that's my least favourite option.

adamralph: this has gone now

andreasohlund (approved):

PR #836, opened 2021-06-11
Custom exchange name for generic messages

mikelneo:
Currently exchange name for generic message type does not contains generic params specific. 
This leads to a problem when endpoint handler needs to handle only `Message<Type 1>` type, it actually receive all `Message<>` messages.

After fix exchange name for generic messages will look like
`NameSpace:TypeName--GenericType1Namespace:GenericType1Name::GenericType2Namespace:GenericType2Name--`

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
+++ b/src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
+	  <PackageReference Include="Moq" Version="4.16.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
     <PackageReference Include="NServiceBus" Version="8.0.0-alpha.1895" />
     <PackageReference Include="NUnit" Version="3.13.2" />
--- a/src/NServiceBus.Transport.RabbitMQ.Tests/Routing/ConventionalRoutingTopologyTests.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.Tests/Routing/ConventionalRoutingTopologyTests.cs
@@ -1,0 +1,87 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.Tests.Routing
+{
+    using System.Collections.Generic;
+    using global::RabbitMQ.Client;
+    using Moq;
+    using NUnit.Framework;
+    using Unicast.Messages;
+
+    [TestFixture]
+    class ConventionalRoutingTopologyTests
+    {
+        class TestMessage
+        {
+        }
+
+        class GenericMessage<TTarget, TSource>
+        {
+        }
+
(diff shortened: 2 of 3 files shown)


src/NServiceBus.Transport.RabbitMQ.Tests/NServiceBus.Transport.RabbitMQ.Tests.csproj
```diff
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
+	  <PackageReference Include="Moq" Version="4.16.1" />
```
sungam3r, line 15:
```suggestion
    <PackageReference Include="Moq" Version="4.16.1" />
```

timbussmann:
Hi @mikelneo , thanks for your PR 👍  we will review your suggestion and get back to you within the next few days.

timbussmann:
Hey @mikelneo ,

Thanks for your patience. We discussed your PR internally but we agree that we currently don't plan to merge this PR. At the moment, we do not support generic message types or handlers, which means that there is also little value in implementing improvements/optimization to this functionality for a specific transport. While we do not actively prevent generic messages, this is something that we haven't intended to be a supported feature. That fact is unfortunately not well documented and we're trying to be more clear about this on our public documentation, e.g. via this PR: https://github.com/Particular/docs.particular.net/pull/5376.

However, this does not mean that we are not going to consider support for generic message types in the future, we need to get a better understanding of the value of generic message types first though. We would definitely love to get a better understanding why you've decided to use generic message types and what value those provide over regular message types or inheritance.

mikelneo:
@timbussmann thanks for answer!

> At the moment, we do not support generic message types or handlers

Now handling of generic message works fine even without any code changes:
```c#
public class UploadDocumentsHandler : IHandleMessages<DocumentUploaded<DocType>> 
{
...
}
```
This code successfully handles published messages. Only problem is that due to same exchange name for `DocumentUploaded<DocType1>` and `DocumentUploaded<DocType2>` it also tries to handle wrong message types (ConventionalRoutingTopology binds exchange to queue).

>We would definitely love to get a better understanding why you've decided to use generic message types and what value those provide over regular message types or inheritance.

Of cause I can solve this problems by inheritance like: `class DocumentUploadedType1: DocumentUploaded<DocType1>`, but from my point of view generic messages gives more flexibility for implementation. I decided to use generic messages to simplify messaging contracts. My generic message looks like:
```c#
public class DocumentUploaded<TDocType> 
{
    public int Id { get; set; }
    ... // other props

    public TDocType Document { get; set; } // requires separate handling for different types
}
```

>However, this does not mean that we are not going to consider support for generic message types in the future

Please, let me know if anything happens.

PS. You competitor library MassTransit supports generic message :)

timbussmann:
thanks for the additional information. How many doc types do you currently have? Given that you already have to define a custom handler for each doc type, and the generic parameter type, is this really saving much code over having a message type per doc type?

> Now handling of generic message works fine even without any code changes

We do not actively try to prevent this at the moment, we're just not recommending it because we do not have automated testing, backwards compatibility and wire level compatibility testing, documentation and or any sort of cross-transport guarantees for this at the moment.

PR #905, opened 2021-11-08
SPIKE: Testing subscribe to object

DavidBoike:


DavidBoike:
Spike served its purpose

PR #990, opened 2022-05-03
Add delay infrastructure migration tool

kentdr:
Adding a command-line tool to be used with the transport to migrate from the v1 delay infrastructure to the quorum-based v2 infrastructure.   It also has a command to install the v2 delay infrastructure and a utility to verify if a broker is properly configured to support quorum queues.


src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysCreateCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+                DelayInfrastructure.Build(channel);
+                channel.Close();
+
+                console.WriteLine("Delay infrastructure v2 created successfully");
+            }
+            catch (Exception ex)
+            {
+                console.WriteLine($"Fail: {ex.Message}");
+            }
+            finally
```
andreasohlund, line 45: Given the usings on line 31 and 32 do we need the finally at all?

src/NServiceBus.Transport.RabbitMQ.CommandLine/ConnectionFactoryBinder.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.CommandLine
+{
+    using System.CommandLine;
+    using System.CommandLine.Binding;
+    using System.Security.Cryptography.X509Certificates;
+
+    class ConnectionFactoryBinder : BinderBase<ConnectionFactory>
+    {
+        readonly Option<string> connectionStringOption;
```
andreasohlund, line 9: nitpick: should we move the fields to the bottom of the class to align with the other classes?

src/NServiceBus.Transport.RabbitMQ.CommandLine/NServiceBus.Transport.RabbitMQ.CommandLine.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <PackageReference Include="RabbitMQ.Client" Version="6.3.0" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta3.22114.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Particular.Packaging" Version="2.1.0" PrivateAssets="All" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="..\NServiceBus.Transport.RabbitMQ\Configuration\ConnectionConfiguration.cs" Link="Transport\ConnectionConfiguration.cs" />
```
andreasohlund, line 27: Are we doing the linking to avoid using internals visible to?

src/NServiceBus.Transport.RabbitMQ.CommandLine/SharedOptions.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            var connectionStringOption = new Option<string>(
+                name: "--connectionString",
+                description: $"Overrides environment variable '{ConnectionStringEnvironmentVariable}'",
+                getDefaultValue: () => Environment.GetEnvironmentVariable(ConnectionStringEnvironmentVariable) ?? string.Empty);
+
+            connectionStringOption.AddAlias("-c");
+
+            return connectionStringOption;
+
```
andreasohlund, line 19:
```suggestion
```

src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs
```diff
@@ -0,0 +1,215 @@
...
+                int skippedMessages = 0;
+                int processedMessages = 0;
+
+                for (int i = 0; i < messageCount && !cancellationToken.IsCancellationRequested; i++)
+                {
+                    var message = channel.BasicGet(currentDelayQueue, false);
+
+                    if (message == null)
+                    {
+                        // Queue is empty
```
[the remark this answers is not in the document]
kentdr: Since we're polling for messages here instead of using consume, null would just mean the queue is empty.  This could happen if the TTL's expire while I"m looping through the messages.

src/NServiceBus.Transport.RabbitMQ.CommandLine/NServiceBus.Transport.RabbitMQ.CommandLine.csproj
kentdr: We only needed a couple of small infrastructure-centric classes so we didn't think it was necessary to reference the entire assembly.

bording (approved):

PR #1006, opened 2022-05-20
Add command to migrate endpoints to quorum queues

andreasohlund:


src/NServiceBus.Transport.RabbitMQ.CommandLine/MigrateEndpointCommand.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                // delete the queue under migration
+                channel.QueueDelete(queueName);
+                Console.WriteLine($"Main queue removed");
+
+                //recreate the queue
+                channel.QueueDeclare(queueName, useDurableEntities, false, false, QuorumQueueArguments);
+                Console.WriteLine($"Main queue recreated as a quorum queue");
+
+                channel.QueueBind(queueName, queueName, EmptyRoutingKey);
```
kentdr, line 98: Would it be better to wait to re-bind the new queue until after the messages have been migrated from the temp queue?  It would better preserve some of the original message order I think.

kentdr (approved):

src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Endpoint/EndpointMigrateCommand.cs
```diff
@@ -118,6 +118,11 @@ void ConvertToQuorum(IModel channel, CancellationToken cancellationToken)
 
             console.WriteLine($"{numMessagesMovedToHolding} messages moved to the holding queue");
 
+            if (channel.MessageCount(queueName) > 0)
```
andreasohlund, line 121: In theory we could spin for some period of time but I think this is good enough. We should spell out clearly that direct sends (since we only support the conventional routing topology it would have to be direct sends from native integrations) are not supported in the docs

bording (approved):

PR #1019, opened 2022-06-09
Create endpoint command

kentdr:
This command will create the specified queue and any endpoints etc using the specified routing topology.


andreasohlund (approved):

bording (approved):
There still things around option descriptions, etc. that I think could use some tweaking, but I'll leave those for the editorial pass at the end.

PR #1031, opened 2022-06-30
Review user-facing text

DavidBoike:
Feel free to back out, edit, or remove any of these suggestions

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Configuration/QueueType.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Configuration/QueueType.cs
@@ -1,16 +1,17 @@
 ﻿namespace NServiceBus
 {
     /// <summary>
-    /// Specifies what queues should be used by the endpoint.
+    /// Specifies what queues should be used by the endpoint. <see cref="Quorum" /> queues should be considered the default option for a replicated queue type.
     /// </summary>
     public enum QueueType
     {
         /// <summary>
-        /// Classic RabbitMQ queues. These are the standard queues and used outside cluster configuration.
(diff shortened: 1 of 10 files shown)


src/NServiceBus.Transport.RabbitMQ/Configuration/QueueType.cs
```diff
@@ -1,16 +1,17 @@
 ﻿namespace NServiceBus
 {
     /// <summary>
-    /// Specifies what queues should be used by the endpoint.
+    /// Specifies what queues should be used by the endpoint. <see cref="Quorum" /> queues should be considered the default option for a replicated queue type.
```
kentdr, line 4:
```suggestion
    /// Specifies what queues should be used by the endpoint. <see cref="QueueType.Quorum">Quorum</see> queues should be considered the default option for a replicated queue type.
```

src/NServiceBus.Transport.RabbitMQ/Configuration/QueueType.cs
```diff
@@ -1,16 +1,17 @@
...
     /// </summary>
     public enum QueueType
     {
         /// <summary>
-        /// Classic RabbitMQ queues. These are the standard queues and used outside cluster configuration.
+        /// The original classic/mirrored RabbitMQ queues. These are provided for backward compatibility, but have weaknesses
+        /// compared to <see cref="Quroum" /> queues which make it less than ideal for replicated queues where reliability is key.
```
kentdr, line 10:
```suggestion
        /// compared to <see cref="QueueType.Quorum">Quorum</see> queues which make it less than ideal for replicated queues where reliability is key. 
```

kentdr (approved):

PR #1071, opened 2022-09-01
Fix immediate retries

bording:


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -26,6 +28,7 @@ sealed class MessagePump : IMessageReceiver, IDisposable
         readonly Action<string, Exception, CancellationToken> criticalErrorAction;
         readonly TimeSpan retryDelay;
         readonly string name;
+        readonly FastConcurrentLru<string, int> deliveryAttempts = new(1000);
```
WilliamBZA, line 31: The change to removing the tight loop means that the only scenario where timeouts need to be tracked are where a single message takes longer than the acknowledgement timeout.

This means the cache size needs to be large enough to track those messages and not have them be evicted by other messages that are processing successfully.

I would expect that a count of 1000 is sufficient to cover that scenario except in the case where the message that needs to be retried is requeued and there are more than 1000 messages in the queue already.

WilliamBZA (approved):

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
bording: > I would expect that a count of 1000 is sufficient to cover that scenario except in the case where the message that needs to be retried is requeued and there are more than 1000 messages in the queue already.

The number of other messages in the queue should only matter for the case where the message is requeued to the end of the queue, but that only happens for quorum queues, which don't use the cached values in the first place.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -371,59 +372,44 @@ async Task Process(AsyncEventingBasicConsumer consumer, BasicDeliverEventArgs me
...
+                Logger.Error($"Failed to retrieve ID from poison message. Moving message to queue '{settings.ErrorQueue}'...", ex);
                 await MovePoisonMessage(consumer, message, settings.ErrorQueue, messageProcessingCancellationToken).ConfigureAwait(false);
 
                 return;
             }
 
-            var processed = false;
-            var errorHandled = false;
-            var numberOfDeliveryAttempts = 0;
+            var numberOfDeliveryAttempts = GetDeliveryAttempts(message, messageId);
```
SzymonPobiega, line 381: Could we change the logic to only track delivery attempts if a message has failed i.e. move this line to 392? This way we won't need as large cache. Successful attempts won't even touch the cache. The difference is we will be tracking failures, not attempts but otherwise it should be fine.

bording: This is already happening because of the early exit when the redelivered flag is false.

bording: Thinking about this more, I can see now how moving the call could have additional benefit even with the early exit.

>  The difference is we will be tracking failures, not attempts but otherwise it should be fine. I think that such a change would only require as to use k => 1 in line 447.

This I'm still not understanding, though. You seem to be implying that there is a difference between attempts and failures, but at the point of needing to populate the `ErrorContext` value, those seem to be identical to me. 

Moving the call seems fine to me, but I don't see how that changes the logic of the new method in any way.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -432,10 +418,38 @@ async Task Process(AsyncEventingBasicConsumer consumer, BasicDeliverEventArgs me
...
+                    Logger.Warn($"Failed to acknowledge message '{messageId}' because the handler execution time exceeded the broker delivery acknowledgement timeout. Increase the length of the timeout on the broker. The message was returned to the queue.", ex);
+                }
+                else
+                {
+                    Logger.Warn($"Failed to acknowledge message '{messageId}' because the channel was closed. The message was returned to the queue.", ex);
+                }
+            }
+        }
+
+        int GetDeliveryAttempts(BasicDeliverEventArgs message, string messageId)
```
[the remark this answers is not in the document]
SzymonPobiega: 👍 for renaming it to convey the side effect nature of the function

SzymonPobiega (approved):

WilliamBZA (approved):

PR #1082, opened 2022-09-12
Bugfix/6.1/connection leaks

refactorinc:
fix various connection leaks

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -233,10 +233,6 @@
                 }
                 Logger.InfoFormat("'{0}': Connection to the broker reestablished successfully.", name);
 
-                if (oldConnection.IsOpen)
-                {
-                    oldConnection.Close();
-                }
                 oldConnection.Dispose();
             }
             catch (OperationCanceledException ex) when (messageProcessing.Token.IsCancellationRequested)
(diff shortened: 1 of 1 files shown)


tmasternak:
Thank you for contributing @refactorinc. Could you share some information on the specifics of the bug?

What kind of problems does this solve?

refactorinc:
Hello,

This pull request solves 2 issues (connection leaks) related to support
reference `ref:_00DD0mqNg._5004G2RnYdD:ref` created on 07.07.2022.
I have also replied to Particular Support with this reference, citing this
pull request.

Essentially, during restart of a node in a RabbitMQ cluster, we observe
between 2 and 1000+ connection leaks from a single application using
NServiceBus.
Concentrating our efforts on the latest version available to us at the time
(6.1.1) and RabbitMQ.Client 6.4.0; our findings are reflected in this pull
request.

I expect our findings apply equally to v7 and v8 codebases;
but if I understand well, v6.1 is the latest version supporting RabbitMQ <
3.10.

Kind regards,

Dominic

tmasternak:
@refactorinc thank you for sharing the context.

ramonsmits:
- Obsoleted by https://github.com/Particular/NServiceBus.RabbitMQ/pull/1112

ramonsmits:
We chose `IsOpen/Close/Dispose` over only `Dispose` to make `Close` more explicit. Only invoking `Dispose` would work too but internally calls `Abort` which sends a different disconnect reason message to the broker which we felt was not the right message.

However, invoking Close could result in an exception. This isn't an issue as in the next while loop iteration the connection should the be closed thus `IsOpen` would return `false`.

PR #1095, opened 2022-09-20
Spike a way to consume a message without the need to invoke the handlers first

SzymonPobiega:
Adds another LRU dictionary to keep track of messages that are to be consumed.

Messages marked as to be consumed are consumed _before_ the handler execution code in order to allow them to be consumed even if handler execution time always exceeds the `consumer_timeout` setting of the RabbitMQ broker.

This is similar to SQL Server transport's early consumption but the difference is that in RabbitMQ the `onError` callback is always executed _after_ the handler while in SQLT the `onError` is executed _before_.


SzymonPobiega:
Replaced by https://github.com/Particular/NServiceBus.RabbitMQ/pull/1096

PR #1096, opened 2022-09-20
Improve handling of immediate and delayed retries

bording:
Follow-up to #1071 

This PR introduces the following changes:

1. When relying on the delivery attempts collection (because of classic queues), we need to consider the delayed retries count as part of the collection key. Without this, the total delivery attempts count never gets reset between delayed retries.
2. When a message should be acknowledged but can't because the channel is closed and the message has been requeued, we now track that and ack it the next time we see it. This prevents the message from getting stuck in an endless retry loop.
3. The log message when we detect a consumer timeout being exceeded has been changed to an error. This error should always be acted on when we log it, so making it an error seems appropriate.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -29,6 +29,7 @@
         readonly TimeSpan retryDelay;
         readonly string name;
         readonly FastConcurrentLru<string, int> deliveryAttempts = new(100);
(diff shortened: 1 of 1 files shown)


SzymonPobiega (approved):

SzymonPobiega (approved):
Smoke-tested both the happy and unhappy paths. In both cases the handler waits for 3 seconds (`consumer_timeout` set to 1 second). In the unhappy path the handler throws an exception.

In the happy path the message is consumed after being returned to the queue and picked up again.

In the unhappy path the message undergoes regular immediate and delayed retries and is eventually sent to the `error` queue.

PR #1143, opened 2022-11-29
Add .NET 7 to test projects on master

abparticular:


jpalac (approved):

tmasternak (commented):
Are the code changes needed by net7?

abparticular:
They are not needed specifically for .NET 7, but adding an extra framework was causing the tests to fail. The quorum queue migration tests weren't cleaning up after themselves and leftover messages in the test queues were causing asserts to fail.

tmasternak (approved):

PR #1145, opened 2022-11-30
Add .NET 7 to test projects on release 7.0

abparticular:


jpalac (approved):

tmasternak:
@abparticular are the changes to tests required by net7 upgrade?

abparticular:
They are not needed specifically for .NET 7, but adding an extra framework was causing the tests to fail. The quorum queue migration tests weren't cleaning up after themselves and leftover messages in the test queues were causing asserts to fail.

tmasternak (approved):

PR #1185, opened 2023-01-16
Resolve infinite loop with ObjectDisposedException

ramonsmits:
- Resolves: https://github.com/Particular/NServiceBus.RabbitMQ/issues/1186

Reconnect method has a loop that did not anticipate on `messageProcessingCancellationTokenSource` to be disposed. This can result in `messageProcessingCancellationTokenSource.Token` throw an `ObjectDisposedException`. When that happens the while loop never breaks and create infinite log entries as fast as it can until the process is killed.

## Cancellation token handling

- Before disposing also canceling, cancellation the CTS should gracefully stop the reconnect loop in situations where Dispose is invoked by the environment but Stop didn't happen yet.
- Moved `.Token` out of loop as that can throw an ODE by passing it as an argument to Reconnect. By obtaining the CT once a ODE cannot occur on the CTS.Token.

## Disposed handling

- Not catching ObjectDisposedException as RabbitMQ client might somewhere throw it. If catching ODE the catches should be around specific APIs

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -117,7 +117,7 @@
         {
             maxConcurrency = limitations.MaxConcurrency;
             Logger.InfoFormat("Calling a change concurrency and reconnecting with new value {0}.", limitations.MaxConcurrency);
-            _ = Task.Run(() => Reconnect(), cancellationToken);
+            _ = Task.Run(() => Reconnect(messageProcessingCancellationTokenSource.Token), cancellationToken);
             return Task.CompletedTask;
         }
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -522,6 +522,7 @@ public void Dispose()
             }
 
             circuitBreaker?.Dispose();
+            messagePumpCancellationTokenSource?.Cancel();
             messagePumpCancellationTokenSource?.Dispose();
             messageProcessingCancellationTokenSource?.Dispose();
```
ramonsmits, line 527:
```suggestion
            messageProcessingCancellationTokenSource?.Cancel();
            messageProcessingCancellationTokenSource?.Dispose();
```

danielmarbach:
A potential alternative https://github.com/Particular/NServiceBus.RabbitMQ/pull/1187

danielmarbach (approved):

adamralph:
@ramonsmits given @danielmarbach has reviewed this, do you still need another review?

ramonsmits:
> @ramonsmits given @danielmarbach has reviewed this, do you still need another review?

@adamralph No, I think were good.

PR #1187, opened 2023-01-19
Fix object disposed exception when reconnect loop is running during shutdown

danielmarbach:
Alternative to #1185 which would preserve the current assumptions of the reconnect loop being around as long as the messages are allowed to process.

Furthermore, this makes sure we shouldn't need extra disposed handling in the reconnect loop. 

A briefly looked at the possibility to add a test for the scenario. It would be possible by adding some abstractions to the internal connection factory and combining it with the testing logger approach by [NServiceBus.Testing](https://docs.particular.net/nservicebus/testing/) but that would have significantly added more work for now.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -117,7 +117,7 @@
         {
             maxConcurrency = limitations.MaxConcurrency;
             Logger.InfoFormat("Calling a change concurrency and reconnecting with new value {0}.", limitations.MaxConcurrency);
-            _ = Task.Run(() => Reconnect(), cancellationToken);
+            _ = Task.Run(() => Reconnect(messageProcessingCancellationTokenSource.Token), cancellationToken);
             return Task.CompletedTask;
         }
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -117,7 +117,7 @@ public Task ChangeConcurrency(PushRuntimeSettings limitations, CancellationToken
         {
             maxConcurrency = limitations.MaxConcurrency;
             Logger.InfoFormat("Calling a change concurrency and reconnecting with new value {0}.", limitations.MaxConcurrency);
-            _ = Task.Run(() => Reconnect(), cancellationToken);
+            _ = Task.Run(() => Reconnect(messageProcessingCancellationTokenSource.Token), cancellationToken);
```
danielmarbach, line 120: I wonder why we add here the cancellation token as a second parameter to `Task.Run` this would help to make sure the task is not run in the scheduler when the token is triggered but when the task is already running it has no effect because then only the token from messageProcessingCancellationTokenSource.Token counts.

ramonsmits:
This requires the same changes as in #1185 to Cancel all owned CTS before disposing. Otherwise the loop will not terminate.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
ramonsmits: I guess it does not harm to have it. It is very unlikely that the reconnect loop task would not ran but there is nothing in it that requires it to be invoked so its safe to pass the CT to the Task.Run

danielmarbach:
We moved the good things from here into https://github.com/Particular/NServiceBus.RabbitMQ/pull/1185

PR #1189, opened 2023-01-19
Stopping the endpoint can cause log flooding with "Reconnecting to the broker failed: System.ObjectDisposedException"

ramonsmits:
Backport of #1185 to `release-6.1` branch

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -162,7 +162,7 @@
             {
                 //log entry handled by event handler registered in ConnectionFactory
                 circuitBreaker.Failure(new Exception(e.ToString()));
-                _ = Task.Run(() => Reconnect());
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -506,6 +506,7 @@ public void Dispose()
         {
             circuitBreaker?.Dispose();
             semaphore?.Dispose();
+            messageProcessing?.Cancel();
```
danielmarbach, line 509:
```suggestion
            // This makes sure that if the stop token hasn't been canceled the processing source is canceled
            // so that any possible reconnect attempt has the possibility to gracefully stop too.
            messageProcessing?.Cancel();
```

danielmarbach (changes requested):

danielmarbach:
The title of the PR should also be reworded to indicate the problem and not the solution

danielmarbach (approved):

PR #1223, opened 2023-04-05
Retries for messages with null headers causes a null ref exception

andreasohlund:


andreasohlund:
@bording pinging you for review of the reproduction (fix should be trivial) since this seems to have been introduced in https://github.com/Particular/NServiceBus.RabbitMQ/pull/1071

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -464,7 +464,7 @@ int GetDeliveryAttempts(BasicDeliverEventArgs message, string messageIdKey)
                 return attempts;
             }
 
-            if (message.BasicProperties.Headers.TryGetValue("x-delivery-count", out var headerValue))
+            if (message.BasicProperties.Headers != null && message.BasicProperties.Headers.TryGetValue("x-delivery-count", out var headerValue))
             {
                 attempts = Convert.ToInt32(headerValue) + 1;
```
dvdstelt, line 469: Seeing how crucial this code can be, should we not use something like Int32.TryParse here as well?

andreasohlund: Makes sense to me 👍

bording: This header value comes from the broker, so I didn't think it would be necessary to protect against the broker sending an invalid value.

andreasohlund: Ah, in that case I see less value in checking this as well 👍

andreasohlund:
@bording did you have some other concerns or should we go ahead with this?

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
dvdstelt: I _will_ remind you of this conversation if the next bug is in that line ;-)

src/NServiceBus.Transport.RabbitMQ.Tests/When_consuming_messages.cs
```diff
@@ -58,11 +58,53 @@ public void Should_move_message_without_message_id_to_error_queue()
...
                 var result = channel.BasicGet(ErrorQueue, true);
 
-                Assert.False(messageWasReceived, "Message should not be processed processed successfully.");
+                Assert.False(messageWasReceived, "Message should not be processed successfully.");
                 Assert.NotNull(result, "Message should be considered poison and moved to the error queue.");
             }
         }
 
+        [Test]
+        public async Task Should_handle_retries_for_messages_without_headers()
```
bording, line 67: The problem should only manifest when using classic queues. It could be worth having a test that shows that it works with quorum queues as well without needing the fix.

bording (approved):

src/NServiceBus.Transport.RabbitMQ.Tests/When_consuming_messages.cs
andreasohlund: I've added this in a separate PR https://github.com/Particular/NServiceBus.RabbitMQ/pull/1224

PR #1224, opened 2023-04-12
Confirm that header collection is always populated on redelivery for quorum queues

andreasohlund:
This switches the tests to use quorum queues by default since classic queues are being phased out


andreasohlund:
@bording the headers seems to be null for quorum queues as well, can you check if I have missed something?

andreasohlund:
@bording made the tweaks we talked about, please review

bording (approved):

Issue #1236, opened 2023-05-02
Automatic rate limiting feature might cause duplicate messages


timbussmann:
#### Description

When using the [automatic rate limiting](https://docs.particular.net/nservicebus/recoverability/#automatic-rate-limiting) feature with the RabbitMQ transport, the functionality might produce duplicate messages in some edge cases.

#### Expected behavior

When automatic rate limiting kicks in, message processing should be reduced to a single message at the time with the configured delay interval between further processing attempts until message processing succeeds again.

#### Actual behavior

In the specific cases where the configured consecutive failure counter is reached and the rate-limiting starts to kick in, a message that should be handed over to delayed retries or to the error queue can be duplicated. There is no impact on messages that are handled via immediate retries. This also only happens when the failure counter reaches the limit, further failures while the rate limiting is enabled do not create message duplicates.

This is caused by the implementation which tries to reduce the transports concurrency setting once the configured failure counter is reached. The concurrency change is achieved by [reconnecting to the broker with adjusted concurrency settings](https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs#LL120C34-L120C34). However, this is done while the (failed) message is still being processed. After the connection is reset, the error handling pipeline tries to send a copy of the message for delayed delivery or to the error queue, while the failed message is expected to be acknowledged afterward. While sending messages works with the new connection, the currently processed message can't be acknowledged via the new connection. Therefore the message will be available for processing again even though a copy has been sent for delayed retries or to the error queue.

#### Versions

- 8.0
- not yet verified on version 7

### Steps to reproduce

With the following configuration:

```c#
endpointConfiguration.Recoverability()
    .Delayed(d => d.NumberOfRetries(1))
    .Immediate(i => i.NumberOfRetries(1))
    .OnConsecutiveFailures(2,
        new RateLimitSettings(TimeSpan.FromSeconds(10),
            token =>
            {
                Console.WriteLine("rate limit started");
                return Task.CompletedTask;
            }, token =>
            {
                Console.WriteLine("rate limit ended");
                return Task.CompletedTask;
            }));
```
send a message to a handler throwing an exception. See the following log example.

### Relevant log output

```shell
2023-05-01 16:07:51.737 INFO  Processing message
message ID: 96dff5e4-f63b-4d60-8a48-aff500e8df38
Delayed retry counter: 0
[...]

PR #1257, opened 2023-06-14
Skip reconnect delay on first reconnect attempt

timbussmann:
An idea to improve the behavior when using [message throughput throttling](https://docs.particular.net/samples/throttling/) in which case the implementation always triggers a reconnect. The default implementation then by default waits 10 seconds causing an unexpected delay in immediate retry processing attempts.

At first thought, the first reconnect attempt can be attempted without any delay, while bringing in the delay in case the first reconnect attempt fails.

Not quite sure if that would cause issues with other calls to `Reconnect`, e.g. in `Channel_ModelShutdown`, `Connection_ConnectionShutdown`, and `Consumer_ConsumerCancelled` though.

Any thoughts on this?

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
@@ -275,16 +275,15 @@
 
                         connection.Dispose();
 
-                        Logger.InfoFormat("'{0}': Attempting to reconnect in {1} seconds.", name, retryDelay.TotalSeconds);
-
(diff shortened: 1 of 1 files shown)


timbussmann:
closing this as the 10 second delay is currently not necessarily bad to prevent duplicate messages as described on [this comment](https://github.com/Particular/NServiceBus.RabbitMQ/issues/1236#issuecomment-1594756717). While the approach proposed in https://github.com/Particular/NServiceBus.RabbitMQ/pull/1258 can gracefully handle the scenario by draining the in-flight messages before reconnecting.

PR #1258, opened 2023-06-15
Don't abort connection in ChangeConcurrency

timbussmann:
A spike to change the `ChangeConcurrency` implementation to not close the receive connection midflight. This behavior currently causes an exception when trying to ack any message being processed when ChangeConcurrency is called.

Note that the `ChangeConcurrency` method can assume that it is being invoked on a separate thread as other transports also handle it that way.

This is an attempt to instead stop the receiver gracefully, waiting for all inflight messages to complete processing before closing the connection and restarting it.

There are some first-attempt tests to verify both the transports capability to call stop/start multiple times on the receiver and simulating some ChangeConcurrency scenarios.

Replaces:

- #1257

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
@@ -16,8 +16,8 @@
         public async Task Should_handle_multiple_stop_calls_to_started_receiver(TransportTransactionMode transactionMode)
         {
             await StartPump(
-                (context, token) => Task.CompletedTask,
-                (context, token) => Task.FromResult(ErrorHandleResult.Handled),
+                (_, _) => Task.CompletedTask,
+                (_, _) => Task.FromResult(ErrorHandleResult.Handled),
                 transactionMode);
 
             await receiver.StopReceive();
@@ -32,12 +32,12 @@
         {
             var messageReceived = CreateTaskCompletionSource();
 
-            await StartPump((context, token) =>
+            await StartPump((_, _) =>
                 {
                     messageReceived.SetResult();
                     return Task.CompletedTask;
                 },
-                (context, token) => Task.FromResult(ErrorHandleResult.Handled), transactionMode);
+                (_, _) => Task.FromResult(ErrorHandleResult.Handled), transactionMode);
 
             await receiver.StopReceive();
             await receiver.StartReceive();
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -180,6 +184,7 @@ public async Task StopReceive(CancellationToken cancellationToken = default)
                 {
                     connectionShutdownCompleted.SetResult(true);
                 }
+                connection.Dispose();
```
timbussmann, line 187: stop should probably also dispose the `messageProcessingCancellationTokenSource` and `messagePumpCancellationTokenSource` which is currently only done in Dispose and would therefore not dispose any intermediary token sources.

src/NServiceBus.Transport.RabbitMQ.TransportTests/Changing_concurrency.cs
```diff
@@ -0,0 +1,86 @@
...
+                }, (_, ct) =>
+                {
+                    Assert.Fail();
+                    return Task.FromResult(ErrorHandleResult.RetryRequired);
+                },
+                transactionMode);
+
+            await SendMessage(InputQueueName);
+            await messageRetries.Task;
+            await Task.Delay(1000); // take some time to process any remaining message
```
timbussmann, line 35: I think this could be lowered quite a bit but not sure we can get rid of some wait time completely.

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
```diff
@@ -0,0 +1,108 @@
...
+                messageReceived.SetResult();
+                return Task.CompletedTask;
+            },
+                (context, token) => Task.FromResult(ErrorHandleResult.Handled), transactionMode);
+
+            await receiver.StopReceive();
+            await receiver.StartReceive();
+
+            await SendMessage(InputQueueName);
+            await messageReceived.Task;
```
danielmarbach, line 46: Is there something we could assert here? Not religious about it. Just a thought.

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
```diff
@@ -0,0 +1,108 @@
...
+                        // run async because the pump might block the return until all inflight messages are processed.
+                        var stopTask = Task.Run(async () =>
+                        {
+                            await receiver.StopReceive(token);
+                            TestContext.WriteLine("Stopped receiver");
+                        }, token);
+
+                        await SendMessage(InputQueueName, new Dictionary<string, string>() { { "Type", "Followup" } },
+                            context.TransportTransaction, cancellationToken: token);
+                        await Task.Yield();
```
danielmarbach, line 76: Can you explain why yielding helps?

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -117,7 +116,11 @@ public Task ChangeConcurrency(PushRuntimeSettings limitations, CancellationToken
         {
             maxConcurrency = limitations.MaxConcurrency;
             Logger.InfoFormat("Calling a change concurrency and reconnecting with new value {0}.", limitations.MaxConcurrency);
-            _ = Task.Run(() => Reconnect(messageProcessingCancellationTokenSource.Token), cancellationToken);
+            _ = Task.Run(async () =>
```
danielmarbach, line 119: I know you are not changing this offloading, and I'm aware of the current flaw with the change concurrency being done in the middle of the pipeline. As far as I recall, though, all other transport implementations execute the start stop sequence inline. Should we either align all other or do it here the same?

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
```diff
@@ -180,9 +189,16 @@ public async Task StopReceive(CancellationToken cancellationToken = default)
                 {
                     connectionShutdownCompleted.SetResult(true);
                 }
+                connection.Dispose();
```
danielmarbach, line 192: Shouldn't the order be something like

```csharp
            messagePumpCancellationTokenSource?.Cancel();
            messagePumpCancellationTokenSource?.Dispose();
            messageProcessingCancellationTokenSource?.Cancel();
            messageProcessingCancellationTokenSource?.Dispose();

            connection?.Dispose();
           circuitBreaker?.Dispose();
```
and also have the connection dispose outside the using block together with the other disposes?

src/NServiceBus.Transport.RabbitMQ.TransportTests/Changing_concurrency.cs
timbussmann: that's a good idea 👍

src/NServiceBus.Transport.RabbitMQ.TransportTests/Changing_concurrency.cs
```diff
@@ -0,0 +1,86 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using NServiceBus.TransportTests;
+    using NUnit.Framework;
+
+    public class Changing_concurrency : NServiceBusTransportTest
+    {
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
```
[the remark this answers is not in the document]
timbussmann: well yeah but Rabbit only has this one. But yes for others, this should be extended. I'm not even quite sure about the expected behavior when using TransactionMode.None though.

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
```diff
@@ -0,0 +1,108 @@
...
+        [TestCase(TransportTransactionMode.TransactionScope)]
+        public async Task Should_handle_multiple_stop_calls_to_started_receiver(TransportTransactionMode transactionMode)
+        {
+            await StartPump(
+                (context, token) => Task.CompletedTask,
+                (context, token) => Task.FromResult(ErrorHandleResult.Handled),
+                transactionMode);
+
+            await receiver.StopReceive();
+            await receiver.StopReceive();
```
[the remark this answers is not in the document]
timbussmann: I've changed the second one to an Assert.DoesNotThrowAsync, have a look if you like this better. Imo it's a bit more straining on the eye because the Assert methods are not task returning and I find that a bit ugly (especially when considering adopting this approach into the next tests) but I also don't have a strong opinion on this though.

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
timbussmann: hm not sure what else other than an invocation counter would make sense here?

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_restarting_endpoint.cs
timbussmann: I think it was originally to help ensure that Stop completed before continuing but I think it's no longer needed.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
timbussmann: yeah I forgot to commit that change - my bad.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs
timbussmann: hm I thought it might be related to triggering the shutdown event that is connected to the task completion source but I think it's not necessary.

danielmarbach (approved):

danielmarbach:
Bump

PR #1270, opened 2023-06-22
Ensure reconnect loop in ChannelProvider terminates when disposed

### LINKED ISSUE #1268: ChannelProvider Publish reconnection attempts continue forever after IEndpointInstance had been stopped
#### Description
Our application manages endpoints on the runtime, e.g. creates and terminates various endpoints depending on application logic. Due to the dynamic nature of this process, it is possible that an endpoint need to be terminated, while the RabbitMQ broker it used to be connected to isn't available, for example had been shutdown.
Note: A precondition is to have the bus endpoint successfully be created and connected to the broker in the first place.
#### Expected behavior
When IEndpointInstance.Stop() had completed without an error (Task await completed without an exception), the RabbitMQ transport that backed up this bus must had been shutdown as well.
#### Actual behavior
When endpoint is stopped while it is in an error state (RabbitMQ broker had been shutdown for example), NServiceBus continues to periodically log entries such as:

```
2023-06-20 14:52:08.0278 [26] INFO NServiceBus.Transport.RabbitMQ.ChannelProvider 'XXX Publish': Reconnecting to the broker failed: RabbitMQ.Client.Exceptions.BrokerUnreachableException: None of the specified endpoints were reachable

 ---> System.AggregateException: One or more errors occurred. (Connection failed)
 ---> RabbitMQ.Client.Exceptions.ConnectFailureException: Connection failed
 ---> System.Net.Sockets.SocketException (10060): A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Threading.Tasks.ValueTask.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
--- End of stack trace from previous location ---
   at RabbitMQ.Client.Impl.TcpClientAdapter.ConnectAsync(String host, Int32 port)
   at RabbitMQ.Client.Impl.TaskExtensions.TimeoutAfter(Task task, TimeSpan timeout)
   at RabbitMQ.Client.Impl.SocketFrameHandler.ConnectOrFail(ITcpClient socket, AmqpTcpEndpoint endpoint, TimeSpan timeout)
   --- End of inner exception stack trace ---
   at RabbitMQ.Client.Impl.SocketFrameHandler.ConnectOrFail(ITcpClient socket, AmqpTcpEndpoint endpoint, TimeSpan timeout)
   at RabbitMQ.Client.Impl.SocketFrameHandler.ConnectUsingAddressFamily(AmqpTcpEndpoint endpoint, Func`2 socketFactory, TimeSpan timeout, AddressFamily family)
   at RabbitMQ.Client.Impl.SocketFrameHandler..ctor(AmqpTcpEndpoint endpoint, Func`2 socketFactory, TimeSpan connectionTimeout, TimeSpan readTimeout, TimeSpan writeTimeout)
   at RabbitMQ.Client.ConnectionFactory.CreateFrameHandler(AmqpTcpEndpoint endpoint)
   at RabbitMQ.Client.EndpointResolverExtensions.SelectOne[T](IEndpointResolver resolver, Func`2 selector)
   --- End of inner exception stack trace ---
   at RabbitMQ.Client.EndpointResolverExtensions.SelectOne[T](IEndpointResolver resolver, Func`2 selector)
   at RabbitMQ.Client.ConnectionFactory.CreateConnection(IEndpointResolver endpointResolver, String clientProvidedName)
   --- End of inner exception stack trace ---
   at RabbitMQ.Client.ConnectionFactory.CreateConnection(IEndpointResolver endpointResolver, String clientProvidedName)
   at NServiceBus.Transport.RabbitMQ.ConnectionFactory.CreateConnection(String connectionName, Boolean automaticRecoveryEnabled, Int32 consumerDispatchConcurrency) in /_/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs:line 142
   at NServiceBus.Transport.RabbitMQ.ConnectionFactory.CreatePublishConnection() in /_/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionFactory.cs:line 114
   at NServiceBus.Transport.RabbitMQ.ChannelProvider.CreateConnection() in /_/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs:line 24
   at NServiceBus.Transport.RabbitMQ.ChannelProvider.ReconnectSwallowingExceptions(String connectionName) in /_/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs:line 51 
2023-06-20 14:52:08.0278 [26] INFO NServiceBus.Transport.RabbitMQ.ChannelProvider 'XXX Publish': Attempting to reconnect in 10 seconds. 
...
```

This means, the transport had not been completely shutdown and cleaned up. This creates needless load on our application as this may occur for multiple bus entries. Additionally it creates error logging (yes, we know we can suppress it for the time being) that's cluttering our application log.

#### Versions
8.0.2

### Steps to reproduce

1. Have a running and accessible RabbitMQ server
2. Create and start an endpoint
3. Stop the RabbitMQ server
4. Wait a few seconds until NServiceBus Transport and core detect the problem and begin re-connection attempts
5. Stop the endpoint, but ensure the process remains active (`await Task.Delay(-1)`)

☝🏼 Observe that the console and log indicate that the transport is still reconnecting.

6. Start the RabbitMQ server

☝🏼 Observe that the console and log indicate that the transport reconnects and that a connection to the broker remains active.

### Relevant log output

_No response_

#### Workarounds

I have not found a way to mitigate it with application-level code.

ramonsmits:
- Fixed #1268

1. ChannelProvider now has a CancellationTokenSource that gets cancelled when it is Disposed to ensure any reconnect while loop to be stopped
2. Shutdown Handler will teardown and dispose existing channel if it exists.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -96,8 +96,15 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public void Dispose()
         {
-            stoppingTokenSource.Cancel();
-            stoppingTokenSource.Dispose();
+            try
+            {
+                stoppingTokenSource.Cancel();
+                stoppingTokenSource.Dispose();
+            }
+            catch (ObjectDisposedException)
+            {
+                // .Cancel can throw if already disposed
+            }
 
             connection?.Dispose();
 
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -21,7 +21,8 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public void CreateConnection()
         {
-            connection = connectionFactory.CreatePublishConnection();
+            connection?.Dispose();
+            connection = connectionFactory.CreatePublishConnection(); // Can take over 5 seconds
             connection.ConnectionShutdown += Connection_ConnectionShutdown;
         }
 
@@ -31,24 +32,33 @@ namespace NServiceBus.Transport.RabbitMQ
             {
                 var connection = (IConnection)sender;
 
-                // Task.Run() so the call returns immediately instead of waiting for the first await or return down the call stack
-                _ = Task.Run(() => ReconnectSwallowingExceptions(connection.ClientProvidedName), CancellationToken.None);
+                // Task.Run() to clarify intent that the call MUST return immediately and not rely on current async call stack behavior
+                _ = Task.Run(() => ReconnectSwallowingExceptions(connection.ClientProvidedName, stoppingTokenSource.Token), CancellationToken.None);
             }
         }
 
-#pragma warning disable PS0018 // A task-returning method should have a CancellationToken parameter unless it has a parameter implementing ICancellableContext
-        async Task ReconnectSwallowingExceptions(string connectionName)
-#pragma warning restore PS0018 // A task-returning method should have a CancellationToken parameter unless it has a parameter implementing ICancellableContext
+        async Task ReconnectSwallowingExceptions(string connectionName, CancellationToken cancellationToken)
         {
-            while (true)
+            while (!cancellationToken.IsCancellationRequested)
             {
                 Logger.InfoFormat("'{0}': Attempting to reconnect in {1} seconds.", connectionName, retryDelay.TotalSeconds);
 
-                await Task.Delay(retryDelay).ConfigureAwait(false);
-
                 try
                 {
+                    await Task.Delay(retryDelay, cancellationToken).ConfigureAwait(false);
                     CreateConnection();
+
+                    // A  race condition is possible where CreateConnection is invoked during Dispose
+                    // where the returned connection isn't disposed so invoking Dispose to be sure
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        connection.Dispose();
+                    }
+                    break;
+                }
+                catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+                {
+                    Logger.InfoFormat("'{0}': Stopped trying to reconnecting to the broker due to shutdown", connectionName);
                     break;
                 }
                 catch (Exception ex)
@@ -86,6 +96,9 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public void Dispose()
         {
+            stoppingTokenSource.Cancel();
+            stoppingTokenSource.Dispose();
+
             connection?.Dispose();
 
             foreach (var channel in channels)
@@ -98,6 +111,7 @@ namespace NServiceBus.Transport.RabbitMQ
         readonly TimeSpan retryDelay;
         readonly IRoutingTopology routingTopology;
         readonly ConcurrentQueue<ConfirmsAwareChannel> channels;
+        readonly CancellationTokenSource stoppingTokenSource = new();
         IConnection connection;
 
         static readonly ILog Logger = LogManager.GetLogger(typeof(ChannelProvider));
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -21,20 +21,23 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public void CreateConnection()
         {
-            connection?.Dispose();
             connection = connectionFactory.CreatePublishConnection(); // Can take over 5 seconds
             connection.ConnectionShutdown += Connection_ConnectionShutdown;
         }
 
         void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
         {
-            if (e.Initiator != ShutdownInitiator.Application)
+            if (e.Initiator == ShutdownInitiator.Application)
             {
-                var connection = (IConnection)sender;
-
-                // Task.Run() to clarify intent that the call MUST return immediately and not rely on current async call stack behavior
-                _ = Task.Run(() => ReconnectSwallowingExceptions(connection.ClientProvidedName, stoppingTokenSource.Token), CancellationToken.None);
+                return;
             }
+
+            var connectionThatWasShutdown = (IConnection)sender;
+            var connectionName = connectionThatWasShutdown.ClientProvidedName;
+            connectionThatWasShutdown.Dispose();
+
+            // Task.Run() to clarify intent that the call MUST return immediately and not rely on current async call stack behavior
+            _ = Task.Run(() => ReconnectSwallowingExceptions(connectionName, stoppingTokenSource.Token), CancellationToken.None);
         }
 
         async Task ReconnectSwallowingExceptions(string connectionName, CancellationToken cancellationToken)
@@ -96,15 +99,13 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public void Dispose()
         {
-            try
+            if (disposed)
             {
-                stoppingTokenSource.Cancel();
-                stoppingTokenSource.Dispose();
+                return;
             }
-            catch (ObjectDisposedException)
-            {
-                // .Cancel can throw if already disposed
-            }
+
+            stoppingTokenSource.Cancel();
+            stoppingTokenSource.Dispose();
 
             connection?.Dispose();
 
@@ -112,6 +113,8 @@ namespace NServiceBus.Transport.RabbitMQ
             {
                 channel.Dispose();
             }
+
+            disposed = true;
         }
 
         readonly ConnectionFactory connectionFactory;
@@ -120,6 +123,7 @@ namespace NServiceBus.Transport.RabbitMQ
         readonly ConcurrentQueue<ConfirmsAwareChannel> channels;
         readonly CancellationTokenSource stoppingTokenSource = new();
         IConnection connection;
+        bool disposed;
 
         static readonly ILog Logger = LogManager.GetLogger(typeof(ChannelProvider));
     }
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -21,36 +21,45 @@ public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay,
...
+        Task reconnectTask = Task.CompletedTask;
+
         void Connection_ConnectionShutdown(object sender, ShutdownEventArgs e)
         {
             if (e.Initiator != ShutdownInitiator.Application)
             {
                 var connection = (IConnection)sender;
 
-                // Task.Run() so the call returns immediately instead of waiting for the first await or return down the call stack
-                _ = Task.Run(() => ReconnectSwallowingExceptions(connection.ClientProvidedName), CancellationToken.None);
```
bording, line 35: The `Task.Run` is intentional here and needs to stay. We need to ensure that the reconnection code is happening on a different thread than the internal thread from the client that event is called from.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -84,8 +93,15 @@ public void ReturnPublishChannel(ConfirmsAwareChannel channel)
...
 
-        public void Dispose()
+#pragma warning disable PS0018 // A task-returning method should have a CancellationToken parameter unless it has a parameter implementing ICancellableContext
+        public async ValueTask DisposeAsync()
+#pragma warning restore PS0018 // A task-returning method should have a CancellationToken parameter unless it has a parameter implementing ICancellableContext
         {
+            stoppingTokenSource.Cancel();
+            stoppingTokenSource.Dispose();
+
+            await reconnectTask.ConfigureAwait(false);
```
bording, line 103: I don't see there being any value in trying to wait for a possible reconnect loop here. Removing this would simplify things and remove some of the other changes in this PR.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -21,36 +21,45 @@ public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay,
 
         public void CreateConnection()
         {
+            if (connection is not null)
+            {
+                connection.ConnectionShutdown -= Connection_ConnectionShutdown;
+                connection.Dispose();
+                connection = null;
```
bording, line 28: There is really no reason to do this, considering the very next line in is to reassign this anyway.

bording (changes requested):
Feels like there are too many changes here that are clouding up the actual fix required, which would be to abort the reconnection loop if the channel provider has been shut down.

There is another reconnection loop that needs to be looked at: The one in the message pump. Has that been examined to see if it would have a similar issue?

Also, this PR should be targeting the `master` branch, not a release branch.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
danielmarbach: That happens anyway because the first thing the code does is asynchronously yielding in the delay.

bording: And then when the next change comes later that removes that, we've then also lost the context for why it needs to be on another thread. Still disagree with this change.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
danielmarbach: I brought up the same thing originally. Because from a shutdown perspective I agree. The problem is though if you don't await it can happen due to the asynchronous nature we race to create a new one while we try to dispose. All possible permutations of things can happen here while probably the most likely outcome will be things are disposed. This would make the shutdown deterministic

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
danielmarbach: Without this we could reliably reproduce races against the field leading to mask the original intended exception no longer being raised but Rsmon has more context

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
ramonsmits: Reverted the Task.Run and clarified the code comment

ramonsmits:
> Also, this PR should be targeting the `master` branch, not a release branch.

Now targeting master

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
ramonsmits: We were not sure if the event unsubscribing was needed but the Dispose + null is definitely required to prevent shutdown race conditions as the call on the next like can take several seconds when the remote is unavailable and fail.

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -21,36 +21,46 @@ public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay,
...
+            }
             connection = connectionFactory.CreatePublishConnection();
```
ramonsmits, line 30:
```suggestion
            connection = connectionFactory.CreatePublishConnection(); // Can take over 5 seconds
```

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
ramonsmits: And also problematic if one endpoint that didn't throw on stop is still holding on remote resources while another endpoint instance is spun up which is dependent on those same resources and rejected by the broker as it reached its connection limit.

If a user isn't interested in that it should simply not await and do `_ = endpointInstance.Stop();` but users should be able to assume that after `.Stop()` returns all resources are released.

ramonsmits:
> There is another reconnection loop that needs to be looked at: The one in the message pump. Has that been examined to see if it would have a similar issue?

Yes, that has been examined by @danielmarbach and that already has cancellation support that will ensure the loop will be terminated:

https://github.com/Particular/NServiceBus.RabbitMQ/blob/aa68a7dbd819453b44bc8fb52e284254fdc06cc8/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs#L262-L301

> Feels like there are too many changes here that are clouding up the actual fix required, which would be to abort the reconnection loop if the channel provider has been shut down.

You are absolutely right. Removed the "wait for reconnect loop" logic.

ramonsmits:
@danielmarbach @bording do you know if in the mean time any improvements on lifetime management / disposal have been applied to master already?

danielmarbach:
I don't

danielmarbach (approved):

src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
```diff
@@ -21,34 +21,47 @@ public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay,
...
+            if (e.Initiator == ShutdownInitiator.Application)
             {
-                var connection = (IConnection)sender;
-
-                // Task.Run() so the call returns immediately instead of waiting for the first await or return down the call stack
-                _ = Task.Run(() => ReconnectSwallowingExceptions(connection.ClientProvidedName), CancellationToken.None);
+                return;
             }
+
+            var connectionThatWasShutdown = (IConnection)sender;
```
abparticular, line 35: From just reading the code, I'm not sure if it is possible for this event handler to run for any other connection apart from the one referenced by the `connection` field?

In any case, if `sender` and `connection` are the same, should we set `connection` to null after disposing it?  This would enable us to avoid trying to use a disposed connection in `GetPublishChannel()`, which looks like it is currently possible.

danielmarbach: @abparticular I think we had some discussion about this in https://github.com/Particular/NServiceBus.RabbitMQ/pull/1270#discussion_r1239042012 but then eventually asynchronously concluded it is not strictly required, so we removed the explicit null assignment because the code today assumes when things are disposed nothing will acquire channels anymore.

Since the existing code already operated on the event arguments, I wanted to keep that in place to not make any other assumptions.

The downside of setting it to null in the case someone would acquire a publish channel is that you'd get NullRefs while today you would at least get some for of stack trace that shows a publish channel trying to access a disposed connection which might be more revealing should we ever run into such an edge case. 

What do you think?

abparticular: > the code today assumes when things are disposed nothing will acquire channels anymore

I don't understand how that assumption is valid.  Won't the message dispatcher potentially continue trying to get channels while the reconnection loop is running?

> The downside of setting it to null in the case someone would acquire a publish channel is that you'd get NullRefs while today you would at least get some for of stack trace that shows a publish channel trying to access a disposed connection which might be more revealing should we ever run into such an edge case.
>
> What do you think?

Was it considered to make GetPublishChannel() async and only return a channel after reconnection is successful (ie. when connection is not null)?

danielmarbach:
Closed in favor of https://github.com/Particular/NServiceBus.RabbitMQ/pull/1435

Issue #1272, opened 2023-06-28
Suppress creation of `nsb.v2.verify-stream-flag-enabled` queue


ramonsmits:
### Describe the feature.

The `nsb.v2.verify-stream-flag-enabled` gets created by NServiceBus to probe for broker features.

A method to not have this created is needed to keep the queues in the broker clean.

If this gets added this option should also be available in [ServiceControl via any of RabbitMQ's transport customizations](https://github.com/Particular/ServiceControl/tree/master/src/ServiceControl.Transports.RabbitMQ)

### Additional Context

The problem is needed to check if the broker supports the stream features which is required by the v2 delayed delivery infrastructure introduced in v7 of the transport:

- https://docs.particular.net/transports/upgrades/rabbitmq-6to7#new-version-of-the-delay-infrastructure

It is highly recommended to migrate delayed messages from the v1 to v2 delayed delivery infrastructure:

- https://docs.particular.net/transports/upgrades/rabbitmq-classic-to-quorum-migration#step-8-migrate-delayed-messages

More information about the delayed delivery infrastructure:

- https://docs.particular.net/transports/rabbitmq/delayed-delivery

bording:
This cannot be suppressed. It's required to verify that the version of the broker the transport is running against can actually be used.

ramonsmits:
Some more context:

1. The streams feature is implicitly used by the NServiceBus RabbitMQ delayed delivery infrastructure.
2. Not having the streams feature enabled on the RabbitMQ broker results in unreliable at most once delayed delivery.
3. The transport creates the streams queue to test if the broker has the streams feature enabled.
4. The transport does not delete this queue to prevent issues with multiple instances creating/deleting this queue.
5. Creating/deleting a queue is expensive. Having this idle queue might not be pretty but it is a lean method. When an instance tries to create an existing queue it basically behaves as a NOOP.
6. Eventually the streams feature will be enabled by default on the broker. The transport could eventually require that minimum version so it doesn't have to create this test stream queue.
8. While we could add a code API to not probe for this feature we will not do this simply because don't want any customer to accidentally lose messages. Such API calls are easy to add for testing and then accidentally remain there forever.

PR #1273, opened 2023-06-28
Delete `nsb.v2.verify-stream-flag-enabled` queue that gets created during feature probing

ramonsmits:
The `nsb.v2.verify-stream-flag-enabled` queue gets created during feature probing but the queue is not deleted resulting in a queue on the broker that isn't used.

Now the queue gets deleted after creation to not have this queue on the broker.

- Resolves: https://github.com/Particular/NServiceBus.RabbitMQ/issues/1272


bording:
This queue cannot be deleted. In isolation, this might seem like an okay change, but as soon as you have multiple endpoints all trying to create and delete this queue on startup, chaos ensues.

PR #1393, opened 2024-05-03
Prevent header dictionary resize & span based overloads

danielmarbach:
The message converter resizing in SC accounts for 20% of the allocation for header dictionaries

image

Before

image

After

image

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
@@ -67,9 +67,9 @@
                 deserializedHeaders[Headers.EnclosedMessageTypes] = properties.Type;
             }
 
-            if (deserializedHeaders.ContainsKey("NServiceBus.RabbitMQ.CallbackQueue"))
+            if (deserializedHeaders.TryGetValue("NServiceBus.RabbitMQ.CallbackQueue", out var callbackQueue))
             {
-                deserializedHeaders[Headers.ReplyToAddress] = deserializedHeaders["NServiceBus.RabbitMQ.CallbackQueue"];
+                deserializedHeaders[Headers.ReplyToAddress] = callbackQueue;
             }
 
             //These headers need to be removed so that they won't be copied to an outgoing message if this message gets forwarded
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
@@ -120,9 +120,9 @@
                 foreach (var kvp in dictionary)
                 {
                     sb.Append(kvp.Key);
-                    sb.Append("=");
+                    sb.Append('=');
                     sb.Append(ValueToString(kvp.Value));
-                    sb.Append(",");
+                    sb.Append(',');
                 }
 
                 if (sb.Length > 0)
@@ -140,7 +140,7 @@
                 foreach (var entry in list)
                 {
                     sb.Append(ValueToString(entry));
-                    sb.Append(";");
+                    sb.Append(';');
                 }
 
                 if (sb.Length > 0)
--- a/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
@@ -51,6 +51,11 @@
                 deserializedHeaders[Headers.ReplyToAddress] = properties.ReplyTo;
             }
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -86,26 +91,26 @@ public static string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
             return properties.MessageId;
         }
 
-        static Dictionary<string, string> DeserializeHeaders(IDictionary<string, object> headers)
+        static Dictionary<string, string> DeserializeHeaders(IDictionary<string, object> headers, int extraCapacity)
         {
-            var deserializedHeaders = new Dictionary<string, string>();
-
-            if (headers is Dictionary<string, object> messageHeaders)
```
danielmarbach, line 93: Technically it would also be possible to keep the pattern match while still doing these optimizations if we think it is worth it

danielmarbach:
Ping @bording

src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -86,26 +91,26 @@ public static string DefaultMessageIdStrategy(BasicDeliverEventArgs message)
...
+            if (headers is not Dictionary<string, object> headerDictionary)
```
bording, line 96: With all the improvements to devirtualization that have happened, I wonder if there is even a point in keeping this in here. 

Honestly, at this point it's probably better to just use the `IDictionary` as its given to us and not worry about it.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -39,7 +39,12 @@ public string RetrieveMessageId(BasicDeliverEventArgs message, Dictionary<string
                 messageHeaders.Remove(BasicPropertiesExtensions.ConfirmationIdHeader);
             }
 
-            var deserializedHeaders = DeserializeHeaders(messageHeaders);
+            // Leaving space for ReplyTo, CorrelationId, DeliveryMode, EnclosedMessageTypes and CallbackQueue conditionally
+            // added below. This is a bit cumbersome and need to be changed when things are conditionally added below
+            // but it prevents the header dictionary from growing and relocating which creates quite a bit of
+            // memory allocations and eats up CPU cycles.
+            const int extraCapacity = 5;
```
bording, line 46: I see that you've counted 5 things here, but there are actually only 4 possible additional things that get added below.

`CallbackQueue` gets looked for, but it's another `ReplyTo` potentially added in that case.

bording: More generally, I'm wondering how important this extra capacity actually is. In most normal cases, I'm not sure these would even be adding anything to the header dictionary.

The more impactful change I suspect is setting the initial capacity at all, using the size of the `IDictionary`.

src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
```diff
@@ -132,10 +137,11 @@ static string ValueToString(object value)
             {
                 var sb = new StringBuilder();
 
-                foreach (var entry in list)
+                for (int index = 0; index < list.Count; index++)
```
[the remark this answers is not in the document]
danielmarbach: It is not strictly necessary but they tend to be slightly faster. Not important or impactful enough though that is is worth discussing it at length. Happy to roll back

src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
danielmarbach: It is important because when you set the capacity and then go over even with one it starts growing again because you asked for an initial capacity. This change is about setting a reasonable capacity that fits all cases and we don't have to grow the capacity. I guess I have to learn counting then :D *back to primary school Daniel!*

src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs
danielmarbach: I remember we discussed this on a call and you were sceptic about me removing it. I brought up the argument of devirtualization. I did a quick investigation sharplab but wasn't entirely grasping yet if the new version would actually be properly devirtualized and given you showed some skepticism and I did not have a lot more time to actually proof things I went back to the initial version. But generally I think we should if possible use the type that is exposed on the public API and not check for internal details if not strictly necessary. So I restored the `IDictionary` version

danielmarbach:
@bording have you had a chance to give this some thought?

bording (approved):

PR #1411, opened 2024-06-19
Optimize the delay delivery routing key calculation for throughput and fewer allocations

danielmarbach:
Based upon #1412 

This PR optimizes the delay delivery routing key algorithm to:

- No longer use the BitArray which is a class that leads to allocations and internally copies the array passed
- No longer use the string builder which allocates quite a bit of char array garbage

The routing key length is well-defined by the number of levels we allow plus the space needed for the address. This means we can use `string.Create` to get rid of the `StringBuilder`.

As a replacement for the `BitArray` the `BitVector32` is a nice fit which is a struct.

To avoid having to loop through the bit vector multiple times or encoding some custom string parts into the string created to then splice the delayLevel out again I had to do a bit of more advanced trickery to be able to write the delayLevel which requires some unsafe declaration. That is fine though and even something the underlying RabbitMQ client uses.

image

<details>

## Benchmark

```csharp
using System;
using System.Collections;
using System.Collections.Generic;
using System.Collections.Specialized;
using System.Runtime.CompilerServices;
using System.Text;
using BenchmarkDotNet.Attributes;

namespace MicroBenchmarks.RabbitMQ;

[SimpleJob]
[MemoryDiagnoser]
public class RoutingKey
{
    [Benchmark(Baseline = true)]
    [ArgumentsSource(nameof(Arguments))]
    public string Before(int delay, string address) => CalculateRoutingKeyOld(10, "some-address", out _);
    
    [Benchmark]
    [ArgumentsSource(nameof(Arguments))]
    public string After(int delay, string address) => CalculateRoutingKeyNew(10, "some-address", out _);
    
    public IEnumerable<object[]> Arguments()
    {
        yield return [0, "some-address"];
        yield return [10, "some-address"];
        yield return [MaxDelayInSeconds - 1, "almost-max"];
        yield return [MaxDelayInSeconds, "max"];
    }
    
    public static string CalculateRoutingKeyOld(int delayInSeconds, string address, out int startingDelayLevel)
    {
        if (delayInSeconds < 0)
        {
            delayInSeconds = 0;
        }

        var bitArray = new BitArray(new[] { delayInSeconds });
        var sb = new StringBuilder();
        startingDelayLevel = 0;

        for (var level = MaxLevel; level >= 0; level--)
        {
            if (startingDelayLevel == 0 && bitArray[level])
            {
                startingDelayLevel = level;
            }

            sb.Append(bitArray[level] ? "1." : "0.");
        }

        sb.Append(address);

        return sb.ToString();
    }
    
    public static unsafe string CalculateRoutingKeyNew(int delayInSeconds, string address, out int startingDelayLevel)
    {
        if (delayInSeconds < 0)
        {
            delayInSeconds = 0;
        }

        startingDelayLevel = 0;

        fixed (int* startingDelayLevelPtr = &startingDelayLevel)
        {
            var addr = (IntPtr)startingDelayLevelPtr;

            return string.Create((2 * MaxLevel) + 2 + address.Length, (address, delayInSeconds, addr), Action);

            static void Action(Span<char> span, (string address, int, IntPtr) state)
            {
                var (address, delayInSeconds, startingDelayLevelPtr) = state;

                var startingDelayLevel = 0;
                var mask = BitVector32.CreateMask();

                var bitVector = new BitVector32(delayInSeconds);

                var index = 0;
                for (var level = MaxLevel; level >= 0; level--)
                {
                    var flag = bitVector[mask << level];
                    if (startingDelayLevel == 0 && flag)
                    {
                        startingDelayLevel = level;
                    }

                    span[index++] = flag ? '1' : '0';
                    span[index++] = '.';
                }

                address.AsSpan().CopyTo(span[index..]);

                Unsafe.Write(startingDelayLevelPtr.ToPointer(), startingDelayLevel);
            }
        }
    }
    
    const int maxNumberOfBitsToUse = 28;

    public const int MaxLevel = maxNumberOfBitsToUse - 1;
    public const int MaxDelayInSeconds = (1 << maxNumberOfBitsToUse) - 1;
}
```
</details>

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
@@ -86,20 +86,20 @@
             }
 
             startingDelayLevel = 0;
-            // Pinning the address of the startingDelayLevel so that we can safely write back to the address of the local
-            // This trickery is done because when string.Create returns there is no way to pass state outside of the lambda
-            // without doing a closure over a local variable which would create DisplayClass allocations for every call.
-            fixed (int* startingDelayLevelPtr = &startingDelayLevel)
+            // Pinning the address of the startingDelayLevel so that we can safely write back to the address of the local.
+            // This trickery is done because when string.Create returns, there is no way to pass state outside of the lambda
+            // without doing a closure over a local variable, which would create DisplayClass allocations for every call.
+            fixed (int* pinnedStartingDelayLevel = &startingDelayLevel)
             {
-                var addr = (IntPtr)startingDelayLevelPtr;
+                var startingDelayLevelPtr = (nint)pinnedStartingDelayLevel;
 
-                // The length of the string is determined by the max level taking into account the number of dots, the address length
+                // The length of the string is determined by the max level, taking into account the number of dots, the address length
                 // and additional space since we are zero based.
-                return string.Create((2 * MaxLevel) + 2 + address.Length, (address, delayInSeconds, addr), Action);
+                return string.Create((2 * MaxLevel) + 2 + address.Length, (delayInSeconds, address, startingDelayLevelPtr), Action);
 
-                static void Action(Span<char> span, (string address, int, IntPtr) state)
+                static void Action(Span<char> span, (int, string, nint) state)
                 {
-                    var (address, delayInSeconds, startingDelayLevelPtr) = state;
+                    var (delayInSeconds, address, startingDelayLevelPtr) = state;
 
                     var startingDelayLevel = 0;
                     var mask = BitVector32.CreateMask();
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs
```diff
@@ -78,30 +78,44 @@ public static void TearDown(IModel channel)
...
+        public static unsafe string CalculateRoutingKey(int delayInSeconds, string address, out int startingDelayLevel)
         {
             if (delayInSeconds < 0)
             {
                 delayInSeconds = 0;
             }
 
-            var bitArray = new BitArray(new[] { delayInSeconds });
-            var sb = new StringBuilder();
             startingDelayLevel = 0;
```
danielmarbach, line 88: To be extra safe, would it be better to using a local variable that we then assign before exit to startingDelayLevel? Memory on the stack doesn't need to be pinned and therefore should require the fixed keyword here but I'm not 100% certain how it works with out-variables.

bording (approved):

PR #1431, opened 2024-08-09
Try new rabbitmq action

danielmarbach:
Validates https://github.com/Particular/setup-rabbitmq-action/pull/38


danielmarbach:
The local docker being fast seems to expose a race in the concurrency tests. One is that the concurrencyTask is null

PR #1446, opened 2024-08-28
Prepare for RabbitMQ.Client v7.0.0

danielmarbach:


src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/ConfigureEndpointRabbitMQTransport.cs
```diff
@@ -43,14 +41,14 @@ void PurgeQueues()
 
         var queues = transport.QueuesToCleanup.Distinct().ToArray();
```
paulomorgado, line 42: Consider using [`Enumerable.ToHashSet` Method](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tohashset)

```suggestion
        var queues = transport.QueuesToCleanup.ToHashSet();
```

paulomorgado:
@danielmarbach, beware that primary constructors are not quite the same as "normal" constructors.

Parameters are either erased or are not read-only fields.

I never use them on public facing APIs because you can't make them read-only and validate on constructor.

For testing and internal APIs they are great because they reduce the ceremony of having to write a constructor.

danielmarbach:
@paulomorgado yeah I'm aware of these restrictions

> I never use them on public facing APIs because you can't make them read-only and validate on constructor.

All the classes I touched are internal ;)

For the cases I touched `readonly` doesn't really matter.

That being said this is also not an important change at all and can easily be rolled back and I was planning to discuss with @bording to split those changes out into a dedicated PR (which is probably better anyway)

src/NServiceBus.Transport.RabbitMQ/Sending/MessageDispatcher.cs
```diff
@@ -38,9 +38,9 @@ public async Task Dispatch(TransportOperations outgoingMessages, TransportTransa
                     tasks.Add(PublishMessage(operation, channel, cancellationToken));
                 }
 
-                channelProvider.ReturnPublishChannel(channel);
-
                 await Task.WhenAll(tasks).ConfigureAwait(false);
+
+                channelProvider.ReturnPublishChannel(channel);
```
danielmarbach, line 43: Would this require now a finally?

bording: Yeah, I made a change there to remove the catch/re-throw and instead just have a finally and rely on the pool to handle the dispose as needed.

danielmarbach:
@bording I wasn't able to properly finish the last change because of an appointment. I pushed it nonetheless. Happy to pick this up later if you have no time to continue

bording (approved):

PR #1468, opened 2024-10-14
Allow setting `content-type` AMQP attribute via `DispatchProperties` for the purposes of native integration directly via the transport seam

SzymonPobiega:


irinascurtu (approved):

tmasternak (approved):

irinascurtu (approved):

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -73,4 +73,8 @@ namespace NServiceBus.Transport.RabbitMQ
         System.Threading.Tasks.ValueTask SetupSubscription(RabbitMQ.Client.IChannel channel, NServiceBus.Unicast.Messages.MessageMetadata type, string subscriberName, System.Threading.CancellationToken cancellationToken = default);
         System.Threading.Tasks.ValueTask TeardownSubscription(RabbitMQ.Client.IChannel channel, NServiceBus.Unicast.Messages.MessageMetadata type, string subscriberName, System.Threading.CancellationToken cancellationToken = default);
     }
+    public static class NativeIntegrationDispatchPropertiesExtensions
```
danielmarbach, line 76: Quick question:

This PR describes a native integration scenario, but as far as other native rabbitmq libraries are concerned it doesn't really care about the content type property at all. Is this "native integration" specific to a specific integration scenario?

It also seems that there are some other "native integration" related scenarios popping up that require some low level dispatcher / receiver configuration but the domain specific language is not aligned for the extension methods etc.

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1072

src/NServiceBus.Transport.RabbitMQ.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -73,4 +73,8 @@ namespace NServiceBus.Transport.RabbitMQ
...
+    {
+        public static void SetContentType(this NServiceBus.Transport.DispatchProperties dispatchProperties, string contentType) { }
```
danielmarbach, line 78: By looking at the dispatch properties it means whenever a transport operation is available this method can be set. For example, it is possible to interact with PendingTransportOperations over the extension bag. If you do that the content type can be set which sets the property in the dispatch properties. Those get serialized into the outbox message which means the information is persisted into the outbox. Is that desired? If yes should this be documented at least with a comment somewhere?

src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs
```diff
@@ -57,7 +57,11 @@ public static void Fill(this IBasicProperties properties, OutgoingMessage messag
                 }
             }
 
-            if (messageHeaders.TryGetValue(NServiceBus.Headers.ContentType, out var contentType) && contentType != null)
+            if (dispatchProperties.TryGetValue(NativeIntegrationDispatchPropertiesExtensions.ContentTypeAttribute, out var contentType) && contentType != null)
```
danielmarbach, line 60: By looking at the PR description on the order of evaluation, it is unclear why dispatch properties should take precedence over the presence of the header. By looking at the code change one could argue that the now content type attribute is a temporary header (one that gets persisted into the outbox too) which the only difference that is never in the "NSB headers" but technically for integration scenarios in a raw dispatch setting the `NServiceBus.Headers.ContentType` would achieve the same and it is not clear why that route is not desirable.

danielmarbach, line 60: By looking at the scenarios in ASB and RabbitMQ it seems that [ASB has the possibility to customize the native outgoing message](https://docs.particular.net/transports/azure-service-bus/native-message-access#access-to-the-native-azure-service-bus-outgoing-message) to the needs of the user while the rabbitmq transport doesn't provide that. Has it been considered to provide a general outgoing native message customization opportunity in the transport to unlock similar customization scenarios like ASB has that could also make it possible to override basic property values where desirable?

FYI I'm not a huge fan of the customization in ASB because it is implemented a bit in awkward ways but at least considering some alignment seems warranted

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -91,6 +93,11 @@ public TimeSpan TimeToWaitBeforeTriggeringCircuitBreaker
             }
         }
 
+        /// <summary>
+        /// The callback to use when customizing the message before dispatching it to the broker.
+        /// </summary>
+        public Action<IOutgoingTransportOperation, IBasicProperties> OutgoingNativeMessageCustomization { get; set; }
```
danielmarbach, line 99: Would it be possible to go the extra mile to describe the potential danger of this operation including the flow of things like transport customization gets applied first and then this thing is called which means you can manipulate things that might lead to things not working elsewhere?

src/NServiceBus.Transport.RabbitMQ/Sending/MessageDispatcher.cs
```diff
@@ -10,11 +10,13 @@
...
     {
         readonly ChannelProvider channelProvider;
+        readonly Action<IOutgoingTransportOperation, IBasicProperties> messageCustomization;
         readonly bool supportsDelayedDelivery;
 
-        public MessageDispatcher(ChannelProvider channelProvider, bool supportsDelayedDelivery)
+        public MessageDispatcher(ChannelProvider channelProvider, Action<IOutgoingTransportOperation, IBasicProperties> messageCustomization, bool supportsDelayedDelivery)
         {
             this.channelProvider = channelProvider;
+            this.messageCustomization = messageCustomization ?? ((_, _) => { });
```
danielmarbach, line 19:
```suggestion
            this.messageCustomization = messageCustomization ?? static ((_, _) => { });
```

src/NServiceBus.Transport.RabbitMQ.TransportTests/When_customizing_outgoing_messages.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.TransportTests;
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using global::RabbitMQ.Client;
+using global::RabbitMQ.Client.Events;
+using NServiceBus.TransportTests;
+using NUnit.Framework;
+
+public class When_customizing_outgoing_messages : NServiceBusTransportTest
```
danielmarbach, line 10: For my own benefit. My thought was that this is more of a user-facing option now so intuitively I would have written an acceptance test. In the acceptance test you also have the possibility to configure the transport at the level of the endpoint, which means you don't have to have the awkward coupling that requires a comment to explain why the customization is added in the configure class. 

Curious to hear why you did this as a transport test

SzymonPobiega: I guess it was the easiest option but I think you are right. I will re-write it as an acceptance test.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
SzymonPobiega: 👍

tmasternak (approved):

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -91,6 +93,11 @@ public TimeSpan TimeToWaitBeforeTriggeringCircuitBreaker
             }
         }
 
+        /// <summary>
+        /// The callback to use when customizing the message before dispatching it to the broker.
+        /// </summary>
```
danielmarbach, line 98: How about something like this?

```suggestion
        /// <summary>
        /// Gets or sets the action that allows customization of the native <see cref="BasicProperties"/> 
        /// just before it is dispatched to the rabbitmq client. 
        /// </summary>
        /// <remarks>
        /// <para>
        /// This customization is applied after any configured transport customizations, meaning that 
        /// any changes made here may override or conflict with previous transport-level adjustments. 
        /// Exercise caution, as modifying the message at this stage can lead to unintended behavior 
        /// downstream if the message structure or properties are altered in ways that do not align 
        /// with expectations elsewhere in the system.
        /// </para>
        /// </remarks>
```

SzymonPobiega: Thanks!

SzymonPobiega:
Folks, given that we were able to verify it work OK in the connector, could you re-approve so that I can merge and release it?

WilliamBZA (approved):

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -5,11 +5,13 @@
     using System.Security.Cryptography.X509Certificates;
     using System.Threading;
     using System.Threading.Tasks;
+    using RabbitMQ.Client;
     using RabbitMQ.Client.Events;
     using Transport;
     using Transport.RabbitMQ;
     using ConnectionFactory = Transport.RabbitMQ.ConnectionFactory;
 
+
```
danielmarbach, line 14:
```suggestion
```

danielmarbach (approved):

ramonsmits (approved):

PR #1471, opened 2024-10-21
Make ConnectionConfiguration public

### LINKED ISSUE #1470: Make ConnectionConfiguration public
### Describe the suggested improvement

Sometimes you need to deal with the RabbitMQ connection string for other reasons. Therefor you've to parse the connection string and handle the different schemas (amqp or sql-style).

You already have a perfect implementation, the `NServiceBus.Transport.RabbitMQ.ConnectionConfiguration.Create` method. Unfortunately it cannot be used because the `ConnectionConfiguration` class is private.

Please make this class public so it can be re-used.

### Additional Context

_No response_

twenzel:
Make ConnectionConfiguration public in order to be re-usable.

Fixes #1470


andreasohlund:
See https://github.com/Particular/NServiceBus.RabbitMQ/issues/1470#issuecomment-2437228609

PR #1500, opened 2024-11-18
Reword API comment

SzymonPobiega:


src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -98,10 +98,10 @@ public TimeSpan TimeToWaitBeforeTriggeringCircuitBreaker
...
         /// <remarks>
         /// <para>
-        /// This customization is applied after any configured transport customizations, meaning that 
-        /// any changes made here may override or conflict with previous transport-level adjustments. 
-        /// Exercise caution, as modifying the message at this stage can lead to unintended behavior 
-        /// downstream if the message structure or properties are altered in ways that do not align 
+        /// When provided the customization is called after all other transport customizations configured, meaning that 
+        /// any changes made by customization logic may override or conflict with previous transport-level adjustments. 
+        /// This extension point should be used with caution, as modifying a native message at this stage can lead to unintended behavior 
+        /// downstream if the message content or properties are altered in ways that do not align
```
WilliamBZA, line 104: I'm not sure, but it doesn't quite read clearly to me. Does this seem better?

```suggestion
        /// When provided, the action is invoked after all other transport customizations have executed.
        /// This means that any changes made by the native customization logic may override or conflict
        /// with previous transport-level adjustments. This extension point should be used with caution,
        /// as modifying a native message at this stage can lead to unintended behavior if the message
        /// content or properties are altered in ways that do not align with expectations.
```

WilliamBZA (approved):

PR #1512, opened 2024-11-27
RabbitMQ 4.0 support

abparticular:
In order to properly support the new RabbitMQ 4.0 quorum queue default delivery limit, the following changes have been made:

- The [RabbitMQ management plugin](https://www.rabbitmq.com/docs/management) must be enabled on the broker, and the management API must be accessible by the transport.
  - By default, the transport infers the necessary settings from the connection string to connect to the management API, but if those are not correct, configuration APIs have been added to override the inferred values.
- The management API is now used to verify all broker requirements, such as minimum supported versions and feature flags.
- The management API is used to validate that the delivery limit value of a quorum queue is set to `unlimited (-1)` to prevent the delivery limit setting from interfering with NServiceBus recoverability and potentially causing message loss.
  - When necessary, a policy is written to apply to the queue to attempt to set the delivery limit to `unlimited (-1)`.
  - If the delivery limit is not `unlimited (-1)` and it's determined that another policy is already applied to the queue, the transport will prevent the endpoint from starting. This validation can be disabled for specific scenarios where the automated validation doesn't pass but manual validation has been done and determined to be safe. In general it is not recommended to disable validation.
 - The command line tool has also been updated to use the management API
   A new `queue validate-delivery-limit` command has been added to enable the scripting of the validation check (and if needed, the creation of the policy)

Connected to https://github.com/Particular/NServiceBus.RabbitMQ/issues/1550


src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
```diff
@@ -0,0 +1,182 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration;
```
bording, line 3:
```suggestion
namespace NServiceBus.Transport.RabbitMQ;
```

src/NServiceBus.Transport.RabbitMQ/Administration/IBrokerVerifier.cs
```diff
@@ -0,0 +1,13 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration;
+
+using System.Threading.Tasks;
+using System.Threading;
+
+interface IBrokerVerifier
```
bording, line 8: I don't really see the benefit in having this interface. Is there a specific reason to have it?

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/Converters/DeliveryLimitConverter.cs
```diff
@@ -0,0 +1,34 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient.Converters;
```
bording, line 3: Let's change all of the management client namespaces to 
```suggestion
namespace NServiceBus.Transport.RabbitMQ.ManagementClient;
```

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/IManagementClient.cs
```diff
@@ -0,0 +1,18 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient;
+
+using System.Threading;
+using System.Threading.Tasks;
+using NServiceBus.Transport.RabbitMQ.Administration.ManagementClient.Models;
+
+interface IManagementClient
```
bording, line 9: Do we really need this interface?

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/IManagementClientFactory.cs
```diff
@@ -0,0 +1,8 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient;
+
+interface IManagementClientFactory
```
bording, line 5: Same question, why the interface?

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/ManagementClientFactory.cs
```diff
@@ -0,0 +1,8 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient;
+
+class ManagementClientFactory(ConnectionConfiguration connectionConfiguration) : IManagementClientFactory
```
bording, line 5: Why do we need this factory class at all? Seems like we could remove and just create the management client directly?

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/Models/FeatureFlags.cs
```diff
@@ -0,0 +1,32 @@
+﻿#nullable enable
+
+namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient.Models;
+
+static class FeatureFlags
+{
+    /// <summary>
```
bording, line 7: We typically don't put XML comments on non-public types since that's mainly for intellisense for consuming projects. If there's not a compelling reason to have them here, I'd prefer to stay consistent and remove them.

src/NServiceBus.Transport.RabbitMQ/Administration/ManagementClient/Models/QueueType.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NServiceBus.Transport.RabbitMQ.Administration.ManagementClient.Models;
+
+/// <summary>
```
bording, line 3: Same comment about xml comments.

src/NServiceBus.Transport.RabbitMQ/Configuration/ConnectionConfiguration.cs
```diff
@@ -9,8 +9,10 @@
     class ConnectionConfiguration
     {
         const bool defaultUseTls = false;
-        const int defaultPort = 5672;
-        const int defaultTlsPort = 5671;
+        const int defaultBrokerPort = 5672;
```
bording, line 12: I don't think any of the changes in this file should be done. Instead, the management client should just take a regular `ConnectionConfiguration` instance. The knowledge of needing to override the ports for the management client should be there instead.

src/NServiceBus.Transport.RabbitMQ/Configuration/RabbitMQTransportSettingsExtensions.cs
```diff
@@ -96,6 +96,36 @@ public static TransportExtensions<RabbitMQTransport> ConnectionString(this Trans
             return transportExtensions;
         }
 
+        /// <summary>
```
bording, line 99: Looks like also need a way to disable the management client usage through the legacy config API.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -39,16 +43,21 @@ public RabbitMQTransport(RoutingTopology routingTopology, string connectionStrin
             ArgumentNullException.ThrowIfNull(connectionString);
 
             RoutingTopology = routingTopology.Create();
-            ConnectionConfiguration = ConnectionConfiguration.Create(connectionString);
+            BrokerConnectionConfiguration = ConnectionConfiguration.Create(connectionString);
+
+            ManagementConnectionConfiguration = string.IsNullOrEmpty(managementConnectionString) ?
+                ConnectionConfiguration.ConvertToManagementConnection(BrokerConnectionConfiguration) :
+                ConnectionConfiguration.Create(managementConnectionString, isManagementConnection: true);
```
bording, line 50: Oh, yeah my comment about not using `ConnectionConfiguration` above... yeah that needs to change.

The management client connection string is going to be an http/s URL, which isn't something `ConnectionConfiguration` knows how to parse, so we should never be trying to pass that in.

`ConnectionConfiguration` is for parsing our own made up connection string format and the `amqp` format that the client understands.

src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
```diff
@@ -39,16 +43,21 @@ public RabbitMQTransport(RoutingTopology routingTopology, string connectionStrin
...
 
         /// <summary>
         /// Creates a new instance of the RabbitMQ transport.
         /// </summary>
         /// <param name="routingTopology">The routing topology to use.</param>
         /// <param name="connectionString">The connection string to use when connecting to the broker.</param>
+        /// <param name="managementConnectionString">The connection string to use when connecting to the management API</param>
         /// <param name="enableDelayedDelivery">Should the delayed delivery infrastructure be created by the endpoint</param>
-        public RabbitMQTransport(RoutingTopology routingTopology, string connectionString, bool enableDelayedDelivery)
+        public RabbitMQTransport(RoutingTopology routingTopology, string connectionString, bool enableDelayedDelivery, string managementConnectionString = null)
```
bording, line 60: We should not use optional parameters in public APIs. They can be hard to version properly. Instead, we should add overloads that take the management connection string.

bording (changes requested):

bording (approved):

PR #1563, opened 2025-02-21
skip check for stream_queue feature flag for 3.12+ versions

stefanmoser:
Hi there! I was looking at the code to verify broker requirements and [got curious why a stream needs to be be created](https://discuss.particular.net/t/why-are-rabbitmq-streams-required-for-delayed-delivery/4237) when streams aren't used in any other part of the system. I now understand that it's because the [at-least-once dead lettering in RabbitMQ relies on the `stream_queue` feature flag](https://www.rabbitmq.com/blog/2022/03/29/at-least-once-dead-lettering#usage).

[The `stream_queue` feature flag was elevated to `required` in version 3.12.0](https://github.com/rabbitmq/rabbitmq-server/blob/v3.12.0/deps/rabbit/src/rabbit_core_ff.erl#L23-L29) of RabbitMQ, meaning that all instances of RabbitMQ running on version 3.12.0 or greater are guaranteed to have the `stream_queue` feature flag enabled. ([Additional reference](https://github.com/rabbitmq/rabbitmq-server/blob/v3.12.0/deps/rabbit/src/rabbit_core_ff.erl#L23-L29)) If someone attempts to run a node without the feature flag enabled, [the node will refuse to start up](https://www.rabbitmq.com/docs/feature-flags#list-of-feature-flags).

This PR first checks the version of RabbitMQ and skips the "create a stream" test for all versions greater than or equal to 3.12.0. With this change, users of NServiceBus.RabbitMQ will no longer see the `nsb.v2.verify-stream-flag-enabled` stream on their broker if they are running RabbitMQ version 3.12.0 or greater.

The presence of the test stream queue has previously been [reported as an issue](https://github.com/Particular/NServiceBus.RabbitMQ/issues/1272), but no suitable solution was found and the issue was closed. I believe that this is a good compromise. It still guarantees that the `stream_queue` feature flag is enabled and prevents unnecessarily creating the test stream queue for version greater than or equal to 3.12.0.

What do you think?

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionExtensions.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ConnectionExtensions.cs
@@ -23,23 +23,27 @@ namespace NServiceBus.Transport.RabbitMQ
             }
 
             var brokerVersionString = Encoding.UTF8.GetString(versionBytes);
+            Version.TryParse(brokerVersionString, out var brokerVersion);
 
-            if (Version.TryParse(brokerVersionString, out var brokerVersion) && brokerVersion < minimumBrokerVersion)
+            if (brokerVersion is not null && brokerVersion < minimumBrokerVersion)
             {
                 throw new Exception($"An unsupported broker version was detected: {brokerVersion}. The broker must be at least version {minimumBrokerVersion}.");
             }
 
-            using var channel = await connection.CreateChannelAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
-
-            var arguments = new Dictionary<string, object?> { { "x-queue-type", "stream" } };
-
-            try
+            var streamFlagRequiredMinimumVersion = Version.Parse("3.12.0");
+            if (brokerVersion is null || brokerVersion < streamFlagRequiredMinimumVersion)
             {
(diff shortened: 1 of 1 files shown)


bording:
Hi @stefanmoser, thanks for the PR.

You are correct that the `stream_queue` feature flag is always going to be enabled on 3.12 and newer, and your proposed changes would a valid way to handle the check with that fact in mind.

However, we are already working on a new version of the transport that completely removes this code (and the stream creation) and replaces it with checking feature flags via the management API. You can see the work in progress in #1512 if you are interested in the details.

Given that we are already planning on making changes in this area that are completely removing the code you modified, it doesn't seem to make sense to merge your PR.

Since the end result of our changes will be that the stream won't be created, does this address the concerns that led you to opening the PR in the first place?

stefanmoser:
Thanks for the info. Yes, your re-write addresses my concern. Do you have an estimate of which version it will be released in?

bording:
> Thanks for the info. Yes, your re-write addresses my concern. Do you have an estimate of which version it will be released in?

The change will be part of the new 10.0 major version of the transport that we are working on now.

bording:
@stefanmoser 10.0.0 has been released, so please give a try and see how it works for you!

stefanmoser:
Thank you! I tested 10.0.0 and it works without creating the stream test queue. 🙌

PR #1592, opened 2025-04-09
Add a way to disable broker requirement checks

bording:
While it's not recommended to ever disable the checks the transport makes to ensure the broker meets the minimum requirements required for correct operation, in some very restricted environments it's not possible to let the transport have access to the management API.

In order to unblock using the transport in those sort of environments, this PR adds a new API to tell the transport to disable the checks it makes.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
@@ -32,6 +32,7 @@ class BrokerVerifier(ManagementClient managementClient, BrokerRequirementChecks 
 
     public async Task Initialize(CancellationToken cancellationToken = default)
     {
+        //This needs to stay in sync with changes to BrokerRequirementChecks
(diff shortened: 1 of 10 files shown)


src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
```diff
@@ -39,6 +32,13 @@ public Version BrokerVersion
 
     public async Task Initialize(CancellationToken cancellationToken = default)
     {
+        var all = BrokerRequirementChecks.Version310OrNewer | BrokerRequirementChecks.StreamsEnabled;
+
+        if (disabledBrokerRequirementChecks == all && !validateDeliveryLimits)
+        {
+            return;
```
DavidBoike, line 39: It looks like in the case of disabling everything we are no longer logging a warning?

bording: That happens in `VerifyRequirements`.

DavidBoike (approved):

Issue #1608, opened 2025-05-20
Cannot migrate v1 delay queues to v2 when the exchange contains dots in the name


afprtclr:
#### Description

When running `delay migrate` command to migrate v1 delay queues to v2, the following error occurs when the exchange contains dot in the name;

```
Unhandled exception: RabbitMQ.Client.Exceptions.OperationInterruptedException: The AMQP operation was interrupted: AMQP close-reason, initiated by Peer, code=404, text='NOT_FOUND - no exchange 'SimpleReceiver' in vhost '/'', classId=40, methodId=30
   at RabbitMQ.Client.Impl.Channel.ExchangeBindAsync(String destination, String source, String routingKey, IDictionary`2 arguments, Boolean noWait, CancellationToken cancellationToken)
   at RabbitMQ.Client.Impl.AutorecoveringChannel.ExchangeBindAsync(String destination, String source, String routingKey, IDictionary`2 arguments, Boolean noWait, CancellationToken cancellationToken)
   at NServiceBus.Transport.RabbitMQ.CommandLine.DelaysMigrateCommand.MigrateQueue(IChannel channel, Int32 delayLevel, CancellationToken cancellationToken) in /_/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs:line 114
   at NServiceBus.Transport.RabbitMQ.CommandLine.DelaysMigrateCommand.Run(CancellationToken cancellationToken) in /_/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs:line 50
   at NServiceBus.Transport.RabbitMQ.CommandLine.DelaysMigrateCommand.Run(CancellationToken cancellationToken) in /_/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs:line 48
   at NServiceBus.Transport.RabbitMQ.CommandLine.DelaysMigrateCommand.Run(CancellationToken cancellationToken) in /_/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs:line 48
[...]

PR #1609, opened 2025-05-20
Handling routing keys with dots in delay delivery

afprtclr:
Fix for:

- https://github.com/Particular/NServiceBus.RabbitMQ/issues/1608

Previously used `LastIndexOf` but that doesn't work if destinations have dots in their names. The routing bits are always a fixed length so new implementation calculates the length of the routing bits (bits *2) as each bit is a `1.` or `0.` in the routing key.

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs
+++ b/src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs
@@ -12,6 +12,7 @@
         const string poisonMessageQueue = "delays-migrate-poison-messages";
         const string timeSentHeader = "NServiceBus.TimeSent";
         const string dateTimeOffsetWireFormat = "yyyy-MM-dd HH:mm:ss:ffffff Z";
+        const int indexStartOfDestinationQueue = DelayInfrastructure.MaxNumberOfBitsToUse * 2;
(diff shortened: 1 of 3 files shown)


ramonsmits (approved):

src/NServiceBus.Transport.RabbitMQ.CommandLine/Commands/Delays/DelaysMigrateCommand.cs
```diff
@@ -143,7 +143,7 @@ public static (string DestinationQueue, string NewRoutingKey, int NewDelayLevel)
         {
             var originalDeliveryDate = timeSent.AddSeconds(delayInSeconds);
             var newDelayInSeconds = Convert.ToInt32(originalDeliveryDate.Subtract(utcNow).TotalSeconds);
-            var destinationQueue = currentRoutingKey.Substring(currentRoutingKey.LastIndexOf('.') + 1);
+            var destinationQueue = currentRoutingKey[(DelayInfrastructure.MaxNumberOfBitsToUse * 2)..];
```
ramonsmits, line 146:
```suggestion
            var indexStartOfDestinationQueue = DelayInfrastructure.MaxNumberOfBitsToUse * 2;
            var destinationQueue = currentRoutingKey[indexStartOfDestinationQueue..];
```

ramonsmits (approved):

PR #1611, opened 2025-05-21
Prepare for NServiceBus 10

tamararivera:
Prepare for NServiceBus 10


src/NServiceBus.Transport.RabbitMQ.CommandLine/BrokerConnectionBinder.cs
```diff
@@ -28,12 +28,22 @@ protected override BrokerConnection GetBoundValue(BindingContext bindingContext)
...
+#pragma warning disable IDE0072 // Add missing cases
+                var certificate = contentType switch
+                {
+                    X509ContentType.Cert => X509CertificateLoader.LoadCertificateFromFile(certPath),
+                    X509ContentType.Pkcs12 => X509CertificateLoader.LoadPkcs12FromFile(certPath, certPassphrase),
+                    _ => throw new NotSupportedException($"Certificate content type '{contentType}' is not supported.")
+                };
+#pragma warning restore IDE0072 // Add missing cases
+
+                certificateCollection = [certificate];
```
tamararivera, line 46: **Suggestion:** Can we turn this into a common method, so this class and RabbitMQTransportSettingsExtensions just call it? 
I'm thinking of the case where we need to modify this logic and we will have to touch two different places.

bording: I had considered this and originally decided this wasn't necessary, but after your comment I took another look at it, and yeah it is better.

tamararivera:
I approve this changes

PR #1620, opened 2025-06-06
Only create a single publish channel

### LINKED ISSUE #1621: Channel leak and exhaustion with NServiceBus.RabbitMQ 10.x during mass publish
#### Description

When performing a mass publish operation using NServiceBus.RabbitMQ 10.x, channels are created for each message dispatch due to the use of Task.WhenAll, but they are not properly cleaned up. This results in exceeding the channel_max limit imposed by RabbitMQ (default: 2047), leading to failures.

#### Expected behavior

The message dispatch process should reuse channels where possible, or at least ensure they are cleaned up properly after use, preventing the exhaustion of RabbitMQ’s channel limit. The system should support high-throughput publishing without requiring artificial throttling.

#### Actual behavior

Channels are "leased" and when necessary created per message dispatch due to concurrency (Task.WhenAll), but they are not released properly. This leads to channel exhaustion, causing publishing failures.

#### Versions

Please list the version of the relevant packages or applications in which the bug exists.        

### Steps to reproduce

- With the message session, send/publish multiple hundred messages concurrently
- With the message handler context batch multiple hundred messages sequentially (which leads to the dispatcher doing concurrent operations when the messages are actually send) _or_
- Send/publish multiple concurrent immediate dispatched messages in the handler

### Relevant log output

```shell

```

#### Workarounds

Reduce the concurrency by batching publishes

```csharp
public static async Task BatchPublish<T>(this IMessageSession context, int batchSize = 500, params List<T> events) where T : IEvent
{
if (batchSize > MAX_BATCHSIZE)
batchSize = MAX_BATCHSIZE;

foreach (var set in events.Batch(batchSize))
await Task.WhenAll(set.Select(x => context.Publish(x, new PublishOptions())));
}
```

#### Possible solutions

See #1620

danielmarbach:
Fixes #1621 

Historically the underlying SDK used was dispatching synchronously the network requests and our code added some additional asynchronous parts using a custom scheduler that waited for the publisher confirms to return. The previous guidance of the RabbitMQ client suggested channels should not be re-used across concurrent operations. That's why the channel provider code was added to make sure during concurrent operations clients are never concurrently accessed. Given the synchronous nature of the SDK calls it was rarely ever possible for more than one or two channels being managed by the channel provider. 

With the new async nature of the SDK that also contains an implementation to manage publisher confirms asynchronously (contributed by us) the SDK calls are now completely asynchronous. During the migration to the new client, this detail was overlooked. This leads to massive concurrent access to the channel provider and the channel provider trying to create N number of channels which leads to slow down of publish operations and channels lingering around until the endpoint is restarted or in the worst case the client operations reaching the maximum number of channels allowed.

This PR changes the channel provider to lazy create a single publish channel (since channels are now thread safe as long as they are not used for publishing and consumptions at the same time). The channel provider still applies a rent and return pattern to make sure a faulty channel is closed when returned and new publish operations acquire a fresh channel when needed. 

I have played around with multiple internal implementations using `AsyncManualResetEvent` or CompareExchange/ Spin but ultimately concluded the semaphore version is best given channels are expensive to create. 

## Benchmark

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Diagnosers;
using BenchmarkDotNet.Jobs;
using NServiceBus.Routing;
using NServiceBus.Transport;

namespace RabbitMQPublishBenchmark;

[Config(typeof(Config))]
public class MessageSessionPublish
{
    private TransportInfrastructure infrastructure;
    private IMessageDispatcher messageDispatcher;

    class Config : ManualConfig
    {
        public Config()
        {
            var baseJob = Job.ShortRun;

            AddDiagnoser(MemoryDiagnoser.Default);

            AddJob(baseJob.WithNuGet("NServiceBus.RabbitMQ", "11.0.0-alpha.1.4").WithId("SemaphoreSlim"));
            AddJob(baseJob.WithNuGet("NServiceBus.RabbitMQ", "10.1.2").WithId("10.1.2"));
            AddJob(baseJob.WithNuGet("NServiceBus.RabbitMQ", "9.2.1").WithId("9.2.1"));
        }
    }

    [IterationSetup]
    public void Setup() // does not support async setup
    {
        var transport = new RabbitMQTransport(RoutingTopology.Conventional(QueueType.Quorum, useDurableEntities: true), "host=localhost;username=guest;password=guest");
        infrastructure = transport.Initialize(new HostSettings("RabbitMQPublishBenchmark", "Benchmark", new StartupDiagnosticEntries(),
            (s, exception, arg3) => { }, true), [], [ "destination"]).GetAwaiter().GetResult();
        messageDispatcher = infrastructure.Dispatcher;
    }

    [IterationCleanup]
    public void Cleanup() // does not support async cleanup
    {
        infrastructure.Shutdown().GetAwaiter().GetResult();
    }

    [Params(1000, 1500, 2000)]
    public int Concurrency { get; set; }

    [Benchmark]
    public Task Concurrent_Publish()
    {
        return Task.WhenAll(Enumerable.Range(0, Concurrency).Select(_ =>
        {
            var message = new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), Array.Empty<byte>());
            var transportOperation = new TransportOperations(new TransportOperation(message, new UnicastAddressTag("destination")));

            return messageDispatcher.Dispatch(transportOperation, new TransportTransaction());
        }));
    }

    [Benchmark]
    public async Task Sequential_Publish()
    {
        for (var i = 0; i < Concurrency; i++)
        {
            var message = new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), Array.Empty<byte>());
            var transportOperation = new TransportOperations(new TransportOperation(message, new UnicastAddressTag("destination")));

            await messageDispatcher.Dispatch(transportOperation, new TransportTransaction());
        }
    }
}
```

### Results

image

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -103,7 +103,7 @@ namespace NServiceBus.Transport.RabbitMQ
                     await oldChannel.DisposeAsync().ConfigureAwait(false);
                 }
 
-                var newChannel = new ConfirmsAwareChannel(connection!, routingTopology);
+                var newChannel = new ConfirmsAwareChannel(connection, routingTopology);
                 await newChannel.Initialize(cancellationToken).ConfigureAwait(false);
                 publishChannel = newChannel;
                 return newChannel;
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -9,16 +9,9 @@ namespace NServiceBus.Transport.RabbitMQ
     using global::RabbitMQ.Client.Events;
     using Logging;
 
-    class ChannelProvider : IAsyncDisposable
+    class ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay, IRoutingTopology routingTopology)
+        : IAsyncDisposable
     {
-        public ChannelProvider(ConnectionFactory connectionFactory, TimeSpan retryDelay, IRoutingTopology routingTopology)
-        {
-            this.connectionFactory = connectionFactory;
-            this.retryDelay = retryDelay;
-
-            this.routingTopology = routingTopology;
-        }
-
         public async Task Initialize(CancellationToken cancellationToken = default) => connection = await CreateConnectionWithShutdownListener(cancellationToken).ConfigureAwait(false);
 
         async Task<IConnection> CreateConnectionWithShutdownListener(CancellationToken cancellationToken)
@@ -154,9 +147,6 @@ namespace NServiceBus.Transport.RabbitMQ
             disposed = true;
         }
 
-        readonly ConnectionFactory connectionFactory;
-        readonly TimeSpan retryDelay;
-        readonly IRoutingTopology routingTopology;
         readonly CancellationTokenSource stoppingTokenSource = new();
         volatile IConnection? connection;
         ConfirmsAwareChannel? publishChannel;
--- a/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Connection/ChannelProvider.cs
@@ -84,43 +84,57 @@ namespace NServiceBus.Transport.RabbitMQ
 
         public async ValueTask<ConfirmsAwareChannel> GetPublishChannel(CancellationToken cancellationToken = default)
         {
-            while (true)
+            if (publishChannel is { IsOpen: true })
             {
-                var existing = Interlocked.CompareExchange(ref publishChannel, null, null);
-                if (existing is { IsOpen: true })
+                return publishChannel;
+            }
+
+            try
+            {
+                await publishChannelSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                if (publishChannel is { IsOpen: true })
                 {
-                    return existing;
+                    return publishChannel;
+                }
+
+                var oldChannel = publishChannel;
+                if (oldChannel is not null)
+                {
+                    await oldChannel.DisposeAsync().ConfigureAwait(false);
                 }
 
                 var newChannel = new ConfirmsAwareChannel(connection!, routingTopology);
                 await newChannel.Initialize(cancellationToken).ConfigureAwait(false);
-
-                var expected = existing;
-                var prev = Interlocked.CompareExchange(ref publishChannel, newChannel, expected);
-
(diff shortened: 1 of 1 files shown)


danielmarbach:
I have also tested another implementation using a Lazy that roughly looks something like that

```csharp
        public async ValueTask<ConfirmsAwareChannel> GetPublishChannel(CancellationToken cancellationToken = default)
        {
            var maybePublishChannel = publishChannel;
            if (maybePublishChannel is not null)
            {
                var channel = await maybePublishChannel.Value.ConfigureAwait(false);
                if (channel.IsOpen)
                {
                    return channel;
                }

                Interlocked.CompareExchange(ref publishChannel, null, maybePublishChannel);
            }

            var maybeNewChannel = new Lazy<Task<ConfirmsAwareChannel>>(() => Task.Run(async () =>
            {
                var newChannel = new ConfirmsAwareChannel(connection!, routingTopology);
                await newChannel.Initialize(cancellationToken).ConfigureAwait(false);
                return newChannel;
            }));

            var actualLazy = Interlocked.CompareExchange(ref publishChannel, maybeNewChannel, null) ?? publishChannel!;

            return await actualLazy.Value.ConfigureAwait(false);
        }

        public async ValueTask ReturnPublishChannel(ConfirmsAwareChannel channel, CancellationToken cancellationToken = default)
        {
            if (channel.IsOpen)
            {
                return;
            }

            var maybePublishChannel = publishChannel;
            if (maybePublishChannel is not null)
            {
                var currentChannel = await maybePublishChannel.Value.ConfigureAwait(false);
                if (ReferenceEquals(currentChannel, channel))
                {
                    Interlocked.CompareExchange(ref publishChannel, null, maybePublishChannel);
                    await channel.DisposeAsync().ConfigureAwait(false);
                }
            }
        }

#pragma warning disable PS0018
        public async ValueTask DisposeAsync()
#pragma warning restore PS0018
        {
            if (disposed)
            {
                return;
            }

            await stoppingTokenSource.CancelAsync().ConfigureAwait(false);
            stoppingTokenSource.Dispose();

            var oldConnection = Interlocked.Exchange(ref connection, null);
            oldConnection?.Dispose();

            var maybeOldChannel = Interlocked.Exchange(ref publishChannel, null);
            if (maybeOldChannel is { IsValueCreated: true })
            {
                var oldChannel = await maybeOldChannel.Value.ConfigureAwait(false);
                await oldChannel.DisposeAsync().ConfigureAwait(false);
            }

            disposed = true;
        }

        readonly CancellationTokenSource stoppingTokenSource = new();
        volatile IConnection? connection;
        volatile Lazy<Task<ConfirmsAwareChannel>>? publishChannel;
        bool disposed;
```

image

but there is no additional benefit and the complexity is much higher.

Issue #1664, opened 2025-10-09
The ContentType (and other properties) from BasicProperties is lost/incorrect when messages are retried by ServiceControl because the transport doesn't propagate it to the NServiceBus ContentType header


afprtclr:
The RabbitMQ transport [supports native integration](https://docs.particular.net/transports/rabbitmq/native-integration). As documented, the incoming message:

> The native message must allow NServiceBus to [detect the message type either via the headers or the message payload](https://docs.particular.net/nservicebus/messaging/message-type-detection).

The incoming `BasicDeliverEventArgs` could use the `BasicProperties.ContentType` property to express the incoming message content type. As of today, [that property is ignored by the transport, and it's not propagated to the `NServiceBus.ContentType` header](https://github.com/Particular/NServiceBus.RabbitMQ/blob/961e5d40dab893fb0d3809c50be648045a48774b/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs#L32-L84).

That could lead to two problematic behaviors.

### Additional deserializers are not invoked as expected

Without users manipulating the incoming message in the NServiceBus pipeline before it gets to the deserialization stage, if there are [additional deserializers](https://docs.particular.net/nservicebus/serialization/#specifying-additional-deserializers) configured, NServiceBus will always use the main serializer to deserialize the incoming message.

### Retrying a failed message loses the ContentType property value

The fact that the transport ignores the incoming message `ContentType` property causes issues when retrying failed messages. If the incoming native message uses the `ContentType` property, e.g., `application/json`, when the incoming message fails, without any recoverability pipeline modification setting the `NServiceBus.ContentType` header, what happens is the following:

- The message will be dispatched by the recoverability pipeline to ServiceControl
- The transport in ServiceControl will ignore the ContentType property
- ServiceControl has no pipeline, so effectively the information is lost, and the ingested message will have an empty `NServiceBus.ContentType` header
- When the message is retried, the ContentType property is not set because there is no `NServiceBus.ContentType` header, and the endpoint receives a different message, one with the same body but without the content type property.

### Additional Context

This is the original message sent using the native integration:

<img width="700" height="275" alt="Image" src="https://github.com/user-attachments/assets/f98550ec-0b06-4568-b236-f1eab9d616e2" />

Notice that it has `app_id` property set to `MyNativeApp`, and `content_type` set to `weird_type`. It also has a payload.

When retried, the message looks like this:

[...]

PR #1666, opened 2025-10-14
Propagating Content-Type property from transport to Core

afprtclr:
For https://github.com/Particular/NServiceBus.RabbitMQ/issues/1664


bording (changes requested):

bording (approved):
Changes look good, title of the PR still needs to be updated

PR #1689, opened 2025-11-20
Use large value instead of unlimited when setting delivery limit via policy

### LINKED ISSUE #1688: Newer versions of RabbitMQ server may not respect the unlimited retry policy created by the RabbitMQ transport
### Symptoms

RabbitMQ [recently updated their documentation](https://github.com/rabbitmq/rabbitmq-website/commit/4d2a9dd0009ae4f79acad6c463d85a6be92e9fbc#diff-586b35623b72739a65c42f77d9e00fb182efb840839930242968158262d151c6) to indicate that setting the delivery limit for quorum queues cannot be set using a RabbitMQ policy:

> The value of `-1`, which disables the limit altogether, cannot be currently set through a policy - this
can only be done with `x-delivery-limit=-1` queue argument during queue declaration.

This conflicts with information we were given when the delivery-limit behavior changed for RabbitMQ 4.0 which [prompted us to update the RabbitMQ transport](https://github.com/Particular/NServiceBus.RabbitMQ/pull/1512) last year.

Unfortunately this means that a change in the RabbitMQ broker may result in message loss and our implementation approach may need to change.

This issue was brought to our attention via a support case where a customer upgraded their RabbitMQ broker in their acceptance testing environment and experienced message loss.

We attempted to reproduce the issue and were unsuccessful which indicates the policy may still work for some customers, but is not guaranteed to work:

Here was our attempt to reproduce:

- ran a new RabbitMQ container using 4.1.4 with docker run -d --hostname my-rabbit-4.1.4 --name some-rabbit-4.1.4  -p 5672:5672 -p 15672:15672 rabbitmq:4.1.4-management
- downloaded the latest simple RabbitMQ transport sample
configured the receiver endpoint to do 25 immediate retries

```
var recoverability = endpointConfiguration.Recoverability();
recoverability.Immediate(x => x.NumberOfRetries(25));
```

- changed the command handler to log the current retry count and then throw an exception

We attempted this reproduction for each version of rabbit, stepping backward in versions until reaching 4.0.4.

For each reproduction, we observed 25 immediate retries before the receiver began doing delayed deliveries.  Each delayed delivery attempt was met again with 25 immediate retries until eventually sending the message to the error queue

We also verified that the policy was set by the transport as we'd expect:

Image

When this issue was observed, the value for the `Delivery limit` setting on the right hand side of the queue details displays `-1` instead of `unlimited`.

### Who's affected

Anyone using the NServiceBus.RabbitMQ transport using RabbitMQ server versions 4.1 and above _may_ be affected.

### Root cause

A change in the way the RabbitMQ broker evaluates the delivery limit queue policies.

### Confirmed workarounds

None

bording:
This PR updates the policy created by the broker to set a queue's delivery limit to a large value (100,000) instead of unlimited (-1), which appears to not work properly in some cases.

Fixes #1688

### DIFF
--- a/src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
+++ b/src/NServiceBus.Transport.RabbitMQ/Administration/BrokerVerifier.cs
@@ -176,6 +176,11 @@ class BrokerVerifier(ManagementClient managementClient, BrokerRequirementChecks 
             return false;
         }
 
+        if (limit == -1 && queue.AppliedPolicyName == $"nsb.{queue.Name}.delivery-limit")
+        {
+            return true;
+        }
+
         if (queue.Arguments.DeliveryLimit.HasValue || (queue.EffectivePolicyDefinition?.DeliveryLimit.HasValue ?? false))
         {
             throw new InvalidOperationException($"The delivery limit for '{queue.Name}' is set to the non-default value of '{limit}'. Remove any delivery limit settings from queue arguments, user policies or operator policies to correct this.");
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.RabbitMQ/Administration/ManagementApi/Models/Queue.cs
```diff
@@ -67,6 +67,9 @@ public int GetDeliveryLimit()
             limit = 0;
         }
 
-        return limit ?? -1;
+        return limit ?? UnlimitedDeliveryLimitIndicatorValue;
     }
+
+    public const int UnlimitedDeliveryLimitIndicatorValue = -1_000;
+    public const int BigValueInsteadOfActuallyUnlimited = 100_000;
```
boblangley: I know we are trying to be as limited as possible since this is a "bug" fix, however, should we consider letting the user configure this? I can see an argument for not doing so, in that NSB should never reach this number of retries. This number being a "problem" presupposes a bug in RabbitMQ transport or NServiceBus core, or misconfigured recoverability settings.

boblangley (approved):

ngallegos (approved):

